Issue #4, opened 2015-02-21
WCS coordinates, not pixel coordinates


jeff-regier:
Encode source locations in WCS coordinates, not pixel coordinates. Stamps are pixel aligned, so pixel coordinates were fine before. Frames in different bands aren't exactly pixel aligned.

rgiordan:
I started a "wcs" branch to work on this.

Just a few questions to get started:
- I should be looking at the files in dat/sample_field, right?
- Is there a reason load_field is commented out?
- What does this comment mean: "# temporary, until dustin gets back to me"?
- What does this one mean: "# this doesn't look right...way too small"?
- Anything else I should know?

jeff-regier:
Yes, start with `dat/sample_field`.

"load_field" is commented out because I never finished writing it. I made some progress, but I stopped when I realized that I'd need to fit the PSF myself, and just used stamps for ICML.

The lines immediately following "temporary..." read in metadata from the stamps instead of fields, just as a placeholder. I hadn't figured out how to populate those values for the fields. 

I hoped that value from column 70 was the sky background noise level. It was called "SKY" somewhere, after all. But it looks too small, in relation to the sky noise in the stamps.

The data you want (except for the PSF as a mixture of Gaussians) is probably mostly contained in `sample_field/calibPhotomGlobal-003900-6.fits` and `sample_field/photoField-003900-6.fits`. They're described on the following SDSS web pages:
http://data.sdss3.org/datamodel/files/PHOTO_CALIB/RERUN/RUN/nfcalib/calibPhotomGlobal.html
http://data.sdss3.org/datamodel/files/BOSS_PHOTOOBJ/RERUN/RUN/photoField.html

There's probably a lot more to know, but I don't know it either. :-) I'll forward you the email I got from Dustin about it, but you'll likely want to follow up with him.

jeff-regier:
It looks like FITSIO.jl has some new high-level routines for reading FITS tables, added after I wrote `load_field`. You'll probably want to use them instead of doing it the way I did.

https://julia-fitsio.readthedocs.org/en/latest/api.html#table-operations

rgiordan:
This is done in https://github.com/jeff-regier/Celeste.jl/pull/39 .

Issue #3, opened 2015-02-21
learn the point spread function


jeff-regier:
The SDSS pipeline doesn't approximate the PSF exclusively with Gaussians, so we need to fit a Gaussian mixture model ourselves. [GaussianMixtures.jl](https://github.com/davidavdav/GaussianMixtures.jl) looks useful.

rgiordan:
I think this will require both loading a field and a catalog.  The load_catalog function seems to be out of date, or at least doesn't match the file calibPhotomGlobal-003900-6.fits in the sample_field directory -- can you point me to some reading about how to fix it up?

jeff-regier:
Yeah, "calibObj-..." is different than "calibPhotomGlobal-...". I wrote `load_catalog` with the former in mind. I might have excluded "calibObj-..." from github because it's large.

http://data.sdss3.org/datamodel/files/PHOTO_SWEEP/RERUN/calibObj.html

I found some calibObj files here:

http://data.sdss3.org/sas/dr12/boss/sweeps/dr9/301/

These are the ones you want for `sample_field`:

http://data.sdss3.org/sas/dr12/boss/sweeps/dr9/301/calibObj-003900-6-star.fits.gz
http://data.sdss3.org/sas/dr12/boss/sweeps/dr9/301/calibObj-003900-6-gal.fits.gz

Just came across this page...might be useful.
http://www.sdss.org/dr12/imaging/imaging_access/

rgiordan:
For reference, there is an issue reading in the psField files:
https://github.com/JuliaAstro/FITSIO.jl/issues/33

jeff-regier:
I think I was able to read in variable length fields using the low level table interface. There may be an example of how to do it the `load_field` function.

rgiordan:
That may be true, though all the "field*.fits" fields currently load fine with the high-level FITS library, so I suspect a problem may remain.  I asked that in the other thread.  Let's see what the FITS guys say.

rgiordan:
It turns out that, like iota and epsilon, the PSF varies over the image, which opens the question of how to store and read it.

I think the right way is to store the Gaussian parameters in a coarse grid and use linear interpolation.  At the start of an ELBO evaluation, I'll evaluate the PSF at our current beliefs of the source locations, so that for the purpose of the ELBO, the each source will have its own PSF.

jeff-regier:
How about estimating the PSF at each known star in the catalog (or some
[...]

Issue #2, opened 2015-02-21
set good patch sizes


jeff-regier:
A patch is the region of images affected by a source. The code currently allows the user to specify the size of the patch for each source. But there isn't code yet that picks a good size patch, based on estimates of the brightness and size of the source.

jeff-regier:
Hi @rgiordan , OK if we close this issue? I recall that for now you're setting the patch sizes conservatively. In the  `dat/sample_field` data all the sources look pretty far apart, except for one pair. Are there sources except for this pair that have overlapping patches? If not, the patch sizes are probably good enough as is.

rgiordan:
I'd like to leave this open for now.  What we have is a first step, but is ridiculously conservative and probably has pretty detrimental performance implications.  There are probably big gains to doing this right, and it is not currently done right.

rgiordan:
This problem is fixed with the `trim_source_tiles` function, which is a better solution anyway.

PR #21, opened 2015-04-24
Unconstrain

rgiordan:
Hi Jeff,

I factored out the optimzation from the ELBO calculation.  The transform "rect_transform" (maybe not the best name) should be identical to the original scripts, and now it will be easy to experiment with different parameterizations without having to touch any of the ELBO calculations.  This will also make it possible to add quasars and easier to calculate the hessian.

It's not clear yet whether the unconstrained parameterization alone will buy much in terms of speed.

Comments are welcome, of course.  The unit tests are passing on my machine, but the github build seems to be missing gsl.

Ryan

Ryan


src/CelesteTypes.jl
```diff
@@ -193,34 +205,14 @@ end
 # A vector of variational parameters.  The outer index is
 # of celestial objects, and the inner index is over individual
 # parameters for that object (referenced using ParamIndex).
+
 # TODO: use a matrix here, in conjunction with ArrayViews.jl (?)
+# TODO: Julia noob question -- is there a way to enforce the
+# differences between these variable types?  For now, this is
+# just helpful notation.
```
jeff-regier, line 212: I was wondering that too. I suspect not, because if VariationalParams wasn't just another name for Vector{Vector{Float64}}, we couldn't index into it, multiply it, etc. using functions designed for vectors.

src/OptimizeElbo.jl
```diff
@@ -5,81 +5,10 @@ module OptimizeElbo
...
-            vp[s][p0] = xs2[p1, s] ./ rescaling[p0]
-        end
-        if ids.kappa[1, 1] in left_ids
-            vp[s][ids.kappa[end, :]] = 1. - vp[s][ids.kappa[1, :]]
-        end
-    end
-end
-
-
 function get_nlopt_bounds(vs::Vector{Float64})
```
jeff-regier: Rename this, since we're no longer using nlopt?

src/SampleData.jl
```diff
@@ -0,0 +1,112 @@
+
+module SampleData
+
+using CelesteTypes
+import SDSS
+import OptimizeElbo
+import ModelInit
+
+using Base.Test
```
jeff-regier, line 9: SampleData probably shouldn't import Base.Test. Are the other imports necessary?

src/Transform.jl
```diff
@@ -0,0 +1,374 @@
...
+    #   where the sensitive float contains partial derivatives with respect to the
+    #   variational parameters and returns a sensitive float with total derivatives with
+    #   respect to the transformed parameters.
+
+	to_vp::Function
+	from_vp::Function
+	to_vp!::Function
+	from_vp!::Function
+    vp_to_vector::Function
+    vector_to_vp!::Function
```
jeff-regier, line 37: Looks like the indentation is off here...tab vs spaces?

src/Transform.jl
```diff
@@ -0,0 +1,374 @@
...
+                               ids_free.size)
+
+free_transform = DataTransform(free_to_vp!, vp_to_free!,
+                               vector_to_free_vp!, free_vp_to_vector,
+                               free_unconstrain_sensitive_float, ids_free.size)
+
+identity_transform = DataTransform(unchanged_vp!, unchanged_vp!,
+                                   unchanged_vector_to_vp!, unchanged_vp_to_vector,
+                                   unchanged_sensitive_float, ids.size)
+
```
jeff-regier, line 373: Nice...looks like this approach makes alternate parametrizations straightforward

src/Celeste.jl
```diff
@@ -1,17 +1,21 @@
...
+require(joinpath(Pkg.dir("Celeste"), "src", "OptimizeElbo.jl"))
+require(joinpath(Pkg.dir("Celeste"), "src", "Synthetic.jl"))
+require(joinpath(Pkg.dir("Celeste"), "src", "SampleData.jl"))
 
 import SDSS
 import ElboDeriv
 import OptimizeElbo
 import ModelInit
+import SampleData
+import Transform
```
jeff-regier: "Transform" is pretty general ... would "Reparameterization" work for the module name? Maybe that's not better actually...

src/CelesteTypes.jl
```diff
@@ -229,13 +221,14 @@ immutable ParamIndex
...
+    # The probability of being a particular celestial object (Ia x 1 vector).
+    # (Currently the probability of being a star or galaxy, respectively.)
+    chi::Vector{Int64}
 
-    # The location of the object (2x1 vector)
+    # The location of the object (2x1 vector).
     mu::Vector{Int64}
 
-    # Ix1 scalar variational parameters for r_s.  The first
+    # Ia x 1 scalar variational parameters for r_s.  The first
```
jeff-regier, line 231: Ah, yeah, `Ia` is better than `I`

bin/unconstrained_sandbox.jl
```diff
@@ -0,0 +1,174 @@
...
+
+using Celeste
+using CelesteTypes
+using Transform
+
+import OptimizeElbo
+import ElboDeriv
+
+using Base.Test
+using SampleData
```
jeff-regier, line 25: Would you `import SampleData`, rather than `using SampleData` ?

jeff-regier:
This all looks good. The tests pass on my system too. I filed an issue about the Travis / GSL.jl problem, but I haven't gotten response yet. jiahao/GSL.jl#26

src/CelesteTypes.jl
rgiordan, line 212: Yeah.  Maybe something with type inheritance?

src/SampleData.jl
rgiordan, line 9: Good catch.  I'm fixing this (and responding to your other suggestions) in the unconstrain branch.

bin/unconstrained_sandbox.jl
rgiordan, line 25: No problem.  I'm just curious, why do you prefer one way or the other?

src/Transform.jl
rgiordan, line 37: Yeah.  I have to figure out what's going on in my text editor.  Without going into details it's behaving kind of strangely here.  (It's Sublime, which I'm currently on the fence about purchasing.  How this resolves itself may affect my decision.)

rgiordan:
> I updated the script bin/benchmark.jl in master, and ran it with the both
> master and unconstrained. It looks like unconstrained isn't faster yet,
> but it isn't slower either, so that's promising.
> 
> It looks like benchmark.jl just calculates the ELBO and its derivative, so
> this won't make any difference there.  And the rect_transform should be
> identical to the old code, and it's currently the default transform, so
> everything should also look the same.

Using the free transform I did get much faster (factor of 5) convergence on
the sample star dataset, though.  The galaxy was about the same.  Do you
have a suite of images that you recommend I used for benchmarking and
checking the optimization (not only for this, but going forward)?

bin/unconstrained_sandbox.jl
jeff-regier, line 25: I do `using CelesteTypes` because that file contains types and functions that are used all over...someone developing Celeste needs to be very familiar with them. For all the other modules, I prefer `import`, because the functions in them aren't as second nature...it'd be easier to forget where and in what context these functions were defined

src/CelesteTypes.jl
jeff-regier, line 212: Maybe, though that's more of Java-style solution: Julia only allows subtypes of abstract types. You could create a composite type that has just a single field, whose type is `Vector{Vector{Float64}}`. That would do it. It's also possible that enforcing the difference between `VariationalParams` and `Vector{Vector{Float64}}` wouldn't buy us much.

jeff-regier:
Yeah, I removed my comment about `benchmark.jl` after making it. You're right, it doesn't test the speed of convergence. Timing all the tests together in `test/test_optimization.jl` is probably a pretty good way to benchmark the optimization.

rgiordan:
How about the stamps you looked at for the ICML paper?  It'd be cool to
start looking at more real data sets.

jeff-regier:
`test_real_stamp_optimization()` in `test_optimization.jl` uses a real stamp, that's stored in `Celeste.jl/dat`. The full dataset of stamps is stored in my home directory on NERSC. I can make that directory world readable if it isn't already.

Issues #2 , #3 and #4 maybe more pressing though. If I get SGVB working in time for NIPS (I think I will), we could present results from full frames rather than stamps if you're through those issues in time

PR #28, opened 2015-05-06
Unconstrain

rgiordan:
Hi Jeff,

I should probably make pull requests more often to keep things easier to merge!

The free_transform in this version is not passing tests due to an NLOpt error, but I think it's worth pulling anyway.

Ryan


src/CelesteTypes.jl
```diff
@@ -215,12 +215,30 @@ typealias FreeVariationalParams Vector{Vector{Float64}}
...
+# c1      = C_s means (formerly beta)
+# c2      = C_s variances (formerly lambda) 
+# a       = robability of being a star or galaxy. (formerly chi)
+# k       = Dx{Ia|Ia - 1} matrix of color prior component indicators. (formerly kappa)
+
+# Parameters for location and galaxy shape.
 ue_params = ((:u, 2), (:e_dev, 1), (:e_axis, 1), (:e_angle, 1),
         (:e_scale, 1))
+
+# Parameters for the colors.
```
jeff-regier, line 236: Only c1 and c2 are colors. You might say "band brightnesses", to refer to both "r" and "c".

test/test_optimization.jl
```diff
@@ -395,19 +395,22 @@ end
...
-test_galaxy_optimization()
-test_real_stamp_optimization()  # long running
-#test_peak_init_2body_optimization()
-#test_peak_init_galaxy_optimization()
+#for trans in [ rect_transform free_transform ]
+
+# Currently the optimization does not work with free_transform due
+# to a mysterious NLOpt failure, so do not include it as part of the
+# unit tests.
+for trans in [ rect_transform ]
```
jeff-regier, line 403: Does  rect_transform support either `Ia > 2` or D > 2? (I want to increase both these quantities shortly.)

PR #39, opened 2015-07-05
World coordinate system

rgiordan:
Hi Jeff,

This isn't really ready to merge yet, I just want to open it to start talking about it.

I can run Celeste on parts of real FITS images, but its results aren't very good, and I'm trying to see what's going wrong.  For example, object 1237662226208063565, a star, is classified with high confidence as a galaxy.  You can run it yourself with wcs_sandbox.jl.  I thought your expertise might give me some ideas for where to look.

I wonder if the PSF isn't getting fit well enough.

Want to meet on campus sometime this week, say Wednesday, to look at this together?

Thanks,
Ryan


jeff-regier:
Hey Ryan, I'm away this week from Tues--Sunday at ICML. If you think the PSF might not be accurate, that's a good place to start troubleshooting...might be a good time to get some unit test coverage of the PSF learning code.

If that doesn't fix it, I'd suggest debugging the problem by working exclusively with synthetic data. At least then you know the assumptions of the model hold.

rgiordan:
This is ready to merge.  It incorporates changes from the other pull request, which is smaller, so you might want to do that one first.

It's still unable to successfully identify stars in the wild, but I'm increasingly convinced that the problem is a real modeling problem, not a glitch in the code anywhere.  This change is already quite bloated, and the existing tests pass, so I think it is best to merge now before making even more changes.

jeff-regier:
Sounds good...looks like some tests are failing though

rgiordan:
This is frustrating -- it's an NLOPT failure.  Before I made the pull
request, tests passed on my machine.  I guess I'll try again.  We may be
forced to get to the bottom of that problem somehow.

rgiordan:
In order to diagnose the NLOPT problem, I want to try forward autodiff
(ForwardDiff.jl) on Celeste.  Even though it won't be practical in
production, it will be useful as a test and also as a way to get us some
ideas of what kind of performance improvement we would expect with
Hessians.  This requires making a lot of type changes, since Float64 is
hardcoded into a lot of the Celeste code.

Do you want me to keep making changes on top of this already high-touch
pull request or merge now with this one test disabled?

Ryan

rgiordan:
ps I read a little more about types and actually it's not such an invasive
change, but the question remains.

jeff-regier:
You're talking about using ForwardDiff.jl, to differentiate the derivatives, right? To get the Hessian? That's a good idea, for testing whether we'll get the speedup we need. I was using ForwardDiff earlier...it's very to straightforward to get derivatives with it.

Yeah, please just comment out the problematic test. Can't wait to be done with NLopt!

rgiordan:
The offending tests have been commented out.  After you merge this, I'll send a request for one that allows for forward differentiation.

PR #44, opened 2015-07-16
Forward diff

rgiordan:
This is pretty exciting.  It computes the Hessian in a reasonable amount of time, and it's very very badly conditioned, which is maybe why we've been having trouble getting it to converge.

Let's do wait for the Travis test before merging -- I ran the tests piecemeal as I went along, but haven't run them on the whole thing yet.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,5 +1,7 @@
 Distributions
+DualNumbers
 FITSIO
+ForwardDiff
 WCSLIB
 NLopt
 GSL
--- a/bin/forward_autodiff.jl
+++ b/bin/forward_autodiff.jl
@@ -1,0 +1,231 @@
+using Celeste
+using CelesteTypes
+
+using DataFrames
+using SampleData
+
+import SDSS
+import PSF
+import FITSIO
+import WCS
+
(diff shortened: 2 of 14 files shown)


jeff-regier:
Hey Ryan, Sorry for the delay. Want to meet some time this week to talk about the Hessian?

These changes to ElboDeriv.jl and CelesteTypes.jl presumably slow the program down...fine for exploring the Hessian, but may better to exclude them from the master branch?

rgiordan:
It would be good to meet, yes.  How about Friday after the call?

I don't see why you think they would slow it down -- doesn't Julia compile
a different version of a function for each numeric type as it is
requested?  Or are you talking about something else?

jeff-regier:
Any time on Thursday, or on Monday after 2:30 pm? I'm pretty busy all Friday...midterm 2 is just around the corner.

rgiordan:
Want to video chat this afternoon?  I wasn't planning to come to campus
either of those days.

jeff-regier:
Sure, 3 pm?

rgiordan:
Works for me, talk to you then.

PR #46, opened 2015-07-24
Fix a comment and (currently benign) bug in the description of k

rgiordan:
Sorry about this.  Please double check that this description is correct.


jeff-regier:
Your changes look right. What's with buld errors? Same issue as with the other pull request? (Problems with Distributions.jl?)

PR #53, opened 2015-08-26
Use the new SloanDigitalSkySurvey package

rgiordan:
I'm not sure how this will work with the CI build while waiting for SloanDigitalSkySurvey to be published (pull request: https://github.com/JuliaLang/METADATA.jl/pull/3188).


jeff-regier:
Looks like a test is failing

rgiordan:
Yep, sorry, I think I just pushed a fix, let's see...

rgiordan:
Right now it's failing because the CI build can't get the SloanDigitalSkySurvey package.  I couldn't find how to put a non-published package in a REQUIRE file, so maybe we just have to wait for the pull request to be approved.

Tests pass on my machine.

jeff-regier:
Yeah, I think there's no way to require a package that isn't published.

PR #54, opened 2015-08-28
Newton's Method

rgiordan:
This contains a function that does Newton's method with autodiff Hessians and some related changes.

The files in bin/ are really getting out of hand.  Sorry about that, I'll do a CL to clean them up one of these days.  For this change you can really just ignore most of them.

Ryan


jeff-regier:
Looks good. Does this change let us stop using NLopt.jl? (I see it's still in the REQUIRE file.)

rgiordan:
Nope -- nm is still not very robust or fast, largely I suspect because of
the number of iterations to compute the hessian with forward autodiff.

jeff-regier:
Makes sense that nm isn't fast because of the autodiff, but why not robust?
Numerical instability?

Are you still considering block coordinate ascent, with each block
optimized by nm?

PR #55, opened 2015-08-31
Parallelize elbo

rgiordan:
Hi Jeff,

This refactoring allows distributed computation of the ELBO and derivatives.  As part of that, I made the psf non-constant within the image, factored out the assignment of tiles to sources to a function outside the elbo, and, most importantly, made a "tiled blob" replace the blob as the fundamental data type.

It needs to be a tiled blob for two reasons.  First, this will allow us to avoid large, unnecessary data transfers to shards that will work with only a small section of the image.  Second, the WCSLIB.wcsprm object is actually a C pointer that Julia, reasonably, doesn't let you distribute to multiple processors.  This pointer is a pain in the butt for other reasons, too -- for example, Julia can't serialize it, which means you can't save loaded blobs to more quickly-read julia data types.

This is a refactoring that had to happen eventually.  Right now it won't be very beneficial without a smarter patch size initialization, which is of course on the docket.

Take a look and let me know what you think.

Ryan


jeff-regier:
We still want 1 core per frame ultimately, right? I think we have more frames than cores, so throughput shouldn't improve by using multiple cores for a single frame. But it might be help us to test future coding enhancements, to be able to use a full node on edison (24 cores) to process a single field. Is that that the idea?

jeff-regier:
What's file name "argparse" in the top level? Did this get committed inadvertently last pull request?

rgiordan:
I don't know what argparse is.  If you don't know then it was probably
accidentally committed.

jeff-regier:
I'll delete it. Looks like a postscript file.

rgiordan:
I think it's more complicated than one core per frame, and there are many
other reasons to do it this way.  For one thing, fitting one frame right
now would be prohibitively slow on a single core.  This will allow us to
experiment with entire frames before scaling up to the whole catalog.  I
think it makes more sense to load balance by tiles than by frame.  When we
deal with overlapping fields, we'll need a many-to-many mapping between
frames and  sources anyway.  Also, the "crop image" function I wrote
doesn't work gracefully with actual frames since things like the WCS and
PSF parameterization are hard-wired to the pixel space of the original
frame, making it a pain in the butt to deal with subsets of a frame.  I
think this organization is the way to go.

jeff-regier:
All good points!

jeff-regier:
How much a speedup do you see when evaluating the elbo on, say, 4 processors rather than 1 processor, for an image with a lot of tiles? Is it near 400%?

jeff-regier:
Oh, I meant "cores", not "processors".

rgiordan:
I don't know, I haven't really tried it out on bigger images yet.  :)  I'll
keep you posted.

jeff-regier:
You might find `bin/benchmark.jl` useful for that. It generates a sizable
synthetic field and computes the elbo once on it. Takes about 10 minutes to
process with 1 core.

rgiordan:
I forgot to say, it seems like the speedup should be closer to linear.  I'm
new to distributed computing in Julia, but I think it may well be paying a
lot of data copying costs to perform a single evaluation.  I'll keep
looking into this.

rgiordan:
I only get a 2x speedup with four processors (vs 1).  Also, it only takes
3:30 minutes on one processor on my computer, though I'm not sure what can
account for that if not my hardware.

There were some bugs in my last pull request, and the new local sources
polygon stuff didn't scale well so I needed to do some preprocessing.  I'm
going to send a pull request shortly.

rgiordan:
Yeah, benchmark.jl only takes 40 seconds on my computer in the Celeste
version from this June.  :(

jeff-regier:
Yeah, that's the kind of thing benchmark.jl is there to test. :-) But your version now is computing the Hessian too, right, not just the elbo and it's derivatives? That's not so bad.

I may have been thinking of a version of benchmark.jl with a different number of sources and perhaps a different number of pixels that took 10 minutes to run.

rgiordan:
No, it's not computing the hessian.  Maybe we should meet sometime to sync
up about the changes I've made just so we're on the same page...

jeff-regier:
To get a linear speedup, you might want to load the tiles into a DArray, so that each process gets it's own subset of the tiles. And you might want each process to write gradients to a SharedArray (shared memory) so they don't have to be copied.

Too bad it isn't easier. Looks like the Julia developers are trying to make parallel programming more automatic. But JuliaLang/julia#9167 hasn't yet been assigned a milestone. I guess it's not an easy problem: http://stackoverflow.com/questions/16420792/parallelism-in-julia-features-and-limitations

jeff-regier:
Sure, how about Friday at 4:30 pm?

rgiordan:
I'm only free on Friday from 2-4 or between 11-12.  Do we have a DESI
meeting?

jeff-regier:
Looks like it. Wednesday at 11 works

rgiordan:
Cool, I sent an invite.  See you then!   (I put my office but we can meet
wherever, or go for coffee.)

rgiordan:
Ah, you meant Friday.

jeff-regier:
Oh, yeah, oops :-)

PR #59, opened 2015-09-15
Trust region method, crop to object, and cleaning up initialization

rgiordan:
The branch name is not very meaningful here.


jeff-regier:
Awesome

PR #62, opened 2015-09-24
Automatically set patch sizes, fix WCS bugs, clean up initialization

rgiordan:
Hi Jeff,

The last run of this CL passed on my computer but failed on Travis only because of another NLOpt failure in test_full_elbo_optimization(), which I temporarily disabled.

This way of setting patch sizes is currently extremely conservative.  But it should allow us to actually run on FITS images if a little inefficiently, for now, and required some worthwhile refactoring of how Celeste is initialized.

Ryan


jeff-regier:
Exciting to have the patch sizes working!

Anyway to quantify how conservative the patch sizes might be?

Did you mean to add the .jld files?

What would it take to replace nlopt with your trust region method in the tests? (Changing each test to use a whole optimization library might be a pain, but could you write wrapper for Optim.jl that's more-or-less a drop in replacement?)

REQUIRE
```diff
@@ -4,7 +4,7 @@ Distributions
 Docile
 DualNumbers
 FITSIO
-ForwardDiff 0.0.2 0.0.3
+ForwardDiff 0.0.2 0.0.3+
```
jeff-regier, line 7: Just curious, what happens with more recent versions of ForwardDiff?

rgiordan, line 7: The current release is only compatible with Julia 4.0.  I asked them to tag 0.0.3 which has the support for the typed lgamma hessians that I added.  I haven't been following the stability of 4.0 much, but from the last pull request it looks like maybe you have.  Any thoughts on when we should switch?

src/ModelInit.jl
```diff
@@ -260,48 +335,71 @@ end
...
+  mp = ModelParams(vp, sample_prior())
+
   mp.patches = Array(SkyPatch, mp.S, length(blob))
   mp.tile_sources = Array(Array{Array{Int64}}, length(blob))
 
   for b = 1:length(blob)
     for s=1:mp.S
-      mp.patches[s, b] =
-        SkyPatch(mp.vp[s][ids.u], patch_radius, blob[b], fit_psf=fit_psf)
+      if radius_from_cat
```
jeff-regier: A bit more succinct:

```
patch_args = radius_from_cat ? cat[s] : (mp.vp[s][ids.u], patch_radius)
mp.patches[s, b] = SkyPatch(patch_args..., blob[b], fit_psf=fit_psf)
```

rgiordan:
The unit test test_set_patch_size() both tells you how conservative it is and gives an idea for how to do it in a slower but more principled way in the future by generating an artificial noiseless image and checking what proportion of light lies within the selected patch size.  Right now it gets over 99.9% of the light, which is surely more than necessary, which is to be expected given all the worst case assumptions I made in the current patch size setting function.

I did mean to add the jld files for now.  I'll put them in the dat directory.  I know it's messy but I'd like them to be checked in because they took a long time to generate.

Changing the whole thing over to the TR method is a one-liner, but I haven't explored how well it works.  One potential problem is that TR doesn't detect convergence very quickly, since it seems better than BFGS at continuing to find descent directions (which are just driving the least likely celestial object to zero brightness).  I have some ideas about hacks to fix this, but they're currently just hacks.

Since the VB workshop deadline was pushed back a couple weeks, I want to make it a priority to just run the damn thing as-is on a few images.  That would justify, in my mind, submitting an extended abstract, and then we can tidy everything up later.

REQUIRE
jeff-regier, line 7: It's stable, they're on release candidate 2 now. I think we just need one more change to `PriorParams` to make Celeste compatible with 0.4

rgiordan, line 7: Let's do it, package caching would make my life much better.

jeff-regier:
Sounds like a good plan!

src/ModelInit.jl
rgiordan: Done

rgiordan: This fails with the error

ERROR: `start` has no method matching start(::CatalogEntry)

I'm not sure why.  Rather than mess with it just at this moment I'm just going to change it back.

jeff-regier: It just needs a coma. I'll fix it.

PR #63, opened 2015-09-25
maximize_f not found in the master branch...rolling back to last version with passing tests

jeff-regier:


rgiordan:
Sorry, I apparently accidentally committed changes to the set_patch_sizes branch that I meant to put in a different branch.  But this is keeping those changes, which will break other things.  Let me submit a pull request that actually fixes it.

PR #68, opened 2015-09-30
Improve benchmark

rgiordan:
This improves the benchmark from 120s to about 90s.


rgiordan:
Strange, tests passed on my machine.  Looking into it.

rgiordan:
Compat seems to be having problems.  https://github.com/JuliaLang/Compat.jl/issues/147

PR #69, opened 2015-10-01
Improve benchmark 2

rgiordan:
This gets down to 40s (relative to 20s for the May version on my computer).  There is still a lot more memory allocation going than in May, though, so hopefully more of this to come.

### DIFF
--- a/bin/benchmark.jl
+++ b/bin/benchmark.jl
@@ -22,11 +22,11 @@ function small_image_profile()
     elbo, elbo_time
 end
 
-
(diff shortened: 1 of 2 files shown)


jeff-regier:
3x improvement! How come all the NumType stuff is necessary now? (I don't think I had anything like that before.)

jeff-regier:
Oh, because of all the DualNumbers stuff now, right?

rgiordan:
Yep, because of the DualNumbers.

PR #71, opened 2015-10-01
removes and adds calls to @compat

jeff-regier:
@rgiordan , I removed unnecessary calls to `@compat`, and added some new calls to `@compat`. On my machine Celeste now works with Julia 0.4, but it issues a lot of warnings from other packages that Celeste calls. On Travis-CI it still only works with 0.3 so far, so I haven't modified `.travis.yml`. I think all the printing of warning from other packages may be causing it to run really slowly with Julia 0.4, causing the unit tests to timeout on Travis after 50 minutes. A lot of the warnings are from DataFrames.jl. I'm told those will get fixed this weekend.

### DIFF
--- a/src/CelesteTypes.jl
+++ b/src/CelesteTypes.jl
@@ -443,7 +443,7 @@ end
 align(::StarPosParams, CanonicalParams) = ids.u
 align(::GalaxyPosParams, CanonicalParams) =
    [ids.u; ids.e_dev; ids.e_axis; ids.e_angle; ids.e_scale]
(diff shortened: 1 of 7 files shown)


rgiordan:
LGTM, thanks for taking this on.

Just for my own edification, what's the deal with the `Union` stuff?

```
@compat(Union{Float64, Array{Float64}}
```

jeff-regier:
Parenthesis -> curly bracket, that's all. Btw, usually better to avoid
using unions

PR #77, opened 2015-10-19
Parallelization

rgiordan:
The main point of this CL is the interprocess communication for distributed ELBO computation, together with a benchmark script.  I also made an iJulia notebook with some examples of how the low-level julia RemoteRef()s work.  There is also an attempt to do an approximate block diagonal Hessian that doesn't work very well and which we'll probably ultimately abandon, but for now I'd like to just check it in.


jeff-regier:
Looks good

PR #78, opened 2015-10-21
Efficient Hessian Calculation

rgiordan:
Here's a scalable version of the forward auto diff Hessian.  Also I make the "vector" parameters into matrices that get flattened just before going into optimization routines.  It's not all the way to where it needs to go, but I'll make the pull request in the spirit of more smaller pulls.

### DIFF
--- a/src/ElboDeriv.jl
+++ b/src/ElboDeriv.jl
@@ -884,6 +884,46 @@ end
 
 
 @doc """
+Calculate the ELBO only for tiles containing a particular source.
+This can save a lot of computation when using autodifferentiation to get
+the Hessian.  The values of the ELBO and derivatives will be incorrect,
+but if only VariationalParams associated with deriv_source have non-zero
+epsilon, then the DualNumber derivatives will be correct.
+
+Args:
+  - tiled_blob: The TiledBlob
+  - mp_dual: A dual representation of the ModelParams with all zero epsilons.
+  - accum: A dual number sensitive float.
+  - deriv_source: An integer index from 1 to mp.S.  Only tiles containing
+      deriv_source as a source will be calculated.
+
+Returns:
+  - Clears and updates the SensitiveFloat accum in place.
+""" ->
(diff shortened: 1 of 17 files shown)


jeff-regier:
Looks good. I take it you're trying to get block coordinate ascent running with dual numbers before switching to a manually computing the block diagonal of the Hessian. (?) That seems like a good approach.

rgiordan:
That's right.  This last change makes the autodiff hessian scale pretty
well over multiple processors, and so should be fast enough to run on an
actual frame.  I think it would be cool to see results on real data sooner
rather than later.

Will you be on campus today between 1 and 4?  We could meet to chat a
little.

src/ElboDeriv.jl
```diff
@@ -872,12 +874,52 @@ function tile_predicted_image{NumType <: Number}(
...
+                       star_mcs,
+                       gal_mcs,
+                       accum)
+end
+
+
+@doc """
+Calculate the ELBO only for tiles containing a particular source.
+This can save a lot of computation when using autodifferentiation to get
+the Hessian.  The values of the ELBO and derivatives will be incorrect,
```
jeff-regier, line 889: How come the derivatives might not be correct? They'll be correct for one source (the one we've used to select the relevant tiles), right? Does this function compute derivatives for sources other than that one?

jeff-regier:
Yeah, sounds good, how about 3 pm in the stat lounge?

src/ElboDeriv.jl
rgiordan, line 889: Yes, they'll be correct for sources in `tile_sources` and incorrect for the others.  Of course, the likelihood function computes derivatives for all sources, since `accum` is the size of `mp.S`.  Consider a source that is not in `tile_sources` but influences a tile that is also influenced by a source in `tile_sources`.  Such a source will have correct second derivatives (with respect to a source in `tile_sources`) but incorrect first derivatives returned by this function.

rgiordan:
SGTM, see you then.

PR #79, opened 2015-10-24
Active sources and make trust region the default

rgiordan:
I add "active sources" to the ModelParams object to control which tiles get computed in the ElboDeriv.  I add an objid field to the ModelParams as well -- we'll need it eventually to match up results across different frames.  This allowed me to replace my last commit's special Hessian function and make the more efficient behavior the default.

I also made Newton TR the default optimization and moved most of the tests over to using it.  A couple tests still require solving optimization problems on a couple parameters that start at spots of semi-negative curvature, which my TR implementation can't handle.  For now, for the purpose of these two tests (and possible similar debugging tasks), I left BFGS in as an option with `optimize_f_bfgs`.

### DIFF
--- a/bin/distributed_benchmark.jl
+++ b/bin/distributed_benchmark.jl
@@ -72,13 +72,13 @@ end
 
 # locally
 @time eval_worker_hessian()
-
 @everywhere println(new_hess_time / length(worker_sources))
 
 # Check that the two Hessians are the same.
-new_hess_sparse =
+hess_sparse =
   OptimizeElbo.unpack_hessian_vals(hess_i, hess_j, hess_val, size(x));
 
+workers_hess_times = [ remotecall_fetch(w, () -> new_hess_time) for w in workers() ];
 workers_hess =
   [ remotecall_fetch(w, () -> (hess_i, hess_j, hess_val)) for w in workers()];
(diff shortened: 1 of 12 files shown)


jeff-regier:
What about tiles that don't receive photons from any source? Do they still get factored into the elbo, even though they never contain any active sources?

rgiordan:
No, they won't.  If we ever get into the business of fitting iota or
epsilon or something, that will be easy to change.

jeff-regier:
How about encoding it as `active_tiles` rather than `active_sources` ? The
former contains the latter as a special case, I think. Also arguably easier
to understand, especially if we next allow some sources to be "frozen",
where they contribute to the elbo but aren't optimized.

rgiordan:
I think it's better to keep the model parameters semantically separate from
the image and tiling.  (Right now tile_sources is in ModelParameters, but
eventually the plan is to move it to a separate, third data structure
mapping the two -- imagine a situation with many sources and several images
per source, and you'll see that you want three distinct data structures,
comprising images, parameters, and the mapping between them.)  And though
active tiles is more general, what we really intend is for certain sources
to be active, and for the elbo evaluation to decide what to do with that
information.  I think it's cleaner to do it this way.

jeff-regier:
Sounds good, whatever gets us running on full fields soonest is probably best

PR #80, opened 2015-10-24
Factorize tr

rgiordan:
Factor out the step-taking part of the trust region code.  This will allow us to take steps in parallel if we want to.

The (passing) unit tests for this module are in my clone of Optim.jl; this is just a copy from that repo while I procrastinate making a pull request to Optim.jl.


jeff-regier:
Weird, travis-ci says it's been running for 11 hours with Julia 0.3 . The tests passed in like 10 minutes with Julia 0.4.

jeff-regier:
This might be a good time to stop supporting 0.3 ... I know we still get quite a few warnings from 0.4. I'm chipping away at the warnings with JuliaStats/Clustering.jl#56 and jiahao/GSL.jl#38 .

rgiordan:
Even weirder -- the tests passed successfully, but travis still registers
them as incomplete.

PR #81, opened 2015-10-26
Steps towards full image processing.

rgiordan:
Use active sources in OptimizeElbo.  Add the option to render without background noise.  Pre-process an image into a JLD file.


jeff-regier:
Looks good Ryan. What steps remain?

rgiordan:
I'm running on a single real object now.  Once I make sure that's sensible
and reasonably fast, and have built julia 0.4 on the SCF cluster, I can
spin it up on a bunch sources and see what we get.

PR #89, opened 2015-11-19
Remove fd package req

rgiordan:
Remove the requirement on the ForwardDiff package.

Incidentally, their last release changed a lot of things and their fundamental types now look a lot like SensitiveFloats.  I haven't played with it much, but it might be worth looking at for Celeste.


REQUIRE
```diff
@@ -4,7 +4,7 @@ Distributions
 Docile
 DualNumbers
 FITSIO
-ForwardDiff 0.0.3 0.1
+ForwardDiff # Currently only required for testing.
```
jeff-regier, line 7: I wonder if we could use ForwardDiff.jl for testing the derivatives too, rather than relying on GSL.jl. That library throws a lot of errors/warnings with Julia 0.4 that I don't think will get fixed any time soon, and I worry finite differences could miss discrepancies...it's tricky to set the tolerance high enough to avoid cating rounding errors, but low enough not to miss real errors.

jeff-regier:
Awesome!

REQUIRE
rgiordan, line 7: I'm looking into it.  Superficially it doesn't look as easy as using DualNumbers.  I think they didn't intend for anyone to interact with their data types except through the ForwardDiff API, but we have to in order to use our special data structures.   It would be really nice if we could, though, for many reasons.

jeff-regier, line 7: Not sure I follow... I'm talking about replacing the calls to `deriv_central` (from the `GSL.jl` library) with a calls to FowardDiff.jl --- thus ending our dependence on `GSL.jl`. Oh, are you saying the ForwardDiff API may not be flexible enough for that?

jeff-regier, line 7: Oh, one other problem with GSL.jl : it doesn't compile on Edison. That
didn't stop me before because we just need GSL.jl for tests, and I'd test
it locally before running it there. But, still a pain.

rgiordan, line 7: Yes, I'm saying it's not immediately obvious that the ForwardDiff API will suffer the contortions that the input vector takes as it passes through Transform.jl and ElboDeriv.  But I'm newly optimistic.  In any case, we can always just use our own FD implementation using the old DualNumbers package, so I'm sure we can get rid of GSL.jl.  I'll let you know what I find out.

jeff-regier, line 7: Oh, got it---the new FowardDiff doesn't use DualNumbers . Yeah either way
would be an improvement over GSL

rgiordan, line 7: Unfortunately, `ForwardDiff` types allocate a ton of memory.  We can't use them as a drop-in replacement for our own derivatives or Hessians.  It might be good to point this out to Miles, though.  He might have some ideas.

```
# ForwardDiff types:
julia> @time for n=1:10000
         foo = sum([ mp_fd.vp[1][i] * mp_fd.vp[1][i] for i=1:length(ids)] )
       end
  0.306251 seconds (2.26 M allocations: 592.808 MB, 23.21% gc time)

# Ordinary types:
julia> @time for n=1:10000
         bar = sum([ mp.vp[1][i] * mp.vp[1][i] for i=1:length(ids)] )
       end
  0.093861 seconds (1.29 M allocations: 22.739 MB, 4.02% gc time)
```

jeff-regier, line 7: That doesn't look so bad, if we're just using ForwardDiff.jl for testing
rather than GSL.jl. Or did you have in mind using ForwardDiff.jl outside of
the test directory?

rgiordan, line 7: > That doesn't look so bad, if we're just using ForwardDiff.jl for testing
> rather than GSL.jl. Or did you have in mind using ForwardDiff.jl outside of
> the test directory?

I admit that when I saw the similarity between some of what they've done
and some of what I've started to do for the Hessian, I thought that my
procrastination had really paid off.  But evidently not.

jrevels, line 7: Hi! Sorry to drop in so randomly, but I just recently saw this package (neat stuff!) and happened to see these comments while doing some exploring. 

> Unfortunately, ForwardDiff types allocate a ton of memory

Unless there was a regression we didn't catch, ForwardDiff's number types should be laid out in memory in almost the exact same fashion as `Dual`s, as long as you configure them correctly (specifically `Dual{T}` ~ `GradientNumber{1,T,Tuple{T}}`). When you did your benchmark, what types did you actually use (i.e. what is the type of `mp_fd.vp`/is that code available somewhere)? 

While the point of ForwardDiffNumbers is to allow simultaneous accumulation of multiple derivatives per function call, another lesser-known feature is that the underlying storage type is swappable. You can use either `Vector` or `Tuple` types, which have different performance characteristics (this is exposed in the API via [the chunk_size keyword](http://www.juliadiff.org/ForwardDiff.jl/chunk_vec_modes.html), and to `C` in `ForwardDiffNumber{N,T,C}`). In many cases, using `Vector` storage can trigger GC, which will kill performance. I'd bet that's what's happening in your loop above.

Anyway, I don't mean to intrude - just thought this was a cool package and was curious about the issue you mentioned. I'm also curious about your use case of nesting `Dual`s, but I don't want to take up too much of your time!

PR #99, opened 2015-12-18
remove GSL, NLopt, Docile, and Compat

### LINKED ISSUE #93: Remove GSL.jl dependency
It won't compile easily on NERSC, and it's less accurate than ForwardDiff.jl anyway.

### LINKED ISSUE #94: Remove NLopt from REQUIRE file (?)
We've still have NLopt in the REQUIRE file, but presumably it would be easier to deploy our code on NERSC with fewer dependencies, and installing NLopt.jl takes compiling some C code. 

@rgiordan , you were saying we still need NLopt for a couple of unit tests, right? Are those tests essential? If this is a problem with negative curvature, might the trust region method have the same problems on the actual images that it has in the unit tests?

### LINKED ISSUE #98: remove Docile and Compat---remnants of Julia v"0.3"

jeff-regier:
This closes #98, closes #94, and closes #93 . However, I left 2 tests of transform commented out. Nonetheless, I think it's worth merging this now because it removes many dependencies.


src/Polygons.jl
```diff
@@ -0,0 +1,242 @@
+module Polygons
```
rgiordan: Could you take this out and make sure you merged with my last pull request?

src/Util.jl
```diff
@@ -25,8 +21,8 @@ function get_bvn_cov{NumType <: Number}(ab::NumType, angle::NumType, scale::NumT
     #    The 2x2 covariance matrix parameterized by the inputs.
 
     #@assert -pi/2 <= angle < pi/2
-    @assert 0 < scale
-    @assert 0 < ab <= 1.
+    #@assert 0 < scale
+    #@assert 0 < ab <= 1.
```
rgiordan: Why are you commenting these out?

test/test_derivs.jl
```diff
@@ -8,38 +8,48 @@ using Transform
...
 println("Running derivative tests.")
 
+function do_deriv_test(f::Function, x::Float64, claimed_dx)
+    # TODO: really should call ForwardDiff.gradient just once, with a whole
+    # vector of x values, rather than once per parameter to test
+    fwd_deriv = ForwardDiff.gradient(f)([x,])[1]
+
+    info("got $fwd_deriv; expected $claimed_dx")
+
+    @test_approx_eq_eps significand(fwd_deriv) significand(claimed_dx) 1e-4
```
rgiordan: I think we can just use test_approx_eq with no epsilon now, right?

test/test_derivs.jl
```diff
@@ -290,12 +297,12 @@ function test_derivative_transform()
...
 end
 
-test_kl_divergence_derivs()
 test_brightness_derivs()
+test_kl_divergence_derivs()
 test_accum_pixel_source_derivs()
 test_elbo_derivs()
-test_derivative_transform()
-test_elbo_derivs_with_transform()
+#test_derivative_transform()
```
rgiordan, line 310: Could you put a comment in here about what's wrong with these tests?  I agree that we can merge this for now, but we should eventually put them back in.

test/test_optimization.jl
```diff
@@ -445,8 +445,8 @@ end
 test_quadratic_optimization()
 test_objective_wrapper()
 test_objective_hessians()
-test_kappa_finding()
-test_bad_a_init()
+#test_kappa_finding()
```
rgiordan: These tests are probably less important, but could you still put a comment in about why they're commented out?  (I presume it's the 1D optimization, right?)

rgiordan:
Pretty exciting!  Could you address the inline comments?

test/test_derivs.jl
jeff-regier: It's still not exact, unfortunately. Rounding errors are one possibility.

test/test_optimization.jl
jeff-regier: Yes, and I'll add a comment

src/Util.jl
jeff-regier: Apartment `<` isn't compatible with the GradientNumbers from ForwardDiff. I wish it were.

jeff-regier:
Yeah, not sure how Polygons.jl got added back....I did merge with your latest PR ahead of time and the other changes are in there. It's deleted now.

src/Util.jl
rgiordan: You can check the "real" part, though, right?

test/test_derivs.jl
rgiordan: That seems strange.  I'm testing the Hessian with autodiff and I'm using test_approx_eq.  Also, back when I had the autodiff in the objective wrapper I also recall that the results were pretty close.  I haven't looked closely at how you're doing it, but are you sure something isn't going wrong?

src/OptimizeElbo.jl
```diff
@@ -407,8 +374,8 @@ Returns:
 function maximize_f_bfgs(
```
rgiordan: Could you please delete this function now that NLOpt is gone?

test/test_derivs.jl
jeff-regier: Well nothing new is going wrong...with this PR the derivatives are checked as closely as before. And the changes are all to the test code. 

It might be worth exploring more why some derivatives don't match before 4 significant digits, but I don't see any red flags here. Numerical stability varies with the method of computing something, for example,

```
@test_approx_eq (1e10 - 1e-6 - 1e10) 1e-6
```

fails on my machine.

src/OptimizeElbo.jl
jeff-regier: Done

test/test_derivs.jl
rgiordan: Of course it does, but if we have that kind of floating point calculation in our ELBO calculation then our tests /should/ fail.  (For example, I used autodiff to find one such calculation in the KL divergence, exposing a bug when we were only bounding the mean and variance of the gamma distribution.)

I thought you were doing exact autodiff here, not finite difference differentiation.  If you are, then the results should match to very high precision.  Looking more carefully I'm not confident I understand what you're doing, though, though.  Why are you not just calculating the full gradient with ForwardDiff.gradient and comparing to the sensitive float values?

rgiordan: Although all this is actually going to be replaced soon by the tests in the `exact_hessian` branch, so maybe this isn't worth bothering with right now.

rgiordan:
If you can get the tests to pass, feel free to merge with master ahead of our 10am meeting.  I'll fix the derivative tests in the exact hessian branch.

jeff-regier:
Our derivatives and those from ForwardDiff match now to the tolerance used by `test_approx_eq`. I don't have a good explanation for why they weren't matching exactly before, or why I thought they weren't matching before, or why they now match. ;-) But I'm glad they match!

PR #102, opened 2016-01-13
Speedup combine_sfs_hessian!

jeff-regier:
Hi @rgiordan -- This PR reduces the runtime of `combine_sfs_hessian!` from 1005 ms to 456 ms. Changes:
- pulled as many adds/multiplies outside of the inner for loop as possible
- reordered iteration--the code was looping over rows first, but matrices are column-major
- disable bounds checking in the inner loop with the `@inbounds` macro

Some of the same techniques could work on the other bottleneck(s) too.


rgiordan:
This is good!  When I also limit the loop to the lower triangular part, we're down to 10x the old version.  I'll see if I can apply a similar trick to other parts of the code.

PR #103, opened 2016-01-21
Exact hessian

rgiordan:
Hi Jeff,

I still need to clean up the documentation, especially in ElboDeriv, but this is mostly ready to go.  (There seems to be a flakiness in ForwardDiff which is causing the tests to break occasionally which I am also looking into.)

Since it's such a big CL -- a rewrite of ElboDeriv and much of Transform -- I thought I'd send it to you to take a look a little in advance.  LMK what you think.  I'd like to get it in by the call tomorrow.

Ryan


bin/benchmark_elbo.jl
```diff
@@ -12,7 +12,7 @@ const dat_dir = joinpath(Pkg.dir("Celeste"), "dat")
 srand(1)
 println("Loading data.")
 
-S = 100
+S = 3
```
jeff-regier: Could you leave this as `S = 100`, and call `ElboDeriv.elbo` with `calculate_hessian=false`? We should make sure the run time for computing derivatives hasn't changed despite all the other changes to the code.

jeff-regier:
Everything else looks great....I'm really glad the Hessian is done.

bin/benchmark_elbo.jl
rgiordan: Sure, done.  However, you should know that it is not as fast -- it's perhaps 50% slower.  I spent a fair amount of time trying to get it back to the original speed and am out of ideas.  After I submit, it would be great if you took a pass.  You have a lot more insight than I do into optimizing julia code.

jeff-regier: Would you try adding `sortedby=:count` to the call to `Print.profile` ? Sometimes bottlenecks really "pop out" once when you view the lines sorted by run times. I'm glad to take a look too once you push the changes to `benchmark.jl`.

rgiordan: Pushed.  With lots of sources it's actually worse than with a single source -- about 5x slower.  PTAL and let me know what you see.

jeff-regier: Looks like `combine_sfs!` alone is taking 78 seconds, whereas the whole
elbo with derivs used to take 24 seconds, so that's where I'd start.

```
 78032 ...e/src/SensitiveFloat.jl combine_sfs!
 138
```

rgiordan: Yep.  (I do know how to use the profiler, so what would be helpful is
another pair of eyes looking for ideas to cut out steps.)

rgiordan:
So there's a lot left to do, especially with performance improvements.  Note that the gradients alone are not much worse on a single celestial object, which is our current use case.  There's also still a flaky test that I don't quite understand.  If this passes, though, want to commit and at least be working from this as a starting point to avoid ugly merges?

jeff-regier:
Yeah, sounds good

PR #105, opened 2016-01-29
Hessian benchmark that takes about 30s

rgiordan:
@kpamnany, here's a benchmark script that uses the Hessian and takes about 30s on my machine.

Ryan


kpamnany:
Thanks! How can I increase the runtime? Increase `S`?

rgiordan:
Yep, that's right.

PR #110, opened 2016-02-05
define DualType before the compiler needs it

jeff-regier:
@kpamnany @rgiordan -- Evidently in Julila 0.5 we can't count on assignment of DualType (at run time) happening before the functions that use it are compiled (at  compile time). That seems reasonable. This PR defines DualType at the module's top level.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,6 +1,5 @@
 DataFrames
 Distributions
-DualNumbers
 FITSIO
 ForwardDiff
 JLD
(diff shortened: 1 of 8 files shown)


rgiordan:
LGTM.  Why don't we just strip out all the DualNumbers stuff entirely, though?

jeff-regier:
Yeah that's preferable. I didn't see your email about it being unnecessary
before I worked on this.

jeff-regier:
Hi @rgiordan -- I stripped out the dual numbers stuff. Please review and merge this PR so @kpamnany isn't blocked from running on julia 0.5.

rgiordan:
LGTM

PR #113, opened 2016-02-11
Updates for SloanDigitalSkySurvey v0.0.4

kbarbary:
Updates for rgiordan/SloanDigitalSkySurvey.jl#4. This PR won't pass until that one is merged and a new version tagged.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -30,6 +30,6 @@ for testfile in testfiles
     catch e
         anyerrors = true
         println("\t\033[1m\033[31mFAILED\033[0m: $(testfile)")
-        throw("tests failed")  # Fail fast.
+        rethrow()  # Fail fast.
     end
 end
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -8,11 +8,28 @@ using SampleData
 using Transform
 import Synthetic
 
-include(joinpath(Pkg.dir("Celeste"), "test", "test_misc.jl"))
-include(joinpath(Pkg.dir("Celeste"), "test", "test_kl.jl"))
-include(joinpath(Pkg.dir("Celeste"), "test", "test_sensitive_float.jl"))
-include(joinpath(Pkg.dir("Celeste"), "test", "test_transforms.jl"))
-include(joinpath(Pkg.dir("Celeste"), "test", "test_elbo_values.jl"))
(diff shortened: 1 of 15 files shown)


jeff-regier:
Hi Kyle, Would you rerun Travis (make a trivial change, I guess) now that the SloanDigitalSkySurvey.jl pr is approved, once you've tagged a new version of SloanDigitalSkySurvey.jl?

If any further breaking changes to SloanDigitalSkySurvey.jl are required, please copy that package over to Celeste.jl rather than continuing to use the standalone version of it...it's not a great workflow to have to make PRs in lock step like this, though this works for now.

kbarbary:
OK, I'll tag a version there. Travis can be rerun through their web API without a new commit; will do that also.

I agree that this workflow is suboptimal. I don't foresee any immediate further changes to SloanDigitalSkySurvey.jl though, so we should be OK for the time being.

test/runtests.jl
```diff
@@ -8,11 +8,28 @@ using SampleData
...
+                 "test_elbo_values.jl",
+                 "test_images.jl",
+                 "test_kl.jl",
+                 "test_misc.jl",
+                 "test_optimization.jl",
+                 "test_sensitive_float.jl",
+                 "test_transforms.jl"]
+end
+
+for testfile in testfiles
```
jeff-regier, line 26: Nice idea!

test/test_elbo_values.jl
```diff
@@ -422,7 +421,7 @@ function test_trim_source_tiles()
       fill(1.0, length(loc_ids)), 0.06)
     @test_approx_eq_eps(
       elbo_full.d[non_loc_ids, 1] ./ elbo_trim.d[non_loc_ids, 1],
-      fill(1.0, length(non_loc_ids)), 2e-3)
+      fill(1.0, length(non_loc_ids)), 4e-3)
```
kbarbary, line 424: Quick note: I had to bump the tolerance here to get tests to pass locally. I assumed the 2e-3 wasn't particularly well-motivated and that 4e-3 is still OK. Not sure how this bit of code could have been affected by these WCS-related changes.

test/runtests.jl
kbarbary, line 26: Stole it from the DataFrames.jl tests. :)

test/test_elbo_values.jl
jeff-regier, line 424: That should be fine

PR #121, opened 2016-02-15
Further I/O optimizations

kbarbary:
`load_raw_field()` was by far the biggest performance offender, but here's a grab-bag of additional optimized I/O functions replacing those in `SloanDigitalSkySurvey.SDSS`. Some of these don't actually matter that much for runtime performance, but do help compile time significantly, which is interesting.

`load_sdss_blob()` is now dominated by `PSF.fit_psf_gaussians(raw_psf)`, and most of that is compile time in GaussianMixtures.

### DIFF
--- a/src/SkyImages.jl
+++ b/src/SkyImages.jl
@@ -19,7 +19,11 @@ export break_blob_into_tiles, break_image_into_tiles
 export get_local_sources
 export stitch_object_tiles
 
-const band_letters = ['u', 'g', 'r', 'i', 'z']
+# The default mask planes are those used in the astrometry.net code.
+const DEFAULT_MASK_PLANES = ["S_MASK_INTERP",  # bad pixel (was interpolated)
+                             "S_MASK_SATUR",  # saturated
+                             "S_MASK_CR",  # cosmic ray
+                             "S_MASK_GHOST"]  # electronics artifacts
 
 """
 interp_sky(data, xcoords, ycoords)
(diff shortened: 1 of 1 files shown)


src/SkyImages.jl
```diff
@@ -237,20 +364,30 @@ function load_sdss_blob(field_dir, run_num, camcol_num, field_num;
...
             load_raw_field(field_dir, run_num, camcol_num,
-                           field_num, b, band_gain[b])
+                           field_num, b, gains[b])
+
+
+        letter = band_letters[b]
+        mask_fname = joinpath(field_dir,
+                              "fpM-$run_num-$letter$camcol_num-$field_num.fit")
+
+        print("masking image... ")
```
jeff-regier, line 383: Feel free to remove old print statements if they aren't useful any more

src/SkyImages.jl
```diff
@@ -19,7 +19,11 @@ export break_blob_into_tiles, break_image_into_tiles
 export get_local_sources
 export stitch_object_tiles
```
jeff-regier, line 21: Do we still need `import Grid`? It shows up in the `REQUIRE` file too. Always glad to have fewer dependencies.

jeff-regier:
lgtm

src/SkyImages.jl
kbarbary, line 21: oops, yeah I meant to remove that. I'll do it on master.

It won't immediately change anything since Grid is still pulled in by SloanDigitalSkySurvey.

Issue #116, opened 2016-02-11
script to evaluate the accuracy of one frame's catalog


jeff-regier:
Sample usage:

```
./run_celeste.jl --stage score --run  3900 --camcol 4 --field 829
```

The following scripts shouldn't be necessary once this issue is closed:
- score_cached.jl
- score_sd.jl

We want to get "ground truth" from a coadd run on CasJobs. Sample CasJobs query:

```
SELECT *
INTO MyDB.square_106_4
FROM Stripe82.photoobj
WHERE ra BETWEEN 37.5 AND 38.5 
  AND dec > 0 AND dec < 1
  AND run = 106
  AND camcol = 4
```

But we probably want to filter out astronomical objects with certain flags set, e.g. `dbo.fPhotoFlags('BAD_MOVING_FIT')`.

We also need data from a non-coadd run, as a baseline to compare Celeste too. Run 4263 looks like a good candidate:

```
SELECT *
FROM DR12.photoobj
WHERE ra BETWEEN 37.5 AND 38.5
  AND dec BETWEEN  0 AND 1
  AND camcol = 4
  AND run = 4263
```

jeff-regier:
@kbarbary I committed a simplified and commented version of `score_cached.jl`. (I meant to submit this as a PR but accidentally committed it to master.) The parts you'd need to modify to make it work with full fields rather than stamps are flagged with "TODO". Let me know if you want to meet to talk about it at 3:30 today.

jeff-regier:
@kbarbary  Would you add a unit test that calls `score_field` and `infer`? Maybe add a sample `reffile` to the testing data directory, containing 10 or 100 astronomical objects (just a few KB), for use with this test? Then test that 1) primary approximately matches coadd for each metric, 2) Celeste initialized with primary and optimized with `maxiters=0` also approximately matches coadd.

Also, would you modify the script so that sources that can't be matched get skipped, perhaps with a warning, rather than throwing an error?

Issue #115, opened 2016-02-11
script to infer a catalog for one pre-processed frame


jeff-regier:
Sample usage:

```
./run_celeste.jl --stage infer --run  3900 --camcol 4 --field 829
```

The scripts below shouldn't be necessary once this issue is closed:
- process_image.jl
- infer_and_cache.jl

If pixel-level parallelism works out, we probably don't need `run_celeste.jl` to process multiple tasks at once to close this issue.

PR #122, opened 2016-02-18
WIP: frontend

kbarbary:
This is the start of a main "do-everything" script that will replace most everything else in `bin/`, to address #115 and #116. Not ready to merge, just opening so you can track progress.

Currently, it reads in SDSS files (images/catalog/ancillary files) for a run/camcol/field and outputs one file per source, e.g., `celeste-$run-$camcol-$field-$i.jld` where `$i` is the index of the source.

So far, this is basically a combination of half of `bin/process_image.jl` and most of `bin/staging/preprocess_image.jl`.


kbarbary:
Do we want to keep the option to use "synthetic data" in the main script, or should it go elsewhere?

Similarly, what should be done with the code to "interactively explore results" in `process_image.jl`?

jeff-regier:
Let's omit synthetic data and any other non essential code

kbarbary:
OK, I want to delete `process_image.jl`. Should I just delete all [this code](https://github.com/jeff-regier/Celeste.jl/blob/c263337e9b16d844138a258230dcb8afdf61d1b1/bin/process_image.jl#L103-L159) or is it useful somewhere else?

jeff-regier:
Let's delete whatever isn't needed for the issue. It'll be in a previous
git commit if we need it in the future

bin/celeste.jl
```diff
@@ -0,0 +1,87 @@
...
+
+        # Some sanity checks and improved initialization.
+        objid = mp.objids[i]
+        cat_row = cat_df[:objid] .== objid;
+        is_star = cat_df[cat_row, :is_star][1]
+        mp_s.vp[i][ids.a[1]] = is_star ? 0.8: 0.2
+        mp_s.vp[i][ids.a[2]] = 1.0 - mp_s.vp[i][ids.a][1]
+
+        # Skip dim sources
+        if cat_df[cat_row, :psfflux_r][1] < 10
```
jeff-regier: I think we want to include even the dimmest objects now...maybe ryan or I wrote this if statement for testing?

bin/celeste.jl
```diff
@@ -0,0 +1,87 @@
...
+
+        mp_s = deepcopy(mp);
+        mp_s.active_sources = [i]
+
+        # Some sanity checks and improved initialization.
+        objid = mp.objids[i]
+        cat_row = cat_df[:objid] .== objid;
+        is_star = cat_df[cat_row, :is_star][1]
+        mp_s.vp[i][ids.a[1]] = is_star ? 0.8: 0.2
+        mp_s.vp[i][ids.a[2]] = 1.0 - mp_s.vp[i][ids.a][1]
```
jeff-regier: @rgiordan Are these initializations still helpful? If so, can we move them to ModelInit.jl?

rgiordan: Wow, I am slow to respond.  Yes, these can be moved to `ModelInit.jl`.  We should probably investigate more systematically the seemingly sensitive dependence of the classification on initial conditions.  Per our previous discussion, I believe it to be a serious shortcoming of the taylor expansion of the expected log term.

kbarbary:
This is more-or-less ready.

```
$ celeste.jl --help

Run Celeste.

Usage:
  celeste.jl infer <dir> <run> <camcol> <field> [--outdir=<outdir> --part=<k/n>]
  celeste.jl score <dir> <run> <camcol> <field> <reffile> [--outdir=<outdir>]
  celeste.jl -h | --help
  celeste.jl --version

Options:
  -h, --help         Show this screen.
  --version          Show the version.
  --outdir=<outdir>  Write output files for each source to this directory.
                     Default is to write to input directory. When scoring,
                     this is where celeste result files are read *from*.
  --part=<k/n>       In inferences, split sources into `n` parts and only
                     process the `k`th part.

The `infer` step creates files in the output directory `<outdir>` with
filename format `celeste-<run>-<camcol>-<field>--part-K-M.jld`. The ability
to process only subsets of sources in a run/camcol/field is simply a
low-effort way to parallelize and to run shorter jobs.

In the `infer` step, `<outdir>` is searched for output files matching
`celeste-<run>-<camcol>-<field>-*.jld` and all sources are
concatenated.  A "truth" catalog from `<reffile>`, a FITS file
manually created by running a CasJobs query on the Stripe82 database,
based on RA/Dec (see Celeste README for the query). It is expected
that every Celeste source should have a match in this "truth"
catalog, otherwise an error is thrown. So, we need to use (run,
camcol, field) combinations fall entirely within the RA/Dec patch used
in that query.
```

kbarbary:
Note that I removed the test files in `dat/sample_field` from the repo. Run `make` in that directory to re-download them. I hope the eventually remove the other large files from the repo and then run [BFG](https://rtyley.github.io/bfg-repo-cleaner/) on it to reduce the repo size to something reasonable.

src/score.jl
```diff
@@ -0,0 +1,473 @@
...
+    result[:gal_color_iz] = fluxes_to_color(gal_flux_i, gal_flux_z)
+
+    # gal shape
+    result[:gal_fracdev] = objs[:frac_dev]
+    result[:gal_ab] = where(usedev, objs[:ab_dev], objs[:ab_exp])
+
+    # TODO: the catalog contains both theta_[exp,dev] and phi_[exp,dev]!
+    # which do we use?
+    #result[:gal_angle] = where(usedev, objs[:theta_dev], objs[:theta_exp])
+    result[:gal_angle] = fill(NaN, size(objs, 1))
```
kbarbary, line 195: Not sure what to do here.

src/score.jl
```diff
@@ -0,0 +1,473 @@
...
+
+    # TODO: No scale parameters in catalog.
+    #result[:gal_scale] = where(usedev, objs[?], objs[?])
+    result[:gal_scale] = fill(NaN, size(objs, 1))
```
kbarbary, line 199: Not sure what to do here.

src/score.jl
jeff-regier, line 195: Is `theta_[exp,dev]` scale and `phi_[exp,dev]` angle? That would resolve both ambiguities at once. Otherwise I think we'd have to ask Dustin.

jeff-regier:
This all looks great Kyle. Very readable code.

jeff-regier:
Removing all the large files with BFG sounds like a good idea

jeff-regier:
We need to add `DocOpt` to the REQUIRE file. I noticed that after merging.

jeff-regier:
It takes about 30 seconds on my machine to print out the usage instructions:

```
jeff@arina~/git/Celeste.jl/bin $ date; ./celeste.jl; date 
Mon Feb 29 10:07:46 PST 2016
Usage:
  celeste.jl infer <dir> <run> <camcol> <field> [--outdir=<outdir> --part=<k/n>]
  celeste.jl score <dir> <run> <camcol> <field> <reffile> [--outdir=<outdir>]
  celeste.jl -h | --help
  celeste.jl --version
Mon Feb 29 10:08:15 PST 2016
```

Seems kind of slow, though it's not a big problem. Is precompilation disabled for some of the packages we're using?

jeff-regier:
@kbarbary  Were you able to run this end-to-end for a particular field? What was the run time like? Any scores?

kbarbary:
I just enabled precompilation for FITSIO, for WCS it's now enabled on master, but I haven't tagged a new version yet. Not sure about all the other packages; we've got a lot of dependencies.

On my machine, its 16 seconds:

```
$ time celeste.jl
[...]
real    0m16.306s
user    0m16.280s
sys 0m0.333s
```

Five seconds of this is `using Celeste` (`time julia -e "using Celeste"`), I'm surprised there's an additional 10 second overhead. I'll keep this in mind as something to look into. It would be nice to get this down.

kbarbary:
I downloaded a stripe 82 catalog in the RA/Dec patch (0, 1), (0, 1) and ran on a field in that patch, just for testing on my laptop. I only ran on a few sources to check that the script worked.

jeff-regier:
Looks a couple of minutes per source, or about 1 day per field, single threaded. That's probably workable.

PR #124, opened 2016-02-26
Hessian performance julia4

rgiordan:
The likelihood now allocates very close to zero memory after initialization.  On a single source in the benchmark script, this gives about a 1.4x speedup.  Without the Hessian, it's about about a 2x speedup.  That probably gets this back up to the speed of the original on the gradient-only calculations (though Kyle's changes have broken the pre-Hessian tag and I haven't actually checked yet).

I haven't carefully investigated the issue of scaling, yet.  But I want to make this pull request before my flight to MX leaves and I'm away for the weekend!

Tests pass on my machine, still waiting for Travis.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,6 @@
 
 # omit converage files
 *.cov
+
+# Omit memory profiling output
+**/*.mem
--- a/bin/benchmark_elbo_with_hessian.jl
+++ b/bin/benchmark_elbo_with_hessian.jl
@@ -12,7 +12,7 @@ const dat_dir = joinpath(Pkg.dir("Celeste"), "dat")
 srand(1)
 println("Loading data.")
 
-S = 4
+S = 2
 blob, mp, body, tiled_blob =
   SampleData.gen_n_body_dataset(S, tile_width=10);
 
@@ -23,7 +23,10 @@ println("Calculating ELBO.")
 
 # let's time it without any overhead from profiling
 @time elbo = ElboDeriv.elbo(tiled_blob, mp, calculate_hessian=true);
+@time elbo = ElboDeriv.elbo_likelihood(tiled_blob, mp, calculate_hessian=true);
 
 Profile.init(10^8, 0.001)
-@profile elbo = ElboDeriv.elbo(tiled_blob, mp, calculate_hessian=false);
+Profile.clear_malloc_data()
+#@profile elbo = ElboDeriv.elbo(tiled_blob, mp, calculate_hessian=false);
+@profile elbo = ElboDeriv.elbo_likelihood(tiled_blob, mp, calculate_hessian=false);
 Profile.print(format=:flat)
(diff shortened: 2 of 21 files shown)


jeff-regier:
Hey Ryan, This PR appears to restore the performance with just 1 source per image--the single source run time is 2x faster again. Memory usage is super low now!

The PR doesn't seem to fix the scaling issue---100 sources run time is now 5x slower (124 seconds) than it was (24 seconds), rather than 7x (178 seconds) slower without this PR.

If I increasing the number of sources in the test image from 1 to 10, the runtime scales pretty much linearly. But from 10 to 100, it's superlinear. Once 100 images are crammed in the same image, presumably a lot more "patches" start overlapping---thus activating the bug. So it's probably an algorithmic thing rather than something like type instability or a non-constant global.

rgiordan:
Now that the Hessian is part of the ELBO, you need to allocate memory for a
Hessian whose size scals with the number of sources squared.  It is not
reasonable to expect this to scale the same way as the old code, and it's
not a use case that we'd ever actually need -- we will never be evaluating
a full Hessian for 100 sources simultaneously.  Why should we be targeting
this benchmark?

jeff-regier:
I thought you were just computing the block diagonal of Hessian. Is that
incorrect? If not, why allocate memory for the full Hessian rather than
just the diagonal?

And why allocate memory for the Hessian at all when compute_hessian is set
to false?

Does the trust region code ever to have evaluate the objective or the
derivatives but not the Hessian?

Yeah we can do away with this benchmark if it's no longer representive of
what matters. If we're only ever going to optimize one source at time we
could change the code so optimizing multiple sources at once isn't even an
option.

rgiordan:
> I thought you were just computing the block diagonal of Hessian. Is that
> incorrect?

That is incorrect.  I compute the Hessian of all the active sources.

> And why allocate memory for the Hessian at all when compute_hessian is set
> to false?

Only because there's no real just case and to keep the semantics of
sensitive floats consistent.  If there were a reason to do this it works be
a trivial change.

> Does the trust region code ever to have evaluate the objective or the
> derivatives but not the Hessian?

No.

> Yeah we can do away with this benchmark if it's no longer representive of
> what matters. If we're only ever going to optimize one source at time we
> could change the code so optimizing multiple sources at once isn't even an
> option.

Setting active sources to a single source already accomplishes this.  There
is no reason to force the code to enforce a single source.  If in the
future we really want to get Hessians for lots of sources simultaneously,
we could just encode the SF Hessian as a sparse matrix, probably with
minimal change to existing code.  Right now there is no need for this
functionality, though.

jeff-regier:
When `compute_hessian = false`, in addition to allocating memory for the Hessian, does the code do some processing of the Hessian? Just allocating and zeroing out the a dense Hessian for 100 sources takes only a fraction of a second:

```
julia> @time zeros(100 * 30, 100 * 30);
  0.047650 seconds (157 allocations: 68.675 MB, 116.85% gc time)
```

But we've got like 100 extra seconds of processing to account for, in `benchmark_elbo.jl`. (I get that this isn't such an important benchmark any more, but I'd prefer to understand what's going on.)

rgiordan:
I tried to avoid extra processing, but you're right, something else must be
going on.  As you say, it does not seem to be such an important benchmark,
though.  Want to open a separate issue so we won't forget about this but
can prioritize more important things?

Issue #123, opened 2016-02-24
use all relevant images for each source


jeff-regier:
Modify `celeste.jl` to process just light source specified by an objid. Sample usage:

```
./celeste.jl --stage <infer|score> --objid 1237663784214070312
```

For the `infer` stage, `celeste.jl` should use all the relevant images from any run except for the coadd runs (106 and 206).

rgiordan:
Don't you think we should reference objects by the objid rather than their coordinates?

jeff-regier:
That might be better. Does objid uniquely identify the object, or does the same object get assigned different objids if it appears in multiple runs?

rgiordan:
It gets assigned different objids if it appears on multiple runs.  (But it
gets assigned different locations, too.)  For every part of the sky,
there's a "principal" frame (I think that's what it's called), which is
used to choose the "canonical" objids for that section of the sky.  We
should just use the objids from the principal image as unique identifiers.

We should compile a list of redundantly imaged objects with their locations
and canonical objids from the CASJobs database.  All the information is
there, but AFAIK you have to infer multiply-cataloged objects by looking at
their locations.

jeff-regier:
OK, I updated the description

jeff-regier:
@rgiordan -- Did you know of a way to look up the catalog entry given an objid?  (This came up during the conference call.) The catalog files names contain run-camcol-field, right? Is there a table somewhere that maps objid to run-camcol-field?

rcthomas:
Looking at this right now in fact; we can build this catalog easily just iterating over files on disk.  In the dr8 directory I count about 1e6 FITS files.  I can read about 11 files per second with a quick Python script I put together; this would be about 1 day straight to pull out the objid's.  I can easily parallelize this or do it using a faster API or both.  More info later.

rcthomas:
Here is an update.  Each source has a unique thingId, which may correspond to multiple detections, each with its own objID.  I am building an index from thingId to the list of all associated objIds.  I am doing this using the flat files we have on disk, processing them in parallel, as there isn't enough space under myDB on CasJobs to do this.

kbarbary:
I think it would be better to make a patch of sky (e.g., spanning some range in RA and Dec) the unit, rather than a single source.

Advantages:
- This is a better infrastructure for the future when we might want to:
  - incorporate our own source detection, rather than initializing from a pre-existing catalog
  - fit multiple overlapping sources simultaneously
[...]

PR #127, opened 2016-02-29
remove result binary files & reorganize data

kbarbary:
This does three things:
- Removes the following files:
  
  ```
  delete mode 100644 dat/filter_curves
  delete mode 100644 dat/nm_all_results.jld
  delete mode 100644 dat/nm_results2.jld
  delete mode 100644 dat/parallel_results_1444286953.JLD
  delete mode 100644 dat/parallel_results_1444288483.JLD
  ```
- It moves the prior files to a directory `cfg` (for configuration; I'm open to renames)
- Moves remaining test data (stamps) to `test/data`.

**Still to do:** I'd like to remove the stamp files from the repo as well as part of this PR. Options:
- put them on a webserver and download them (do we have a NERSC project allocation under mantissa?)
- get the script originally used to create them and create them on the fly (I assume dustin created them?)

After this PR is merged, I'll run a cleaner on the repo, after which everyone will have to get a fresh clone.


jeff-regier:
After issue #123 we'll have the ability to process just one source at a
time. How about waiting until then, and then doing away with stamps
entirely? That way we'd avoid having to  support 2 data formats, one of
which that only gets used in tests, and the tests will cover more of the
code base and use data more like the actual data.

kbarbary:
I agree it would be best to remove the stamp stuff entirely. Let's go ahead and merge this and remember to clean the repo of large files after stamps are eventually removed (we want to clean it only once).

jeff-regier:
Looks good! All the Python and R, deps.jl, and the file named PYTHON can go too, at least once we've got a new analysis pipeline running.

PR #129, opened 2016-03-02
Fix init

rgiordan:
@jeff-regier I'm still waiting for the Travis tests, but if they pass, this should fix your problem, so I'm making the PR before going to my next meeting.


dat/.gitignore
```diff
@@ -0,0 +1,3 @@
+**/*.fit
+**/*.fits
+**/*.jld
```
kbarbary: The `dat` dir has been moved to `test/data`.

bin/celeste.jl
```diff
@@ -75,7 +90,7 @@ function infer(dir, run, camcol, field, outdir, partnum, parts)
     # initialize tiled images and model parameters
     tiled_blob, mp = ModelInit.initialize_celeste(images, cat_entries,
                                                   tile_width=TILE_WIDTH,
-                                                  fit_psf=true)
+                                                  fit_psf=true);
```
kbarbary, line 93: curious: why the semicolon?

src/Transform.jl
```diff
@@ -876,5 +877,73 @@ function get_identity_transform(P::Int64, S::Int64)
 end
 
 
+@doc """
+Put the variational parameters within the bounds of the transform.
+""" ->
```
kbarbary: Don't need the `@doc` or `->` here.

dat/.gitignore
rgiordan: Ah, thanks.

bin/celeste.jl
rgiordan, line 93: When debugging, I run these interactively, and it's nice not to have a screenful of output.

src/Transform.jl
rgiordan: Done.

src/Transform.jl
```diff
@@ -875,6 +876,82 @@ function get_identity_transform(P::Int64, S::Int64)
...
+Returns:
+  Updates mp.vp in place.
+"""
+function enforce_bounds!{NumType <: Number}(
+  mp::ModelParams{NumType}, transform::DataTransform)
+
+  @assert mp.S == transform.S
+  @assert length(mp.active_sources) == transform.active_S
+
+	for sa=1:transform.active_S, (param, constraint_vec) in transform.bounds[sa]
```
jeff-regier: looks like a tab snuck in here

src/Transform.jl
```diff
@@ -875,6 +876,82 @@ function get_identity_transform(P::Int64, S::Int64)
...
+        end
+        if sum(mp.vp[s][ids.(param)]) != 1.0
+          println("Warning: $param[$s] is not normalized.")
+          mp.vp[s][ids.(param)] = mp.vp[s][ids.(param)] / sum(mp.vp[s][ids.(param)])
+        end
+      end
+    end
+	end
+end
+
```
jeff-regier, line 953: Why can't we just initialize the variational parameters to values that aren't out of bounds?

rgiordan, line 953: Because the catalogs are read in before the transform is defined.  For example, the bounds for location are per-source, and will depend on which sources are active.  You don't know that when you're reading in the catalog -- in fact, you get the location bounds _from_ the catalog.

rgiordan:
The tabs are fixed.

src/Transform.jl
jeff-regier, line 953: Is `u` the only parameter like that, where the bounds depend on the
catalog? Would it be cleaner to define `u` to be the offset from the
initialization, rather than a parameter with a dynamic bound? I mean we can
just commit this if you prefer, but it's a lot of code.

rgiordan, line 953: No, I think it is cleaner to define the transform after reading in the catalog data.

Anyway, it is not that much code, and much of it is just annoying duplication due to the two different ways that the ids index simplexes.  That  should be fixed, but it doesn't seem like a priority.

PR #131, opened 2016-03-02
Interactive celeste

rgiordan:
@kbarbary , do you mind if I factor the infer function out of the binary for easier debugging / interactive sessions?

Note that it might not be clear from the diff, but I also increased MAX_ITERS to 50 and put the mask planes back in.

Ryan


kbarbary:
I agree: `celeste.jl` should be just a thin wrapper that calls function(s) in the `Celeste` module. However, I'd include `api.jl` from within the Celeste module, then in the script, do `Celeste.infer(...)`.

The `mask_planes=[]` was just left over from a previous existing script (`run_preprocessing.jl` or something), I guess someone had put it in for testing? So that's fine.

Does `bin/celeste_interactive.jl` really need to be in the repo? I thought we were trying to reduce the proliferation of scripts in `bin`... Not a big deal one way or the other though.

rgiordan:
Sounds good to me, I'll include it in the module.

Yes, `mask_planes=[]` was a debugging artifact.

I would really like a place to put interactive scripts that are being used for code development, though I agree maybe `bin` isn't the right place.  How about a folder called `dev`?  @jeff-regier , any opinions?

kbarbary:
A dev directory sounds good to me.

jeff-regier:
Creating a dev directory is fine. The value of sharing dev code is pretty
limited though....it usually takes longer to understand someone else's
scrap code than to write your own. Keeping it in personal repo or a Celeste
fork may be cleaner.

rgiordan:
I'm less interested in sharing than in having a historical cloud backup of development scripts.  I may be lacking some github-fu, but if I develop in a fork and then make a pull request, I'll have to delete my development scripts before making a pull request, right?  That defeats the point of keeping a record of the scripts.  Is there some way to make a PR for some subset of files?  Or have I misunderstood you, Jeff?

jeff-regier:
Maybe creating a dev directory is the way to go then. I don't usually keep
any scripts like that around so I haven't thought through that use case.
The alternative might be to assign versions to different Celeste releases,
and write your scripts to run against a particular version, or to have your
scripts checkout  a particular Celeste commit. Kyle, any thoughts? I don't
really see other projects where they have a dev directory.

kbarbary:
For a larger project with more contributors, it wouldn't make sense to have personal scripts in the repo. Since there's only a few of us though, I think its fine to go ahead and put them in a `dev` directory, with the understanding that we'll delete it at some point later.

rgiordan:
All done, PTAL.  (Note that this PR incorporates a necessary change also in PR https://github.com/jeff-regier/Celeste.jl/pull/132, which you may want to merge first.)

kbarbary:
:+1: for moving dev scripts out of `bin`.

If this PR is merged, I'm pretty sure that github will automatically mark #132 as merged as well, so no need to merge it first. (In fact, doing so would create a new merge commit that would then probably have to be merged into this branch.)

jeff-regier:
Should api.jl be in the dev directory rather than src? Up to you though, I don't want to delay things.

kbarbary:
No, `api.jl` is `include()`d into the Celeste module (as with `score.jl`).

PR #132, opened 2016-03-03
Use vecdot instead of deprecated Optim functions

rgiordan:
PTAL presuming Travis passes.


jeff-regier:
Cool. I haven't used `vecdot` before.

PR #134, opened 2016-03-05
Grab-bag of cosmetic changes

kbarbary:
This PR contains no changes in functionality. Rather, these are changes to better conform to Julia conventions and changes to make the code easier to understand. I was able to clear up a few TODOs in the code along the way. The commit messages are a pretty good summary of the changes, but here's a bit more:

Minor:
- remove `@doc` macro everywhere; not needed on Julia 0.4
- `Int64`->`Int` everywhere. More correct (...if we ever ran on 32-bit) and shorter
- use `function` keyword for multi-line constructors rather than `begin ... end` block.
- merge `benchmark_elbo.jl` and `benchmark_elbo_with_hessian.jl` scripts; they were almost identical.

More significant:
- Write-out `ParamSet` type definitions rather than using metaprogramming to define them. Only slightly longer and much more understandable.
- Use submodules rather than putting `src` directory on the Julia load path. This makes it a little more obvious what namespace things are defined in.
  Files containing submodules are in `CamelCase.jl` whereas files that are just included into other modules are `lowercase_with_underscores.jl`.


kbarbary:
What's the deal with [`forward_diff_model_params()`](https://github.com/jeff-regier/Celeste.jl/blob/207a7bc47fc4938bfcf94a8651074eade4a0f731/src/CelesteTypes.jl#L575-L576)? It doesn't seem to be used anywhere in Celeste, but there's a bunch of tests on it.

jeff-regier:
Looks good Kyle. We might want to search for dead code some time too. @rgiordan will know about `forward_diff_model_params`.

rgiordan:
Thanks for doing all this, @kbarbary !  `forward_diff_model_params` is only for unit tests -- it converts a `ModelParams` object containing floats to one containing ForwardDiff GradientNumbers (or whatever), which allows us to text coded derivatives against autodiff derivatives.

kbarbary:
Should probably move `forward_diff_model_params` to `tests` somewhere then, just to make this clear for the future.

PR #135, opened 2016-03-05
Support julia 0.5

### LINKED ISSUE #118: Migrate to Julia 0.5-dev
That is, add "nightly" to .travis.yml without causing tests to fail.

jeff-regier:
This PR turns on travis testing for Julia nightly, and makes the tests pass. It isn't always pretty:
- `runtests.jl` checks out `Optim.jl` master and `DataFrames.jl` master
- I forked `GaussianProcesses.jl` and fixed just enough of it. `runtests.jl` clones the fork.
- `JLD.jl` doesn't work yet, though it may soon: JuliaLang/JLD.jl#57  This doesn't stop the tests because the JLD stuff is all in bin.

Still, I'd say the PR closes #118

### DIFF
--- a/src/ModelInit.jl
+++ b/src/ModelInit.jl
@@ -464,7 +464,7 @@ function limit_to_object_data(
   # Get the sources that overlap with this object.
   relevant_sources = Int[]
   for b = 1:length(blob), tile_sources in mp.tile_sources[b]
-    if s_original in tile_sources > 0
+    if s_original in tile_sources
       relevant_sources = union(relevant_sources, tile_sources);
     end
   end
--- a/src/elbo_kl.jl
+++ b/src/elbo_kl.jl
@@ -72,9 +72,10 @@ function subtract_kl!{NumType <: Number}(
     mp::ModelParams{NumType}, accum::SensitiveFloat{CanonicalParams, NumType};
     calculate_derivs::Bool=true, calculate_hessian::Bool=true)
 
+
   # The KL divergence as a function of the active source variational parameters.
-  function subtract_kl_value_wrapper{NumType <: Number}(vp_vec::Vector{NumType})
-    elbo_val = zero(NumType)
+  function subtract_kl_value_wrapper{NumType2 <: Number}(vp_vec::Vector{NumType2})
+    elbo_val = zero(NumType2)
     vp_active = reshape(vp_vec, length(CanonicalParams), length(mp.active_sources))
     for sa in 1:length(mp.active_sources)
       vs = vp_active[:, sa]
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,12 +1,5 @@
 #!/usr/bin/env julia
 
-if VERSION > v"0.5.0-dev"
-    Pkg.checkout("Optim")
-    Pkg.checkout("DataFrames")
-    Pkg.rm("GaussianMixtures")
-    Pkg.clone("https://github.com/jeff-regier/GaussianMixtures.jl.git")
-end
-
 using Celeste: Types, SampleData, Transform
 import Celeste.Synthetic
 
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,5 +1,12 @@
 #!/usr/bin/env julia
 
+if VERSION > v"0.5.0-dev"
+    Pkg.checkout("Optim")
+    Pkg.checkout("DataFrames")
+    Pkg.rm("GaussianMixtures")
+    Pkg.clone("https://github.com/jeff-regier/GaussianMixtures.jl.git")
+end
+
 using Celeste: Types, SampleData, Transform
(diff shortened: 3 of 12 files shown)


src/ModelInit.jl
```diff
@@ -464,7 +464,7 @@ function limit_to_object_data(
   # Get the sources that overlap with this object.
   relevant_sources = Int[]
   for b = 1:length(blob), tile_sources in mp.tile_sources[b]
-    if s_original in tile_sources > 0
```
jeff-regier, line 467: We probably didn't mean to compare a boolean to an int before, but for whatever reason 0.4 doesn't complain, while 0.5 stops.

src/elbo_kl.jl
```diff
@@ -72,9 +72,10 @@ function subtract_kl!{NumType <: Number}(
     mp::ModelParams{NumType}, accum::SensitiveFloat{CanonicalParams, NumType};
     calculate_derivs::Bool=true, calculate_hessian::Bool=true)
 
+
   # The KL divergence as a function of the active source variational parameters.
-  function subtract_kl_value_wrapper{NumType <: Number}(vp_vec::Vector{NumType})
-    elbo_val = zero(NumType)
+  function subtract_kl_value_wrapper{NumType2 <: Number}(vp_vec::Vector{NumType2})
```
jeff-regier, line 77: This was causing a seg fault in 0.5, rather than an error

test/test_derivatives.jl
```diff
@@ -16,6 +16,9 @@ else
     end
 end
 
+const datadir = joinpath(Pkg.dir("Celeste"), "test", "data")
```
jeff-regier: I might be redefining a constant here, but it's nice to run just one file or tests without having to run all the tests. (Is there a better way?)

kbarbary: Yes, just remove this. `datadir` is already defined in `runtests.jl`. All the `test_*.jl` files are included into that file, so they'll be in the same namespace/scope as `datadir` already: including just a subset of the `test_*.jl` files doesn't affect this.

test/test_images.jl
```diff
@@ -4,12 +4,14 @@ using DataFrames
...
-import Celeste: ModelInit, SkyImages, ElboDeriv
+import Celeste: ModelInit, SkyImages, ElboDeriv, Synthetic
 
 
 println("Running SkyImages tests.")
 
-field_dir = joinpath(dat_dir, "sample_field")
+const datadir = joinpath(Pkg.dir("Celeste"), "test", "data")
+
+field_dir = joinpath(datadir, "sample_field")
```
kbarbary: There's no `sample_field` subdirectory in `test/data` anymore; the files are all in `test/data`.

test/test_transforms.jl
```diff
@@ -5,7 +5,7 @@ using Base.Test
 
 using Celeste: Types, SampleData, Transform, SensitiveFloats
 import Celeste: ModelInit, ElboDeriv
-
+using Compat
```
kbarbary, line 8: Sorry for  removing all the `using Compat`s in #134! I figured most were for 0.3->0.4 changes and thus no longer needed.

test/runtests.jl
```diff
@@ -1,5 +1,12 @@
 #!/usr/bin/env julia
 
+if VERSION > v"0.5.0-dev"
+    Pkg.checkout("Optim")
+    Pkg.checkout("DataFrames")
+    Pkg.rm("GaussianMixtures")
+    Pkg.clone("https://github.com/jeff-regier/GaussianMixtures.jl.git")
```
kbarbary: These run-time changes to which packages are installed seem like a bad idea. Better to document in the README the requirements for running on v0.5.

kbarbary: (and for travis, add these commands to `.travis.yml`)

test/test_derivatives.jl
jeff-regier: Oh I see, I can call runtests with ARGS now.

test/test_images.jl
jeff-regier: It's weird we have both `dat_dir` defined in SampleData.jl and `datadir` defined `runtests.jl`. SampleData.jl probably shouldn't know about the test data's location.

test/test_transforms.jl
jeff-regier, line 8: Removing them before seemed like a good thing to do...I didn't realize how soon we'd need them again

test/test_images.jl
kbarbary: > It's weird we have both dat_dir defined in SampleData.jl and datadir
> defined runtests.jl. SampleData.jl probably shouldn't know about the test
> data's location.

I agree. Probably SampleData.jl should eventually be moved to tests, as its
not necessary for the core functionality in the app. This is something to
think about after SC though!

Issue #125, opened 2016-02-26
Understand benchmark_elbo.jl run time


jeff-regier:
We expect it to take ~25 seconds, but after the Hessian commit, it takes around ~125 seconds. Because performance with `S=1` matters most now, this issue isn't too pressing. But it would be good to understand what's happening.

kbarbary:
I've been studying the code in ElboDeriv.jl and included files and I have a bit of unsolicited Julia advice. Feel free to ignore if you already know this, or if I'm forgetting something obvious here!

Most of the code seems to be written as functions that update some object, in order to avoid allocating memory. However, in some cases, the code can be rewritten in a more purely functional style without allocating any memory. Here's an example:

``` julia
function eval_bvn_pdf_in_place!{NumType <: Number}(
    elbo_vars::ElboIntermediateVariables{NumType},
    bmc::BvnComponent{NumType}, x::Vector{Float64})

  elbo_vars.bvn_derivs.py1[1] =
    bmc.precision[1,1] * (x[1] - bmc.the_mean[1]) +
    bmc.precision[1,2] * (x[2] - bmc.the_mean[2])
  elbo_vars.bvn_derivs.py2[1] =
    bmc.precision[2,1] * (x[1] - bmc.the_mean[1]) +
    bmc.precision[2,2] * (x[2] - bmc.the_mean[2])
  elbo_vars.bvn_derivs.f_pre[1] =
    bmc.z * exp(-0.5 * ((x[1] - bmc.the_mean[1]) * elbo_vars.bvn_derivs.py1[1] +
                        (x[2] - bmc.the_mean[2]) * elbo_vars.bvn_derivs.py2[1]))

  true # return type
end
```

This function updates 3 numbers in `elbo_vars.bvn_derivs`. It could be rewritten as a non-updating function that simply returns a 3-tuple:

``` julia
function pdf{T <: Number}(bmc::BvnComponent{T}, x::Vector{Float64})
    a = (bmc.precision[1,1] * (x[1] - bmc.the_mean[1]) +
         bmc.precision[1,2] * (x[2] - bmc.the_mean[2]))
    b = (bmc.precision[2,1] * (x[1] - bmc.the_mean[1]) +
         bmc.precision[2,2] * (x[2] - bmc.the_mean[2]))
    c = bmc.z * exp(-0.5 * ((x[1] - bmc.the_mean[1]) * a +
                            (x[2] - bmc.the_mean[2]) * b))
    return a, b, c
end
[...]

PR #137, opened 2016-03-07
Factor out bnv derivs

rgiordan:
This is a first step towards fitting the PSF in Celeste.  I'd like to submit the PR now to avoid nasty merges while I finish the rest.  Tests pass on my machine, but currently waiting on Travis.


jeff-regier:
Have/will you re-run the benchmark script to make sure the changes haven't slowed down the objective function evaluations?

src/bivariate_normals.jl
```diff
@@ -18,9 +20,9 @@ immutable BvnComponent{NumType <: Number}
     z::NumType
     dsiginv_dsig::Matrix{NumType}
 
-    function BvnComponent{T1 <: Number, T2 <: Number, T3 <: Number}(
+    BvnComponent{T1 <: Number, T2 <: Number, T3 <: Number}(
         the_mean::Vector{T1}, the_cov::Matrix{T2}, weight::T3;
-        calculate_siginv_deriv::Bool=true)
+        calculate_siginv_deriv::Bool=true) = begin
```
kbarbary, line 25: Let's use `function` syntax everywhere for functions rather than begin/end blocks.

kbarbary:
The file `SensitiveFloat.jl` is a duplicate of `SensitiveFloats.jl`. The rename to plural was done in #134 -- necessary because `SensitiveFloats` is now a module and has to be plural to avoid name collision with the type `SensitiveFloat`. (This sort of pluralization is typical in Julia module names for this reason.)

src/SensitiveFloat.jl
```diff
@@ -0,0 +1,275 @@
...
+      in a local_P x local_S matrix.
+  h:  The second derivative with respect to each variational parameter,
+      in the same format as d.  This is used for the full Hessian
+      with respect to all the sources.
+  hs: An array of per-source Hessians.  This will generally be reserved
+      for the Hessian of brightness values that depend only on one source.
+""" ->
+type SensitiveFloat{ParamType <: CelesteTypes.ParamSet, NumType <: Number}
+    # Actually a single value, but an Array to avoid memory allocation
+    v::Vector{NumType}
```
kbarbary, line 20: While we're looking at SensitiveFloat, I wanted to ask why `v` is an array instead of just `v::Numtype` here. The effect of using an array is opposite of what the comment implies.

rgiordan, line 20: I found when looking at memory allocation for the full Hessian that memory
was allocated each time the value of a SensitiveFloat was changed and (in
some cases at least) even accessed.  I don't know enough about the julia
compiler to account for this, but when I switched to using an array for the
value the memory allocation disappeared.  So when I say "to avoid memory
allocation" I mean "to avoid memory allocation whenever the value is used
or changed" rather than to avoid memory allocation up front.

src/bivariate_normals.jl
rgiordan, line 25: Agreed.  Sorry about this, I just missed this in the merge.

src/SensitiveFloat.jl
jeff-regier, line 20: Any luck duplicating the phenomenon---memory allocation on access---in a minimal working example?

rgiordan, line 20: I didn't try.  Do you think it's important?

jeff-regier, line 20: It might turn out the be nothing, but it's kind of unsettling. Could it be related to #125 ?

kbarbary, line 20: Typically, if there's no type instability, modifying a bitstype field like Float64 should not allocate. Example:

``` julia
julia> type A
       val::Float64
       end

julia> function update!(a)
           for i=1:10_000_000
               a.val = rand()
           end
       end

julia> a = A(0.0);

julia> @time update!(a)
  0.048819 seconds (19.14 k allocations: 826.487 KB)

julia> @time update!(a)
  0.039082 seconds (4 allocations: 160 bytes)
```

I don't know whether this is absolutely always true though.

rgiordan, line 20: The problem occurred in accumulate_source_brightness!.  I also recall that
before I refactored ElboDeriv, there was a construction where we defined

fsm = (fsm[1], fsm[2])

and then referred to

fsm[i].v

Is it possible that the construction of this tuple was causing some type
instability that led to this memory allocation?  I do think that not every
reference to a SF's value caused allocation.  This is interesting if
someone wants to try switching back and looking at memory allcoation, but I
do think there are bigger fish to fry before the SC deadline.

kbarbary, line 20: Yes, if `fsm` is a Vector before the assignment

```
fsm = (fsm[1], fsm[2])
```

then you'll have a type instability.

kbarbary:
@rgiordan Just checking: are you going to remove SensitiveFloat.jl and make the other minor changes as part of the next PR finishing the work?

BTW `git rebase` can be helpful for situations like this. Rather than submitting a half-done PR to merge, run `git rebase [remote]/master` locally - you still have to resolve conflicts like merge, but it makes it look like your branch started from the current commit of `[remote]/master`, making later merging easier.

rgiordan:
@kbarbary , yes, I'll fix them.  Thanks for catching it.  (I think I'll leave v as an array for the moment if you don't mind, though.)

Thank you for the pointer about rebasing.  I am learning a lot on this project these days!

kbarbary:
Yeah I agree it's not the time to fiddle with `v`.

PR #141, opened 2016-03-09
Minor Cleanup

rgiordan:
@kbarbary , PTAL while the tests run?

I changed `benchmark_elbo` to run the representative test rather than the non-understood but not currently relevant scaling problem.  IIRC this actually caused some confusion recently.


kbarbary:
Looks good to me, thanks Ryan.

Issue #143, opened 2016-03-10
Make multi-object initialization less redundant


rgiordan:
Right now, in order to efficiently run on one object at a time, we need run `initialize_celeste` twice -- once to get the tile_sources, and again to fit just the psfs that we're interested in.  This is unnecessary, and some minor tidying should be able to eliminate this redundancy.

rgiordan:
Fixed in https://github.com/jeff-regier/Celeste.jl/pull/164

PR #144, opened 2016-03-10
Convergence problems

### LINKED ISSUE #130: Make sure all the necessary sources are included
@kbarbary , the way `celeste.jl` divides up the celestial objects isn't correct -- you can't partition the catalog entries by slicing the array.  If two objects are in the same tile, then you need to make sure both the objects are in the model params object.

One correct way to do it is something like the block of code in `limit_to_object_data` where I populate the `relevant_sources` vector.

rgiordan:
@kbarbary , @jeff-regier PTAL.  This fixes https://github.com/jeff-regier/Celeste.jl/issues/130 (and opens https://github.com/jeff-regier/Celeste.jl/issues/143)

### DIFF
--- a/dev/celeste_interactive.jl
+++ b/dev/celeste_interactive.jl
@@ -8,6 +8,8 @@ using PyPlot
 import FITSIO
 import JLD
 import SloanDigitalSkySurvey: SDSS
+import SloanDigitalSkySurvey: WCSUtils
+
 
 using Celeste.Types
 import Celeste.SkyImages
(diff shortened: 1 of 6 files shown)


rgiordan:
NB: 0.4 tests pass, 0.5 tests fail (as usual)

kbarbary:
Looks good to me.

Given that we're moving towards just single objects in #123, I think it's fine to get rid of all the `part` stuff and `cat_entries[minidx:maxidx]` if you want to go ahead and do that here.

rgiordan:
If it's ok with you, I'll let you and Rollin do this since the two of you
are deciding together on the API.

kbarbary:
Makes sense.

PR #149, opened 2016-03-14
RA/Dec-based script and NERSC-specific API

kbarbary:
This PR does two things:
1. `Celeste.infer()` runs all sources in a given patch of sky (by ra/dec range) for the provided fields.
2. The command line script `bin/celeste.jl` is now NERSC-specific, with hard-coded paths. It is run like:
   
   ```
   celeste.jl infer-nersc 164.3 164.4 39.6 39.7
   ```
   
   where the numbers are the min and max RA and min and max Dec. The hard-coded paths are obviously not ideal in the long run, but are a convenience for us while doing test runs. (The relevant files are scattered over the file system, so passing them on the command line would be quite unwieldy.)

There's a lot of reasons that I found to favor a spatially-based rather than source-based API like this. Rather than outlining them here, it will probably be more efficient to discuss in person. If we later decide to go back to source-based, that is easily done.

There's a lot of code churn here due to reorganizations -- I apologize for that. If anyone actually reviews this in detail, I recommend just looking at the changed files on my branch, rather that the diff. The main reorganization is that SDSS-related functions that can be considered not specific to Celeste have been moved into a new submodule `SDSSIO`, with minimal non-Celeste dependencies (FITSIO, WCS, HTTPClient). In the process, I renamed the `RawPSFComponents` to `SDSSPSF` beacause it is the SDSS-specific PSF representation: `SDSSIO.read_psf()` returns an `SDSSPSF` instance, which seems to make sense.

I haven't actually gotten a chance to test the NERSC script at NERSC due to dependency issues there, but the Celeste tests themselves should pass, and I wanted to get this merged before we meet on Tuesday (I won't have much time Monday).
### Next steps: update `score`

The next step is to update the score function to work with the new output files. (The spatial nature of this new API makes this much easier, since the master truth catalog is based on a spatial query.)


kbarbary:
I'm going to update this to use a pre-built flat file rather than HTTPClient for the RA/Dec field query.

src/api.jl
```diff
@@ -12,114 +12,259 @@ import .OptimizeElbo
...
-  cat_entries_s = cat_entries[relevant_sources];
-  tiled_blob, mp = ModelInit.initialize_celeste(images, cat_entries_s,
-                                                tile_width=20,
-                                                fit_psf=true);
-  active_s = findfirst(mp.objids, objid)
-  mp.active_sources = [ active_s ]
+"""
+Query the SDSS database for all fields that overlap the given RA, Dec range.
+"""
+function query_overlapping_fields(ramin, ramax, decmin, decmax)
```
rgiordan, line 21: No need to change it in this commit, but IMO it would make more sense for api.jl to be a thin wrapper around substantive functionality that's available in modules.  This function is a good example of many currently in api.jl that seem more like useful functions than APIs.

jeff-regier:
Looks like SampleData.jl may not have the right convert method in its scope:

```
Running hessian tests.
    FAILED: test_derivatives.jl
ERROR: LoadError: LoadError: MethodError: `convert` has no method matching convert(::Type{Celeste.Types.CatalogEntry}, ::Array{Float64,1}, ::Bool, ::Array{Float64,1}, ::Array{Float64,1}, ::Float64, ::Float64, ::Float64, ::Float64, ::ASCIIString)
This may have arisen from a call to the constructor Celeste.Types.CatalogEntry(...),
since type constructors fall back to convert methods.
Closest candidates are:
  Celeste.Types.CatalogEntry(::Array{Float64,1}, ::Bool, ::Array{Float64,1}, ::Array{Float64,1}, ::Float64, ::Float64, ::Float64, ::Float64, ::ASCIIString, !Matched::Int64)
  Celeste.Types.CatalogEntry(::Any, ::Any, ::Any, ::Any, ::Any, ::Any, ::Any, ::Any, ::Any, !Matched::Any)
  call{T}(::Type{T}, ::Any)
  ...
 in call at essentials.jl:57
 in gen_sample_star_dataset at /home/travis/.julia/v0.4/Celeste/src/SampleData.jl:63
 in test_brightness_hessian at /home/travis/.julia/v0.4/Celeste/test/test_derivatives.jl:886
 in include at ./boot.jl:261
 in include_from_node1 at ./loading.jl:304
 [inlined code] from /home/travis/.julia/v0.4/Celeste/test/runtests.jl:35
 in anonymous at no file:0
 in include at ./boot.jl:261
 in include_from_node1 at ./loading.jl:304
 in process_options at ./client.jl:280
 in _start at ./client.jl:378
while loading /home/travis/.julia/v0.4/Celeste/test/test_derivatives.jl, in expression starting on line 962
while loading /home/travis/.julia/v0.4/Celeste/test/runtests.jl, in expression starting on line 33
```

src/api.jl
```diff
@@ -12,114 +12,259 @@ import .OptimizeElbo
...
+    # because for each object there should only be one "primary" occurance.)
+    rawcatalog = deepcopy(rawcatalogs[1])
+    for i=2:length(rawcatalogs)
+        for key in keys(rawcatalog)
+            append!(rawcatalog[key], rawcatalogs[i][key])
+        end
+    end
+
+    # Limit catalog to ra, dec range.
+    mask = ((rawcatalog["ra"] .> ra_range[1]) &
```
rgiordan: If there is a celestial object near the edge of an [RA, DEC] rectangle that contributes photons to an image but is not included in this rectangle, I think this would exclude it from inference, right?  That will lead to problems similar to what I described in https://github.com/jeff-regier/Celeste.jl/issues/130.

src/api.jl
```diff
@@ -12,114 +12,259 @@ import .OptimizeElbo
...
+               photofield_dirs=frame_dirs,
+               photoobj_dirs=frame_dirs)
+
+    # the code below assumes there is at least one field.
+    if length(fieldids) == 0
+        return Dict{Int, Dict}()
+    end
+
+    # Read in all photoobj catalogs.
+    rawcatalogs = Array(Dict, length(fieldids))
```
rgiordan, line 102: All this code looks like something you might want to run outside infer (e.g. in interactive mode for exploration or debugging).  Could you wrap it in a function rather than including it in infer()?

src/SDSSIO.jl
```diff
@@ -0,0 +1,467 @@
...
+
+function get_query_url(data_release)
+    suffix = (data_release < 11)? "sql.asp": "x_sql.aspx"
+    return string(SKYSERVER_BASEURL, "/dr$(data_release)/en/tools/search/",
+                  suffix)
+end
+
+
+"""Remove comments and newlines from SQL statement."""
+function sanitize_query(stmt::AbstractString)
```
rgiordan: This should be unit tested.

src/api.jl
```diff
@@ -12,114 +12,259 @@ import .OptimizeElbo
...
+    end
+
+    for i in eachindex(fieldids)
+        info(string("field ", i, ": ", length(rawcatalogs[i]["objid"]),
+                    " primary entries"))
+    end
+
+    # Merge all catalogs together (there should be no duplicate objects,
+    # because for each object there should only be one "primary" occurance.)
+    rawcatalog = deepcopy(rawcatalogs[1])
```
rgiordan, line 131: A lot of this code really should be factored out and unit tested IMO.

rgiordan:
I'd like to take a look at this in detail before merging, but I think I'm not going to get a chance to pull the branch and look at the diff tonight.  @kbarbary , @jeff-regier , can we wait for me to get to it tomorrow morning?

Kyle, maybe we should talk in person about objid vs RA / DEC windows.  I'm open to being convinced but and still not convinced.  In particular, I'm worried about how this script will handle source near but not in the RA / DEC rectangle.

kbarbary:
@rgiordan I need to update this PR regarding the HTTP query (which will be removed) and fixing the failing tests. In the process, I'll try to address the comments so far. Maybe wait until then for further review.

We can definitely discuss partitioning by thing_id vs RA/Dec before merging. Maybe we can resolve this first thing on Tuesday.

src/SDSSIO.jl
kbarbary: I'm going to remove this for the time being.

src/api.jl
kbarbary, line 21: Agreed. They're sitting in API for the time being because it wasn't clear which submodule they belong in. Many of the score-related function may be refactored in future work though.

jeff-regier:
If possible, let's merge this before meeting tomorrow, so we don't have to deal with difficult merges then. Once the unit tests are passing this is clearly an improvement over what we have.

rgiordan:
I have to head to BSTARS now.  It would be helpful if Kyle would give me an
extra ping when unit tests are passing.

rgiordan:
I see the tests are still failling.  Any update, @kbarbary ?

kbarbary:
I've been away from my laptop all day but am just starting to look at this now. The failing test is a trivial fix, but I'm also going to try to address some other issues tonight. We should be able to merge tomorrow AM at the meeting.

rgiordan:
Sounds good to me!  Thanks for the update.

kbarbary:
5e2e7ca should address Ryan's concerns about not including all relevant sources.

Mostly as a reminder to myself, I compiled a list of arguments in favor of this RA/Dec-based API:
- I/O (or data transfer) cost can be amortized over many adjacent sources
  in the same patch.
- Opens another level of parallelism: a node can load all data for a
  sky patch, each thread can process a source in that patch. Otherwise
  all threads would work on a single source then get more data.
- Ensures that all images containing a source are used, not just the
  images where the source is in the catalog. Counter-argument: _we're
  cutting out dim objects, so the sources of interest should be
  detected in all images._ Counter-counter argument: It is unclear
  that this is actually the case.  Better to have a single known cut
  on brightness rather than also cutting images based on source
  detectability in each image (at an unknown level).
- Catalog initialization: Loading all relevant images ensures that we get the
  "primary" detection for each source. If we don't load all images in the
  patch, there will be nearby objects for which the "primary" detection image
  is not loaded. This results in either: (1) we can't use "primary" to
  initialize or (2) we miss sources that we are actually interested in
  including.
- If we care about corner cases where nearby objects are off the image but
  still have relevant flux contributions, we can easily load images & catalogs
  for a slightly larger area than the one we're fitting in.
- Scoring: The "truth" catalog from Stripe 82 is generated based on
  RA/Dec, and we will be scoring a subset of Stripe 82. If we run
  Celeste based on thing_id, it will be more complicated to run the
  subset that we want to score: we'd have to first run a query for
  sources in the patch of interest and run just those. Then
  afterwards, we'd have to again do that query when figuring out which
  results on disk to score (assuming we run a larger area than we're scoring).
- We don't need 260 GB of lookup index files. (I'm assuming these are not easy to create).
- Single-source inference can still be done by getting the position of
  the source to fit, and running on a very small patch around that source.
  (This would require a source->RA/Dec lookup table similar to the one Rollin
  built for source->fields.)

PR #152, opened 2016-03-15
Remove GaussianMixtures from REQUIRE

rgiordan:


rgiordan:
Travis has actually passed, but the pull request screen hasn't updated yet.

PR #160, opened 2016-03-16
3x speedup in initialize_celeste

jeff-regier:


rgiordan:
@kbarbary , @rcthomas , PTAL, since this changes the arguments to `get_tiled_image_sources`.

PR #162, opened 2016-03-17
refactor get_tiled_image_sources()

kbarbary:
`get_tiled_image_sources()` now takes patch centers and radii in pixel
coordinates rather than patches themselves. This will make it possible in
the future to figure out relevant sources (which requires
`get_tiled_image_sources`) without constructing `SkyPatch` objects first.

Along the way, the two functions `local_source_candidates()` and
`get_local_sources()` were moved from SkyImages -> ModelInit, because they
are only used by `get_tiled_image_sources()` and that lives in ModelInit.

Also along the way, `ModelParams.tile_sources` now has a concrete type,
which should increase performance.


src/Types.jl
```diff
@@ -554,7 +554,7 @@ type ModelParams{T <: Number}
     vp::VariationalParams{T}
     pp::PriorParams
     patches::Array{SkyPatch, 2}
-    tile_sources::Vector{Array{Array{Int}}}
+    tile_sources::Vector{Array{Vector{Int}, 2}}
```
kbarbary, line 557: Hopefully we should see a performance bump from this being a concrete type.

src/SkyImages.jl
```diff
@@ -449,86 +449,6 @@ world_to_pix{T <: Number}(patch::SkyPatch, world_loc::Vector{T}) =
...
-    of a point in the tile, though it might not.
-"""
-function local_source_candidates(
-    tiles::TiledImage, patches::Vector{SkyPatch})
-
-  # Get the largest size of the pixel ellipse defined by the patch
-  # world coordinate circle.
-  patch_pixel_radii =
-    Float64[patches[s].radius * maximum(abs(eig(patches[s].wcs_jacobian)[1]))
-            for s=1:length(patches)];
```
kbarbary, line 470: In the new version of `local_source_candidates` in `ModelInit`, we no longer call `eig`, so hopefully this should be a performance bump.

src/Types.jl
jeff-regier, line 557: That'd be great. I'm going to try profiling the unit tests to see what else is taking so long. (And it's concerning that unit test on julia 0.5 take like 50% long than julia 0.4)

src/SkyImages.jl
```diff
@@ -449,86 +449,6 @@ world_to_pix{T <: Number}(patch::SkyPatch, world_loc::Vector{T}) =
...
-Args:
-  - tiles: A TiledImage
-  - patches: A vector of patches (e.g. for a particular band)
-
-Returns:
-  - An array (over tiles) of a vector of candidate
-    source patches.  If a patch is a candidate, it may be within the patch radius
-    of a point in the tile, though it might not.
-"""
-function local_source_candidates(
```
jeff-regier, line 463: We may want to get rid of `SkyImages.jl` entirely...most of what's left seems to belong in `ModelInit.jl` or one of the files that we're going to merge into Celeste from the `SloanDigitalSkySurvey` package.

PR #164, opened 2016-03-18
Only fit psf for relevant sources

kbarbary:
@rgiordan should review the second commit 5b88c06 to make sure it still does what its supposed to.

The basic idea here is that we:
- initialize model parameters without fitting the PSF (same as before)
- compile a list of all relevant sources (`get_all_relevant_sources(mp)`)
- for each relevant source, replace the corresponding `SkyPatch` in `mp` with a new one that has a fitted PSF. 

This includes #163, which didn't actually turn out to be too relevant.

### DIFF
--- a/src/ModelInit.jl
+++ b/src/ModelInit.jl
@@ -242,6 +242,13 @@ function SkyPatch(ce::CatalogEntry, img::Image; fit_psf=true,
              scale_patch_size * pix_radius, psf, wcs_jacobian, pixel_center)
 end
 
+"""
+Initialize a SkyPatch from an existing SkyPatch and a new PSF.
+"""
+SkyPatch(patch::SkyPatch, psf::Vector{PsfComponent}) =
+    SkyPatch(patch.center, patch.radius, patch.radius_pix,
+             psf, patch.wcs_jacobian, patch.pixel_center)
+
 
 """
 A fast function to determine which sources might belong to which tiles.
@@ -465,6 +472,15 @@ function get_relevant_sources{NumType <: Number}(
   relevant_sources
 end
 
+function get_all_relevant_sources{NumType <: Number}(
+    mp::ModelParams{NumType}, idx::Vector{Int})
+    out = Int[]
+    for i in idx
+        out = union(out, get_relevant_sources(mp, i))
+    end
+    return out
+end
+
 
 """
 Return a reduced Celeste dataset useful for a single object.
@@ -630,6 +646,5 @@ function initialize_objid(objid::ASCIIString, mp_all::ModelParams{Float64},
                                            noise_fraction=0.1)
   return trimmed_tiled_images, mp, active_s, s
 end
-
 
 end # module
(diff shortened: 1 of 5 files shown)


rgiordan:
Yes, I will look at it tomorrow morning!  Sorry, my Thursday evenings are
always away from my computer.

src/ModelInit.jl
```diff
@@ -468,6 +472,15 @@ function get_relevant_sources{NumType <: Number}(
   relevant_sources
 end
```
rgiordan, line 472: I know this function is simple, but please add function documentation saying at least what the input and output is.

src/api.jl
```diff
@@ -140,18 +140,34 @@ function infer(ra_range::Tuple{Float64, Float64},
...
+    for j in 1:size(mp.patches, 2)  # loop over images
+        for i in relevant_idx  # loop over relevant sources
+            patch = mp.patches[i, j]
+            psf = SkyImages.get_source_psf(patch.center, images[j])
+            mp.patches[i, j] = ModelInit.SkyPatch(patch, psf)
+        end
+    end
 
     results = Dict{Int, Dict}()
     for i in idx
```
rgiordan: Just FYI, elsewhere in the code I have (tried to) consistently used "s" for indexing into sources and "sa" for indexing into active sources.  This improves readability a lot IMO.  You don't have to change it for this PR, but maybe consider it in the future.

Remind me to tell you sometime about a major bug at Google that went undetected for months because of non-descriptive index variable names.

src/Types.jl
```diff
@@ -336,6 +336,7 @@ Attributes:
 immutable SkyPatch
     center::Vector{Float64}
     radius::Float64
```
rgiordan: I don't think the world radius is used anywhere in the code after this change.  Can you delete it?

rgiordan:
This looks good to pull.  Thanks, Kyle!  After this is all done we should clean up this whole initialization process (it's my fault it's a mess), but this will be good for now.

src/ModelInit.jl
kbarbary, line 472: Will do. I think I just forgot the docstring rather than intentionally omitting it.

src/Types.jl
kbarbary: Yes, will do. Similarly, `initialize_objid` is no longer used. Can we delete it?

rgiordan: I'm in a meeting so I can't look at the code.  But before deleting it could
you make sure something similar is factored out of infer() so that I can
use it interactively?

rgiordan: If you've done that then go ahead and delete it.

kbarbary:
Comments addressed. In the interest of consistency I've changed source index to `s`. I've left `initialize_objid` as is, since it no longer has an analog in `infer()` and so it isn't obvious how to factor it out. Do what you will with it in future work.

rgiordan:
SGTM, go ahead and merge if you like, since I'm in a meeting.

Issue #168, opened 2016-03-20
PosDefException from neton_trust_region.jl


jeff-regier:
```
jregier@cori10~/Celeste/bin $ ./celeste.jl infer-nersc 164.39 164.5 39.11 39.5 .
...
20-Mar 15:47:47:INFO:root:3925 primary sources
20-Mar 15:47:47:INFO:root:2448 primary sources after MIN_FLUX cut
20-Mar 15:47:47:INFO:root:processing 512 sources in RA/Dec range
20-Mar 15:47:47:INFO:root:reading field (3840,1,69)
20-Mar 15:47:50:INFO:root:reading field (3840,1,70)
20-Mar 15:47:52:INFO:root:reading field (3900,6,269)
20-Mar 15:47:53:INFO:root:reading field (3900,6,270)
20-Mar 15:47:54:INFO:root:reading field (4469,6,343)
20-Mar 15:47:56:INFO:root:reading field (4469,6,344)
20-Mar 15:47:57:INFO:root:initializing celeste without PSF fit
Getting VP from sources.

20-Mar 15:48:41:INFO:root:fitting PSF for all relevant sources
ERROR: LoadError: Base.LinAlg.PosDefException(10)
 in chol! at linalg/cholesky.jl:28
 in solve_tr_subproblem! at /global/homes/j/jregier/.julia/v0.4/Celeste/src/newton_trust_region.jl:208
 in __newton_tr#21__ at /global/homes/j/jregier/.julia/v0.4/Celeste/src/newton_trust_region.jl:326
 in fit_psf at /global/homes/j/jregier/.julia/v0.4/Celeste/src/PSF.jl:121
 in fit_raw_psf_for_celeste at /global/homes/j/jregier/.julia/v0.4/Celeste/src/PSF.jl:702
 in get_source_psf at /global/homes/j/jregier/.julia/v0.4/Celeste/src/PSF.jl:221
 in fit_object_psfs! at /global/homes/j/jregier/.julia/v0.4/Celeste/src/ModelInit.jl:282
 in infer at /global/homes/j/jregier/.julia/v0.4/Celeste/src/api.jl:155
 in infer_nersc at /global/homes/j/jregier/.julia/v0.4/Celeste/src/api.jl:332
 in main at /global/u2/j/jregier/Celeste/bin/celeste.jl:36
 in include at ./boot.jl:261
 in include_from_node1 at ./loading.jl:304
 in process_options at ./client.jl:280
[...]

PR #174, opened 2016-03-22
calculate relevant sources by image for PSF fitting

kbarbary:
Addresses #168 but I haven't written a test yet. Any advice on what to test or prior art in the existing tests?

### DIFF
--- a/src/ModelInit.jl
+++ b/src/ModelInit.jl
@@ -264,13 +264,15 @@ Returns:
     relevant_sources.
 """
 function fit_object_psfs!{NumType <: Number}(
-    mp::ModelParams{NumType}, relevant_sources::Vector{Int}, blob::Blob)
+    mp::ModelParams{NumType}, target_sources::Vector{Int}, blob::Blob)
 
   # Initialize an optimizer
   initial_psf_params = PSF.initialize_psf_params(psf_K, for_test=false);
   psf_transform = PSF.get_psf_transform(initial_psf_params);
   psf_optimizer = PSF.PsfOptimizer(psf_transform, psf_K);
+
   @assert size(mp.patches, 2) == length(blob)
+
   for b in 1:length(blob)  # loop over images
     Logging.debug("Fitting PSFS for band $b")
     # Get a starting point in the middle of the image.
@@ -278,6 +280,10 @@ function fit_object_psfs!{NumType <: Number}(
     raw_central_psf = blob[b].raw_psf_comp(pixel_loc[1], pixel_loc[2])
     central_psf, central_psf_params =
       PSF.fit_raw_psf_for_celeste(raw_central_psf, psf_optimizer, initial_psf_params)
+
+    # Get all relevant sources *in this image*
+    relevant_sources = get_all_relevant_sources_in_image(mp, target_sources, b)
+
     for s in relevant_sources
       Logging.debug("Fitting PSF for b=$b, source=$s, objid=$(mp.objids[s])")
       patch = mp.patches[s, b]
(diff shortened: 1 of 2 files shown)


src/api.jl
```diff
@@ -134,12 +134,12 @@ function infer(ra_range::Tuple{Float64, Float64},
     # Get indicies of entries in the  RA/Dec range of interest.
     entry_in_range = entry->((ra_range[1] < entry.pos[1] < ra_range[2]) &&
```
rgiordan, line 135: Am I missing something, or is this still going to fail to find sources immediately outside the RA, DEC rectangle, but which influence the rectangle?  We talked about adding a border.  When will we do that?  Is the plan to do the run without it?

src/ModelInit.jl
```diff
@@ -278,6 +280,10 @@ function fit_object_psfs!{NumType <: Number}(
     raw_central_psf = blob[b].raw_psf_comp(pixel_loc[1], pixel_loc[2])
     central_psf, central_psf_params =
       PSF.fit_raw_psf_for_celeste(raw_central_psf, psf_optimizer, initial_psf_params)
+
+    # Get all relevant sources *in this image*
+    relevant_sources = get_all_relevant_sources_in_image(mp, target_sources, b)
```
rgiordan, line 285: This will still fail to consider as relevant sources that are outside the image but which influence tiles on the edge of the image.  The problem is ultimately in get_tiled_image_sources, which only considers a single tiled image.

The easiest fix is to ignore this, but if we're going to do that, we should put comments in the code and open an issue at least.  The next easiest fix would be to enforce not fitting objects near the edge of an image.  Doing it right means building tile_sources from all objects rather than just a single image.

rgiordan:
The rule I try to follow is that if you fix a bug (especially one that is
time-consuming to track down), you should write a test that will ensure
that the bug would be detected by a test were it to re-occur.

Here, the problem was that you asked to fit a PSF for an object not in the
image.  One way you can do that is to have a test that loads two images
with imperfect overlap.  Choose three sources, one that is in both, and two
that are in one but not the other.  Run the same code that you have in
api.jl (you may have to factor it out) with some special default values in
the initial PSF.  Verify that the one in both images has both its PSFs
updated, and that the other two have only their relevant PSFs updated.

There are other ways to do it, but that's what I'd do.

src/api.jl
kbarbary, line 135: I think we would want to add the border when we call  `query_overlapping_fieldids(ramin, ramax, decmin, decmax)` in `infer_nersc`. The `find(entry_in_range, catalog)` should only get sources in the target sky patch.

BTW, this code hasn't changed in this PR - this was just a variable rename for consistency.

rgiordan, line 135: What will happen if you choose an (RA, DEC) rectangle in
query_overlapping_fieldids that contains the edge of a frame?  This will
not work correctly, right?

I know this code didn't change with this PR, but unfortunately I didn't
look over the original code carefully, so I'm just pointing out problems
now as I find them.

kbarbary:
I am likely not going to have time to do this right this week.

src/api.jl
kbarbary, line 135: > What will happen if you choose an (RA, DEC) rectangle in query_overlapping_fieldids that contains the edge of a frame?

That frame will be included.

rgiordan:
That's understandable.  Could we just comment up the code pointing out this
problem (so it's easier to spot next time) and open an issue, then?  I
think we're not likely to actually use the results of our first run for
astronomy, so maybe ignoring this isn't a catastrophe.  It will also be
easy to identify problematic sources ex-post (they will lie near the edge
of some frame).

src/api.jl
rgiordan, line 135: Here's a picture.  Object 1 and object 2 both contribute flux to each
other's region of the sky, but neither shows up in the catalog of each
other's frames.  Consequently, they will not be listed in each other's
tile_sources.  Please correct me if I've misunderstood something.

rgiordan, line 135: Sorry, I mean that object 2 doesn't occur in frame 1's catalog, and so
won't be included in object 1's tile sources for tiles in frame 1.  (I
could have done the labels more clearly but I think you get the idea.)

rgiordan, line 135: Heh, last time's the charm.  Object two is not in frame one's catalog, and
so object one's tiles in frame two will be missing object two's
contributions.  I really should have labeled them differently.

kbarbary, line 135: If you sent an image, I don't think they work through github over email.

If it helps, note that all catalogs for all frames are combined together at the start of processing.

rgiordan, line 135: Here's a link to the image with the names fixed:

https://docs.google.com/drawings/d/1mb6vv77Yz4j3xlgzxlr_wBFuck2vhEH2QnEovLAizk0/edit?usp=sharing

My question is whether object 2 gets considered when processing object 1 in frame 1.

rgiordan, line 135: If you do concatenate all the catalogs at the beginning, then that would, I think, answer my objection.  Each object gets matched to tiles in a particular image using its catalog entry and the PSF from the center of the image.  I'd be reassured by a unit test demonstrating that this works as described, though.

Issue #177, opened 2016-03-23
AssertionError in Transform.jl (NaNs?)


jeff-regier:
@rgiordan This log is from a running with ---level=DEBUG but the particular source that caused the problem isn't mentioned. I see some NaN values in the gradient. 

This region of the sky may be at the intersection of 4 fields. (It processes 20 bands before the crash.)

I'll post more information shortly...I'm running a lot of these tests and I seem to have misplaced the commands that generated this issue.

```
got -1.1856831651767843 at [-0.0,-0.0,0.41053708344388096,0.6665306141015759,0.07507302531925193,-3.480180142973772,-0.5214352454880979,-0.9062361012694099,12.772000209753038,13.988704637899467,0.324299592
16452775,0.14386006177556185,0.06006008616367584,0.03390907615489898,0.14915619891513954,0.15357521618696623,0.07022668259806904,0.05244422774246335,-3.2303880357606527,-5.45112759694562,-6.074736657496397
,-6.201570803957329,0.1083329116385043,-3.129492895171235,-4.819055377073362,-3.7355576405005233,-2.566608268947726,-14.35332313696428,-4.592330763099313] after 22 function evaluations (22 Newton steps)

Processing band 1...
Processing band 2...
Processing band 3...
Processing band 4...
Processing band 5...
Processing band 6...
Processing band 7...
Processing band 8...
Processing band 9...
Processing band 10...
Processing band 11...
Processing band 12...
Processing band 13...
Processing band 14...
Processing band 15...
Processing band 16...
Processing band 17...
Processing band 18...
Processing band 19...
Processing band 20...
Done trimming.
f_evals: 1 value: -10.53748750728998
...
f_evals: 18 value: NaN
32x3 DataFrames.DataFrame
| Row | names     | val1       | grad1 |
|-----|-----------|------------|-------|
| 1   | "u_1"     | 200.163    | NaN   |
| 2   | "u_2"     | 38.1681    | NaN   |
| 3   | "e_dev"   | 0.985      | NaN   |
| 4   | "e_axis"  | 0.8        | NaN   |
| 5   | "e_angle" | 0.00895981 | NaN   |
| 6   | "e_scale" | 0.2        | NaN   |
| 7   | "r1_1"    | 3.09768    | NaN   |
| 8   | "r1_2"    | 2.16548    | NaN   |
342213256
| 24  | "c2_2_2"  | 0.0420028  | NaN   |
[...]

Issue #175, opened 2016-03-23
skip fields with score=0


jeff-regier:
@kbarbary Would take a look at this when you have chance? I'm running infer on tall and narrow slice of sky here, that intersects two different stripes. 

```
jregier@cori11~/Celeste/bin $  ./celeste.jl infer-nersc 350 350.1 0 10 . --logging=DEBUG
...
gunzip --stdout /global/projecta/projectdirs/sdss/data/sdss/dr12/boss/photo/redux/301/7814
/objcs/4/fpM-007814-g4-0084.fit.gz > celeste/fpm/7814-4/fpM-007814-g4-0084.fit
gunzip --stdout /global/projecta/projectdirs/sdss/data/sdss/dr12/boss/photo/redux/301/7814
/objcs/4/fpM-007814-r4-0084.fit.gz > celeste/fpm/7814-4/fpM-007814-r4-0084.fit
gunzip --stdout /global/projecta/projectdirs/sdss/data/sdss/dr12/boss/photo/redux/301/7814
/objcs/4/fpM-007814-i4-0084.fit.gz > celeste/fpm/7814-4/fpM-007814-i4-0084.fit
gunzip --stdout /global/projecta/projectdirs/sdss/data/sdss/dr12/boss/photo/redux/301/7814
/objcs/4/fpM-007814-z4-0084.fit.gz > celeste/fpm/7814-4/fpM-007814-z4-0084.fit
ERROR: LoadError: TypeError: read_photoobj: in typeassert, expected FITSIO.TableHDU, got F
ITSIO.ImageHDU
 in read_photoobj at /global/homes/j/jregier/.julia/v0.4/Celeste/src/SDSSIO.jl:334
 in read_photoobj at /global/homes/j/jregier/.julia/v0.4/Celeste/src/SDSSIO.jl:331
 in read_photoobj_primary at /global/homes/j/jregier/.julia/v0.4/Celeste/src/api.jl:70
 in infer at /global/homes/j/jregier/.julia/v0.4/Celeste/src/api.jl:142
 in infer_nersc at /global/homes/j/jregier/.julia/v0.4/Celeste/src/api.jl:413
 in main at /global/u2/j/jregier/Celeste/bin/celeste.jl:37
 in include at ./boot.jl:261
 in include_from_node1 at ./loading.jl:320
 in process_options at ./client.jl:280
 in _start at ./client.jl:378
[...]

Issue #169, opened 2016-03-20
access to undefined reference


jeff-regier:
```
jregier@cori10~/Celeste/bin $ ./celeste.jl infer-nersc 164.3 164.35 39.1 39.15 .
...
20-Mar 15:56:44:INFO:root:field (3900,6,268): 808 entries
20-Mar 15:56:44:INFO:root:field (3900,6,269): 805 entries
20-Mar 15:56:44:INFO:root:field (3900,6,268): 725 primary entries
20-Mar 15:56:44:INFO:root:field (3900,6,269): 703 primary entries
20-Mar 15:56:45:INFO:root:1428 primary sources
20-Mar 15:56:45:INFO:root:1023 primary sources after MIN_FLUX cut
20-Mar 15:56:45:INFO:root:processing 24 sources in RA/Dec range
20-Mar 15:56:45:INFO:root:reading field (3900,6,268)
20-Mar 15:56:49:INFO:root:reading field (3900,6,269)
20-Mar 15:56:51:INFO:root:initializing celeste without PSF fit
Getting VP from sources.

20-Mar 15:56:59:INFO:root:fitting PSF for all relevant sources
20-Mar 15:57:17:INFO:root:processing source 275: objid= 1237662226207998451
Processing band 1...
Processing band 2...
Processing band 3...
Processing band 4...
Processing band 5...
Done trimming.
ERROR: LoadError: UndefRefError: access to undefined reference
 in elbo_likelihood! at /global/homes/j/jregier/.julia/v0.4/Celeste/src/ElboDeriv.jl:968
 in elbo_likelihood! at /global/homes/j/jregier/.julia/v0.4/Celeste/src/ElboDeriv.jl:1003
 in elbo_likelihood at /global/homes/j/jregier/.julia/v0.4/Celeste/src/ElboDeriv.jl:1046
 in anonymous at /global/homes/j/jregier/.julia/v0.4/Celeste/src/OptimizeElbo.jl:213
 in f_objective at /global/homes/j/jregier/.julia/v0.4/Celeste/src/OptimizeElbo.jl:107
 in f_value_grad at /global/homes/j/jregier/.julia/v0.4/Celeste/src/OptimizeElbo.jl:124
 in f_grad at /global/homes/j/jregier/.julia/v0.4/Celeste/src/OptimizeElbo.jl:148
 in f_grad! at /global/homes/j/jregier/.julia/v0.4/Celeste/src/OptimizeElbo.jl:152
[...]

PR #179, opened 2016-03-24
Use Dtree for distributed parallelism

kpamnany:
This will run as before on a single node and additionally output timing. Multi-node support is in progress.

Done:
- Using Dtree to distribute work.
- Add load imbalance measurement.
- Added timing for various Celeste parts.

TODO:
- Tune work item ra/dec based on load imbalance, other criteria?
- Use a thread to drive Dtree in parent nodes.

### DIFF
--- a/bin/celeste.jl
+++ b/bin/celeste.jl
@@ -1,7 +1,13 @@
 #!/usr/bin/env julia
 
 using DocOpt
+using Dtree
 import Celeste
+
+const cpu_hz = 2.3e9
+
+const wira = 0.015
+const widec = 0.015
 
 const DOC =
 """Run Celeste.
@@ -19,7 +25,7 @@ Usage:
 Options:
   -h, --help         Show this screen.
   --version          Show the version.
-  --logging=<LEVEL>  Level for the Logging package (OFF, DEBUG, INFO, WARNING, ERROR, or CRITICAL). [default: INFO]
+  --logging=<LEVEL>  Level for the Logging package (OFF, DEBUG, INFO, WARNING, ERROR, or CRITICAL). [default: WARNING]
   --stage            In `infer-box`, automatically stage files. Default: false.
 
 The `stage-box` subcommand copies and/or uncompresses all files covering the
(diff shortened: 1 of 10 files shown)


jeff-regier:
Bugs #168 and #169 are resolved. Ryan wanted a unit test for them, but that's not a priority imo given the timeline we're on, and because I've tested the new code pretty thoroughly. Bugs #175 and #177 persist, but #175 doesn't look like a bad one at least. I'm trying to further pin down #177.

kpamnany:
Stuck on a Julia multithreading bug (https://github.com/JuliaLang/julia/issues/15620). This won't interfere with multi-node scaling, on which I am continuing work.

kpamnany:
This isn't completely ready yet. I still have to add some instrumentation to report load imbalance and clean up the timing output so it isn't so messy. Hopefully by EOD, then the output can be verified for correctness as follows.

Grab a Cori node with `salloc -N 1 -p debug -t 00:30:00`. Make sure to `export JULIA_NUM_THREADS=1`. Then you can run multi-process with `mpiexec.hydra -np X bin/celeste.jl infer-nersc 200.0 200.05 38.1 38.15 $SCRATCH` where `X` is the number of processes you want (up to 32 on a single node).

True multi-node runs should also work, but I haven't tested those yet. Right now, for the 27 sources in this patch of sky, it looks like 659 seconds with 1 process, 309 seconds with 2 processes (super-linear? how?), and 201 seconds with 4 processes, all with static scheduling.

kpamnany:
Okay. I think this is ready. I'll work on multi-threading and other improvements on another branch.

kpamnany:
Note: there are a number of ugly hacks here. There's probably a better place for `nputs`. Passing `nid` and `nnodes` around is clunky. As @rgiordan has pointed out, the `Threaded` global in `ElboDeriv.jl` should ideally not be a global there. I'm not sure how best to resolve some of these, so feel free to fix/suggest fixes.

bin/celeste.jl
```diff
@@ -1,8 +1,14 @@
 #!/usr/bin/env julia
 
 using DocOpt
+using Dtree
 import Celeste
 
+const cpu_hz = 2.3e9
+
+const wira = 0.015
```
rgiordan, line 10: These are the width of ra / dec rectangles.  Could you please add a comment to this effect?

src/ModelInit.jl
```diff
@@ -671,7 +671,7 @@ function trim_source_tiles(
       has_source = s in tile_sources
       bright_pixels = Bool[];
       if has_source
-        #println("Tiles $h $w has source $s")
+        info("Tiles $h $w has source $s")
```
rgiordan: I don't think you want to display this -- it was just for debugging.  My bad for commenting it out and not just deleting it.

src/Transform.jl
```diff
@@ -890,7 +890,7 @@ function enforce_bounds!{NumType <: Number}(
         if !(constraint.lower_bound <=
              mp.vp[s][ids.(param)[ind]] <=
              constraint.upper_bound)
-          println("Warning: $param[$s][$ind] was out of bounds.")
+          info("Warning: $param[$s][$ind] was out of bounds.")
```
rgiordan, line 893: info("Warning:...") seems like a strange thing to say.

src/api.jl
```diff
@@ -161,13 +182,14 @@ function infer(fieldids::Vector{Tuple{Int, Int, Int}},
 
     # If there are no objects of interest, return early.
     if length(target_sources) == 0
-        return Dict{Int, Dict}()
+        return Dict{Int, Dict}(), (0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0)
```
rgiordan, line 195: It would be nice to put all these times in a data structure.

src/api.jl
```diff
@@ -15,6 +30,7 @@ const TILE_WIDTH = 20
 const DEFAULT_MAX_ITERS = 50
 const MIN_FLUX = 2.0
 
+@inline nputs(nid, s) = ccall(:puts, Cint, (Ptr{Int8},), string("[$nid] ", s))
```
rgiordan, line 36: Just curious, why use this and not Logging?

src/ElboDeriv.jl
```diff
@@ -11,8 +11,8 @@ import ..KL
 import ForwardDiff
 import ..WCSUtils
 
-Threaded = true
-if VERSION > v"0.5.0-dev"
+Threaded = false
```
rgiordan, line 17: Is the threaded stuff in ElboDeriv.jl used at all anymore?  If not, are you just leaving it in in case we change our mind in the future?  If so, could you maybe mention that in a comment?

bin/celeste.jl
```diff
@@ -27,7 +33,175 @@ field, using only that field.
...
+function celeste_skyarea(ramin, ramax, decmin, decmax, outdir)
+    # ---
+    if dt_nodeid == 1
+        println("celeste -- running on $dt_nnodes nodes (system clock speed is $cpu_hz Hz)")
+	println("  sky area: $(ramax - ramin) X $(decmax - decmin)")
+    end
+
+    n_infers = 0
+    n_fits = 0
+    tm_infer = tm_fldids = tm_rdprim = tm_filcat = tm_rdimg = tm_imp = tm_psf = tm_fit = tm_out = tm_done = 0.0
```
rgiordan, line 91: There are a lot of line overruns and tab misalignments -- I won't comment each one.

bin/celeste.jl
```diff
@@ -27,7 +33,175 @@ field, using only that field.
...
+end
+
+"""
+Given a SkyArea that is to be divided into `skya.nra` x `skya.ndec` patches,
+return the `i`th patch. `i` is a linear index between 1 and
+`skya.nra * skya.ndec`.
+
+This function assumes a cartesian (rather than spherical) coordinate system!
+"""
+function sky_subarea(skya::SkyArea, i)
```
rgiordan, line 64: Just my opinion, but I think it's good style to give functions verb names and variables noun names.

bin/celeste.jl
```diff
@@ -27,7 +33,175 @@ field, using only that field.
...
+
+"""
+Divide the area of sky subtended by `ramin`, `ramax`, `decmin`, and `decmax`
+into patches of `wira`X`widec`. Each such patch is a work item. Use Dtree to
+distribute these to nodes. Within the node, run inference on each patch
+received from Dtree.
+"""
+function celeste_skyarea(ramin, ramax, decmin, decmax, outdir)
+    # ---
+    if dt_nodeid == 1
```
rgiordan, line 84: Where are `dt_nodeid` and `dt_nnodes` set?

rgiordan:
This looks good to me substantively (without knowing the workings of Dtree).  I just had some superficial comments and questions -- we can always tidy up after the deadline.  :)

bin/celeste.jl
kpamnany, line 84: They come in from Dtree.

src/api.jl
kpamnany, line 36: This runs "atomically", so the text from multiple nodes isn't interleaved. Also, Logging is used extensively all over Celeste, and I need to turn all that off otherwise there is too much screen spam.

src/Transform.jl
kpamnany, line 893: I was just removing all the println()s. I'll let you guys decide if these info/warning/print messages are required.

bin/celeste.jl
kpamnany, line 91: I think I fixed all of them.

src/api.jl
kpamnany, line 195: Where should I define this type (and other stuff), so that it is visible in `bin/celeste.jl` and in `src/api.jl`?

rgiordan, line 195: I don't think it needs to be visible in `bin/celeste.jl`, does it?  In either case, simply defining it in `src/api.jl` should do since (for now at least) it's simply included in the Celeste module.

kpamnany, line 195: Some additional time measurements are made in `bin/celeste.jl` and they are all output there. Also, I want to put `nputs` and other such stuff that is used in both places somewhere... putting stuff in `src/api.jl` makes it necessary to add the `Celeste.` prefix, or to explicitly export things. :(

rgiordan, line 195: Well, IMO you're putting things in `bin/celeste.jl` that should be in `api.jl` (and tested), and there are things in `api.jl` that should be in `SDSSIO.jl` (or something like it).  Again, IMO, timing measures are attributes of an API and should be defined with the API in its own module, not in a binary script.

After the paper I think we should organize the whole thing better.  In the meantime, just go ahead and do what you think is best that gets you up and running the fastest that doesn't interfere with the rest of us by moving stuff around too much, and we'll be fine.  :)

jeff-regier, line 195: It's fine to hold off on writing unit tests, but I do think it's going to be confusing if `celeste.jl` contains code that isn't just about parsing arguments and calling the right function in `api.jl`.

kpamnany:
Closing this; will open a new PR with a refactored implementation.

PR #181, opened 2016-03-28
handle missing table in SDSSIO.read_photoobj

kbarbary:
Addresses #175.


jeff-regier:
It may not suffice to just skip the photoobj files. For fields that have timed out, I think some of the other files are missing too. We probably want to use window_flist file to decide what fields to skip.

kbarbary:
Do you have some specific reason to believe that such fields would have other missing files, or is this just being cautious?

jeff-regier:
I tried just skipping the bad photoobj files earlier, but then problems
with other files for those fields appeared. I think we just want to skip
the fields with `score=0` in the window_flist file.

kbarbary:
Fair enough. Instead of reading score from the window_flist file, I added the requirement score!=0 to the original query used to get all the field info (in the `/project/projectdirs/dasrepo/celeste-sc16/field_extents.fits` file). This eliminated a few percent of the fields. I'm not completely positive that the SkyServer table has the same info as `window_flist` in this respect, but it seems to. The query used to get the field extents file is

``` sql
select run, camcol, field, ramin, ramax, decmin, decmax
into mydb.field_extents
from dr12.field
where rerun=301 and score != 0.0
```

which is documented in `/project/projectdirs/dasrepo/celeste-sc16/README`.

This PR shouldn't strictly be necessary but should probably be merged. The possibility of a missing table seems to be part of the SDSS data model (at least in practice), and this PR allows `read_photoobj()` to handle that case gracefully.

jeff-regier:
Sounds good, I'll try this out on regions of the sky that have crashed Celeste before.

kbarbary:
If there are still problems, I'll investigate whether the window_flist file contains score=0 entries that are score!=0 in the database.

PR #182, opened 2016-03-29
infer one field & catch exceptions

jeff-regier:
@kbarbary Would you look over these changes, primarily to `api.jl`? They let me run Celeste on one field at a time, without using the "best" detection from any field as initialization.

I'm also catching any exceptions inside the source processing loop now. That way @kpamnany can run on any region of the sky right away, and we can get better statistics about what kinds of errors Celeste is throwing. (Unfortunately #168 seems to be back.)


kbarbary:
I think this makes sense. One comment: `ignore_primary_mask` is a strange keyword to pass to a function called `read_photoobj_primary`.  It might be a little clearer to instead have `infer()` take an `initialize` key that could be either "primary" or "first". Then do something like

```
if initialize == "primary":
    catalog = read_photoobj_primary(fieldids, photoobj_dirs)
else
    catalog = read_photoobj_single(fieldids[1], photoobj_dirs[1])
end
```

(`read_photoobj_single` would be a new short function that doesn't do any primary checking). The benefit is that options don't have to be passed so deep into `read_photoobj_primary`.

Up to you though.

jeff-regier:
Makes sense, I'll get rid of `ignore_primary_mask`. But I'm wary of duplicating code, creating a `read_photoobj_single` whose functionality is a subset of `read_photoobj_primary`. I may have a solution.

Issue #186, opened 2016-03-30
star position is off


jeff-regier:
Many objects' positions seem to be off. For the object with objid `1237663784734491542`, for example, Celeste reports its ra/dec as 0.556922/0.624469, but primary finds it at 0.557017/0.624494 . (Why does it move away from the initialization? )

The [Sky Server Navigation Tool](http://skyserver.sdss.org/dr12/en/tools/chart/navi.aspx) suggests Primary (and Photo on the coadd catalog) are right, and that Celeste is pretty far off. (But I don't know exactly how that web tool accounts for the PSF.) @rgiordan  Do you have time to look into this? There are other problems with the scores too, and I could use help getting to the bottom of this one. Is it strange the optimizer converges here in just 2 steps?

With PR #185, these results can be recreated with two commands:

```
./celeste.jl --logging=DEBUG infer-object 4263 5 119 1237663784734491542 .

./celeste.jl --logging=DEBUG score-object 4263 5 119 1237663784734491542 .  /global/homes/j/jregier/coadd_catalog_from_casjobs.fit 
```

rgiordan:
Yes, I can try to take a look at this.  I have some 205B to do but I'll
prioritize this when it's done.

rgiordan:
This is a very dim object and I strongly suspect that the trim_source_tiles
function is being too conservative when it eliminates pixels.

I can think of two solutions, and we may want a combination.  One is to
make sure that at least a certain radius of pixels around the original
object are selected.  Another is to set the noise_fraction to be a function
of the brightness of the object.  The former is easier and probably better
-- let me put together a PR.

Ryan

rgiordan:
Fixed in https://github.com/jeff-regier/Celeste.jl/pull/187

PR #187, opened 2016-03-30
Minimum source radius

rgiordan:
This seems to do the trick -- the problem source in https://github.com/jeff-regier/Celeste.jl/issues/186 is fit to nearly the catalog value with a minimum radius of six pixels.

Tests pending.

### DIFF
--- a/dev/celeste_interactive.jl
+++ b/dev/celeste_interactive.jl
@@ -7,9 +7,9 @@ using PyPlot
 
 import FITSIO
 import JLD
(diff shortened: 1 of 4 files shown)


rgiordan:
@jeff-regier , tests passing, I'll let you merge at your convenience.

jeff-regier:
Great! I'll start rerunning inference on a field tonight, and check whether the scores for the whole field improve tomorrow morning.

Issue #191, opened 2016-03-31
inaccurate galaxy angles


jeff-regier:
Celeste appears to not be finding galaxy's angles very accurately. (Or else something is wrong with the score function.) Celeste is currently off by 45 degrees on average, while primary (and Celeste in the ICML paper) is off by about 20 degrees. (Random guessing is 90 degrees off.)

Consider objid = 1237663784734490990. It's a bright object. It's axis ratio is low enough that angle means something. (With an axis ratio of 1, the galaxy appears as a circle, and angle doesn't have a meaning.) In the sky navigator tool (coordinates: 0.537991, 0.553257), it's obvious which direction it lies. Both Primary and Photo determine that this galaxy is about 175 degrees in angle. Celeste returns 85 degrees.

rgiordan:
I'll bet we're 90 degress off of the catalog's notion of "angle" in Celeste.  Check out these images from that 1237663784734490990 as Celeste sees them.  I will look into this.

![screenshot from 2016-03-31 15-05-52](https://cloud.githubusercontent.com/assets/4494910/14192393/495c9ce6-f752-11e5-94ed-40ee7f96f961.png)

Wouldn't random guessing be forty-five degrees off?

jeff-regier:
You're right, 45 degrees off is random. 90 is the most possible.

And yeah I'm pretty sure it's 90 off, with perhaps one computing angles
clockwise and the other counterclockwise.

The problem may affect initialization too, not just scoring... I notice
some really big Newton steps away from the initial value, suggesting the
initialization could be off

rgiordan:
Yes, looking at this catalog image, it's clear that the initialization is off by 90 degrees.  Fixing this could result in a modest speedup, too.

rgiordan:
Here's the logic that has apparently been in Celeste for a while:

```
        # use tractor convention of defining phi as -1 * the phi catalog
        fits_phi *= -1.0

        phi90 = 90 - fits_phi
        phi90 -= floor(phi90 / 180) * 180
        phi90 *= (pi / 180)
```

@jeff-regier , do you remember why you did it this way?  The [photoObj documentation](http://data.sdss3.org/datamodel/files/BOSS_PHOTOOBJ/RERUN/RUN/CAMCOL/photoObj.html) is not very helpful.

jeff-regier:
That's the way the worked for the stamps. You can tell when it works by
examining individual galaxies, but it was not easy. It seems like the
[...]

PR #195, opened 2016-03-31
fixes the scoring part of the angle problem

jeff-regier:
Hey @rgiordan -- This fixes half of #191 -- the scoring part of the problem. I did some weird things with angles a year ago when I was scoring stamps, but apparently that logic isn't needed for the current setup.

The initialization is probably still off by 90 degrees so I'm glad you're working on that.


jeff-regier:
Average angle error is 20.7 degrees for Celeste now---much better than random guessing

PR #202, opened 2016-04-01
Dont use mp_array

rgiordan:


rgiordan:
NB: I sanity checked it on Cori.

PR #206, opened 2016-04-02
Parallelize psf

rgiordan:
@kpamnany , PTAL.  I've run it on NERSC but we're not currently set up to unit test this.

### DIFF
--- a/src/api.jl
+++ b/src/api.jl
@@ -380,7 +380,7 @@ function fit_object_psfs_threaded!{NumType <: Number}(
       PSF.fit_raw_psf_for_celeste(raw_central_psf, psf_optimizer_vec[1], initial_psf_params)
 
     # Get all relevant sources *in this image*
-    relevant_sources = get_all_relevant_sources_in_image(mp, target_sources, b)
+    relevant_sources = ModelInit.get_all_relevant_sources_in_image(mp, target_sources, b)
 
     mp_lock = SpinLock()
     @threads for s in relevant_sources
--- a/src/api.jl
+++ b/src/api.jl
@@ -370,7 +370,7 @@ function fit_object_psfs_threaded!{NumType <: Number}(
     pixel_loc = Float64[ blob[b].H / 2.0, blob[b].W / 2.0 ]
     raw_central_psf = blob[b].raw_psf_comp(pixel_loc[1], pixel_loc[2])
     central_psf, central_psf_params =
-      PSF.fit_raw_psf_for_celeste(raw_central_psf, psf_optimizer, initial_psf_params)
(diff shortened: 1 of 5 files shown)


src/api.jl
```diff
@@ -429,7 +492,11 @@ function infer(fieldids::Vector{Tuple{Int, Int, Int}},
     # fitting the PSF for the whole catalog above)
     info("fitting PSF for all relevant sources")
     tic()
-    ModelInit.fit_object_psfs!(mp, target_sources, images)
+    if Threaded
+      fit_object_psfs_threaded!(mp, target_sources, images)
+    else
+      ModelInit.fit_object_psfs!(mp, target_sources, images)
+    end
     times.fit_psf = toq()
```
kpamnany, line 518: It should be possible to have a single function `fit_object_psfs!()` rather than separate ones for threads and serial -- see `infer()`. There should be no need to test `Threaded` because if it is disabled, things like `@threads` do nothing, and `threadid()` and `nthreads()` always return 1.

kpamnany:
I'll be raising a PR shortly that adds code to `infer()` to drive Dtree with one of the threads. Dtree determines a few nodes (roughly one per thousand) to be parent nodes; these must reserve a thread to run work distribution. The way this is done is pretty clunky, but there's no alternative right now. I will need to add the same sort of patch to this code as well, after you merge it.

src/api.jl
rgiordan, line 518: This is only true for Julia 0.5, right?  To be compatible with 0.4 we'd
need to make your threading definitions from api.jl global.  That would be
OK with me if you feel strongly.  IMO, until we test with multiple threads,
there's not much point.

kpamnany, line 518: Right, and right. We can wait on this until Julia threads are more stable, which will be a good while yet.

kpamnany:
If you're curious, see https://github.com/jeff-regier/Celeste.jl/pull/207 for what needs to be done with this parallel loop as well. I'll add that functionality after this is merged.

rgiordan:
@kpamnany , I'm not sure exactly what you're referring to in #207 that needs to be done here.

jeff-regier:
Can we merge this? I'd like to test the frozen code once on stripe 82 to before we start the big runs, to be sure the scores remain good.

rgiordan:
I think Kiran needs to help me integrate it with #207, which I don't
understand.

rgiordan:
Ping @kpamany

kpamnany:
Just merge this please; I'll make the necessary changes in a later PR.

rgiordan:
My rebasing attempts have gotten out of hand.  I made a different PR that's cleaner.

PR #207, opened 2016-04-02
{WIP} Drive Dtree with a thread

kpamnany:
Adds calls to drive Dtree work distribution by parent nodes in a number of places in `divide_and_infer()`, and with a reserved thread in the `infer()` parallel loop.

Still seeing signs of https://github.com/JuliaLang/julia/issues/15620.

### DIFF
--- a/src/api.jl
+++ b/src/api.jl
@@ -47,8 +47,8 @@ else
 end
 
 # A workitem is of this ra / dec size
-const wira = 0.05
-const widec = 0.05
+const wira = 0.04
+const widec = 0.04
 
 """
 Timing information.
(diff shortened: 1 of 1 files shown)


src/api.jl
```diff
@@ -47,8 +47,8 @@ else
 end
 
 # A workitem is of this ra / dec size
-const wira = 0.05
-const widec = 0.05
+const wira = 0.04
```
rgiordan, line 50: How about making these command line arguments with defaults?

rgiordan:
I don't think I know Dtree well enough to understand this PR, but I'll have some faith and merge to move things forwards.

PR #208, opened 2016-04-02
Fix nm bug

rgiordan:
As I was looking into improving the quality of the PSF fit, I found a bug in my Newton's Method implementation.  When the Hessian is not PD, this bug can cause the algorithm to terminate early in certain extreme cases even when not at a local minimum.  I suspect it was not affecting the astronomy results, but it was affecting the PSF fit.

The substantive change is the line `elseif m > 0`.

Since the manifestation was an early algorithm termination, I hope that this fix should only make things better.

### DIFF
--- a/dev/fit_psf.jl
+++ b/dev/fit_psf.jl
@@ -2,6 +2,7 @@ using Celeste
 using Celeste.Types
 using Celeste.SkyImages
 using Celeste.PSF
+using DataFrames
 
 import Celeste.SDSSIO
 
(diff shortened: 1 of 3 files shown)


rgiordan:
NB: Cori results are not dramatically different than before.

```
bin/celeste.jl infer-box 200.0 200.05 38.1 38.15 $SCRATCH &> $SCRATCH/logfile
<snip>
[1] timing: read_photoobj=1.16107849
[1] timing: read_img=7.973080223
[1] timing: init_mp=4.61964249
[1] timing: fit_psf=25.918580121
[1] timing: opt_srcs=22.041329892
[1] timing: num_srcs=27
[1] timing: average opt_srcs=0.8163455515555555
[1] timing: write_results=3.840896484
[1] timing: wait_done=0.0
```

PR #213, opened 2016-04-05
Enable running with Dtree even single-threaded

kpamnany:
Tested with multiple MPI processes on a single node. Almost no change here, but a number in Dtree that I've just merged into master. So, until this PR is merged, tests on Travis will break.

After you `salloc` a node, `export I_MPI_PIN_DOMAIN=auto`, and then run with `mpiexec.hydra -np X bin/celeste.jl ...` where X is the number of processes you want -- up to 32 on a single node. Please check that the results are okay. I'll do more scaling tests and also test multi-node tomorrow.

### DIFF
--- a/src/api.jl
+++ b/src/api.jl
@@ -16,7 +16,7 @@ const MIN_FLUX = 2.0
 
 # Use distributed parallelism (with Dtree)
 const Distributed = true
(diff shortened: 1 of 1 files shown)


jeff-regier:
Looks good to me. I'll try it out tomorrow on Cori

PR #214, opened 2016-04-05
Minor cleanup

kpamnany:
Remove GC safepoint call (not in Julia 0.4). Fix for timing output. Other minor cleanups.

### DIFF
--- a/src/ModelInit.jl
+++ b/src/ModelInit.jl
@@ -894,7 +894,6 @@ function initialize_model_params(
                                                      patch_ctrs_pix(patches),
                                                      patch_radii_pix(patches))
     end
-    print("\n")
 
     # improved initializations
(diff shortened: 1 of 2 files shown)


jeff-regier:
@kpamnany  Would you register Dtree, so we can add it to the REQUIRE file? After the recent PRs Celeste no longer works because it needs Dtree.

kpamnany:
I'll register it, but was hoping to put it off till post-deadline. You just need to pull the latest version of Dtree and I think everything works. Travis is passing...

jeff-regier:
OK, that's fine. Unfortunately the libraries Dtree (MPI?) needs aren't
installed correctly on my development workstation. In the meantime, I'm
going change Celeste so that it only uses Dtree if an environment named
"USE_DTREE" is set.

PR #224, opened 2016-04-26
moves initialization and psf fitting inside the threaded loop

jeff-regier:
This PR moves 1) model initialization, and 2) psf fitting instead the main threaded loop over sources.

It adds a new routine outside the threaded loop, that finds each sources' neighboring sources, ahead of time. This serial routine has quadratic runtime, but in practice in finishes in about 1 second for thousands of sources.

Scores on stripe 82 continue to look good, though the `relevant_sources` aren't always the same.

@kpamnany  Will you try this out with threading enabled? There may be less interaction between threads now.

### DIFF
--- a/src/ModelInit.jl
+++ b/src/ModelInit.jl
@@ -238,9 +238,9 @@ function trim_source_tiles(s::Int,
                     end
                 end
             else
-                # This tile does not contain the source.    Replace the tile with a
+                # This tile does not contain the source. Replace the tile with a
                 # pseudo-tile that does not have any data in it.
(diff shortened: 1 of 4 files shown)


kpamnany:
It segfaults, apparently in `initialize_model_params()`. Something is not thread-safe? Any ideas @rgiordan?

jeff-regier:
I've refactored away `initialize_model_params()` now---it's broken into a few different functions. Could you try it again, and see if we can get any more information about where the seg fault happens? I'm wondering whether code interacting with WCSLIB (a C library), in the function `get_tile_source_map()` may be the source of the problem.

PR #234, opened 2016-06-23
Fix solve tr subproblem

jeff-regier:


test/test_newton_trust_region.jl
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env julia
```
rgiordan, line 1: Thanks for doing all this!

There's no reason you would know, but I wrote a unit test that's in the trust_region branch of my fork of Optim.jl.  You may want to give it a try, or LMK if you would like me to do it.  I can get to it after ICML.

https://github.com/rgiordan/Optim.jl/blob/trust_region/test/newton_trust_region.jl

jeff-regier, line 1: Hey Ryan, That'd be great if you'd add those tests whenever you have the
chance.

What's left to do to merge it with Optim.jl?

PR #265, opened 2016-08-11
ForwardDiff in tests only

rgiordan:
This moves all the ForwardDiff tests into a single file, adds a command line flag to runtests, and uses analytic gradients and Hessians for the KL divergence, which removes ForwardDiff from the actual inference.

Many of the tests currently sequestered to test_derivatives.jl are important.  I'd like to merge this now before finding a more permanent solution to avoid more merges as everyone else continues to make changes.


REQUIRE
```diff
@@ -4,7 +4,7 @@ DataFrames
 Distributions
 DocOpt
 Dtree
-ForwardDiff 0.1.8 0.2
+ForwardDiff 0.2
 JLD
 Optim 0.5 0.5.1
```
jeff-regier, line 9: Can we remove the upper bound from Optim now that ForwardDiff is free?

doc/celeste_equations.lyx
```diff
@@ -1630,6 +1630,170 @@ Also,
```
jeff-regier, line 1630: Why are the equations changing? (Is it somehow related to ForwardDiff?)

src/Model.jl
```diff
@@ -30,7 +30,6 @@ import Base.+
 import Distributions
 import FITSIO
 import WCS.WCSTransform
-import ForwardDiff
```
jeff-regier, line 33: Nice! Glad to have this out of here.

jeff-regier:
Looks good Ryan. 

I think it'd be good practice to start squashing commits into one, or into logical groups, before committing to master, i.e. `git rebase HEAD~13` in this case. There's a post about it here:
https://programmers.stackexchange.com/questions/263164/why-squash-git-commits-for-pull-requests

Main idea is to keep the history readable.

doc/celeste_equations.lyx
jeff-regier, line 1630: I think with `git rebase`, you can reorder commits as well as squash related commits into one. If these doc changes are different from the ForwardDiff change, you might squash the ForwardDiff changes together, but leave this one as its own.

REQUIRE
rgiordan, line 9: I think so, but I'm just going to fix the Optim stuff in a separate CL.

doc/celeste_equations.lyx
rgiordan, line 1630: I added the formulas for the KL gradients and Hessians.

rgiordan:
I'll close this and re-open a PR that's been rebased.

jeff-regier:
Actually I've found that of just rebasing the branch automatically updates
the pr. GitHub is pretty slick...I figured that changing the history would
throw it off

PR #267, opened 2016-08-11
Fd test rebase

rgiordan:
ForwardDiff's compile time has become really slow, but the next version of Optim depends on upgrading ForwardDiff.  This PR removes ForwardDiff from the main inference by using analytic derivatives for the KL divergences.  It also factors out the ForwardDiff-based tests into a separate file which is only run optionally.


jeff-regier:
Looks like a merge gone bad...some << characters in the commit perhaps?

rgiordan:
Don't merge this yet.  The rebase re-added some tests that I had removed.  I think an interactive rebase after a merge (of your recently committed PR) may not actually be so slick.

jeff-regier:
Rebase once more? Or we can just start doing this in the future if you prefer, and just want to be done...

rgiordan:
I rebased and tried to push, but it said

```
error: failed to push some refs to 'https://github.com/jeff-regier/Celeste.jl'
hint: Updates were rejected because the tip of your current branch is behind
hint: its remote counterpart. Integrate the remote changes (e.g.
hint: 'git pull ...') before pushing again.
```

This makes sense, but I thought based on your previous comment that I could rebase and push like this.  Am I doing something wrong?

jeff-regier:
I'm not sure.... probably something about doing rebase to apply changes to
master to your branch. We can certainly wait until next time to worry about
keeping the history tidy. We might want to have a policy of only making
commits to master with passing unit tests. But please just merge this pr
whenever you're happy with it

PR #269, opened 2016-08-12
Use Optim version of trust region

rgiordan:
In addition to removing the Celeste trust region code, I move most of the derivative tests back into the main test suite, leaving out only three very slow tests to be run with a custom flag.


rgiordan:
This is failing because Optim.jl hasn't tagged a version with the trust region code yet.

@jeff-regier , do you have a preferred method to get a particular commit for Travis?  I'm going to try checking out master on Optim in travis, but I'm open to other ways of doing it.

jeff-regier:
Hey Ryan, I think the best way forward is to ask them to tag it. Any other
solution I can think of is really a hack, with drawbacks. You contributed
some good code to their package, so they should tag a release with it for
you.

rgiordan:
How about that, the problem literally solved itself overnight.  :)  Thought it seems it hasn't been pushed to METADATA yet.

rgiordan:
Waiting on https://github.com/JuliaLang/METADATA.jl/pull/5947.

pkofod:
> Hey Ryan, I think the best way forward is to ask them to tag it. Any other solution I can think of is really a hack, with drawbacks. You contributed some good code to their package, so they should tag a release with it for you.

Don't worry, I'm doing what I can to get it live ;) NelderMead deprecation handling is holding the merge back, but that should be fixed.

rgiordan:
@pkofod no problem!

pkofod:
By the way, this is the new PR https://github.com/JuliaLang/METADATA.jl/pull/5957 

I'm very happy to see you using `OptimizationOptions` by the way. I think keywords in `optimize` will go at some point, so :)

PR #274, opened 2016-08-15
source-division parallelism

jeff-regier:


jeff-regier:
@kpamnany -- This code implements the new method of parallelism, but wo the parallelism. It's all in the file "src/source_division_inference.jl", except for one boolean flag in `api.jl`. Lines where you'd insert calls to GlobalArrays.jl, or Dtree.jl, or `Base.threads` are marked with `TODO`.

This new type of parallelism isn't actually all that simple. Lmk if you want to chat about the code. I was able to avoid using any multi-dimensional array. I also avoided using arrays with variable-size entries, by recording offsets into array per field. But I had to introduce this data structure called `tasks` in addition to `sources`, because some sources are needed for rendering the background, but they shouldn't be optimized. 

I suspect we're going to encounter unexpected problems while to get this new scheme working at scale, so please start attempting large runs early.

PR #291, opened 2016-09-14
Jr/mcmc

andymiller:
This PR adds a generic log joint probability function to the Model module, in files `log_prob.jl` and `log_prob_utils.jl`, which is suitable for MCMC and perhaps stochastic gradient VI. 

The main functionality comes from `Model.make_star_logpdf` and `Model.make_galaxy_logpdf`, which return function handles for the log joint probability of a vector of TiledImages given a flat parameter `Vector{Float64}` object.  For stars and gals, the `Vector{Float64}` object represent the following parameters

```
  star_params = [log(r), log(u/g), log(g/r), log(r/i), log(i/z), ra, dec]
  gal_params  = [log(r), log(u/g), log(g/r), log(r/i), log(i/z), ra, dec,
                           logit(fracdev), logit(ab), angle, log(scale)]
```

I also added 'correctness' (more like not completely wrong) tests for the star logpdf, as well as a handful of unit test for parameter transformations.


jeff-regier:
Looks good, thanks for getting this together into a PR.  I just made a commit that merges master into this branch.

bin/celeste_mcmc.jl
```diff
@@ -0,0 +1,166 @@
+#!/usr/bin/env julia
+
+# ./celeste.jl infer-box 200 200.5 38.1 38.35
+using Distributions
+import Celeste: Model, Infer, RunCamcolField, load_images, init_source
+import Celeste.Model: CatalogEntry, TiledImage, ElboArgs,
+                      get_active_pixels
+
+# Mamba for MCMC tools
+import Mamba
```
rgiordan, line 10: `Mamba` needs to be added to `REQUIRE`.  Also, the fact that tests pass without it indicate that some core functionality (`run_single_source_sampler` I think) should is missing a test.

bin/celeste_mcmc.jl
```diff
@@ -0,0 +1,166 @@
...
+
+const MIN_FLUX = 2.0
```
rgiordan, line 12: Let's not define a constant `MIN_FLUX` in two separate places (here and `api.jl`).  If you like, maybe pass it as an argument to both functions.

bin/celeste_mcmc.jl
```diff
@@ -0,0 +1,166 @@
...
+
+type SampleResult
```
rgiordan, line 14: Please document the type.

bin/celeste_mcmc.jl
```diff
@@ -0,0 +1,166 @@
...
+    Infer.trim_source_tiles!(ea)
+    active_pixels = get_active_pixels(ea)
+
+    # generate the star logpdf and initial MCMC chain state
+    # states are packed together [lnr, cu, cg, cr, ci, ra, dec]
+    # TODO create a sample_from_star_prior function (over dispersion...)
+    star_logpdf, star_logprior =
+        Model.make_star_logpdf(images, active_pixels, ea)
+    star_state       = Model.init_star_state(entry)
+    star_param_names = ["lnr", "cug", "cgr", "cri", "ciz", "ra", "dec"]
```
rgiordan, line 56: Let's declare the param names in exportable, constant types, and document them.  Same for the galaxy types.

Better still, perhaps define parameter types and follow the convention we used to name the elements of variational parameter vectors?

bin/celeste_mcmc.jl
```diff
@@ -0,0 +1,166 @@
...
+    gal_state   = Model.init_galaxy_state(entry)
+    gal_param_names = [star_param_names;
+                       ["gdev", "gaxis", "gangle", "gscale"]]
+    gal_sim = run_slice_sampler(gal_logpdf, gal_state, 200, gal_param_names)
+
+    # generate pointers to star/gal type (to infer p(star | data))
+    #type_samples, type_lls = run_star_gal_switcher(star_lls, gal_lls)
+
+    # return all samples
+    #return SampleResult(star_samples, gal_samples, type_samples,
```
rgiordan, line 72: Do you mean for this to be commented out?

bin/celeste_mcmc.jl
```diff
@@ -0,0 +1,166 @@
...
+    #                    star_lls, gal_lls, type_lls)
+end
+
+
+"""
+Run a slice sampler for N steps
```
rgiordan, line 78: Please document.  (E.g., it is not easy to see at a glance what `th0` is.)

bin/celeste_mcmc.jl
```diff
@@ -0,0 +1,166 @@
...
+      sim[i, :, 1] = th
+    end
+    Mamba.describe(sim)
+    return sim
+end
+
+
+function run_star_gal_switcher(star_lls::Vector{Float64},
+                               gal_lls::Vector{Float64})
+    #TODO switch between states of the chain, alt to RJMCMC
```
rgiordan, line 102: You can put `error("unimplemented")` here and in other places that only have placeholders for now.

bin/celeste_mcmc.jl
```diff
@@ -0,0 +1,166 @@
...
+end
+
+
+function one_node_infer_mcmc(rcfs::Vector{RunCamcolField},
+                             stagedir::String;
+                             objid="",
+                             box=BoundingBox(-1000., 1000., -1000., 1000.),
+                             primary_initialization=true)
+    # catalog
+    duplicate_policy = primary_initialization ? :primary : :first
```
rgiordan, line 112: Let's not duplicate all this code between here and `api.jl`.  It is hard to maintain and error-prone.  Could you please wrap the common pre-processing into a single function?

bin/celeste_mcmc.jl
```diff
@@ -0,0 +1,166 @@
...
+    neighbors = [catalog[m] for m in neighbor_map[s]]
+
+    # generate samples for source entry/neighbors pair
+    samples = run_single_source_sampler(entry, neighbors, images)
+    return samples
+end
+
+
+# Main test entry point
+function run_gibbs_sampler_fixed()
```
rgiordan, line 156: You can put this in the testing file if it's only for testing.

src/ElboDeriv.jl
```diff
@@ -7,470 +7,16 @@ using ..Model
...
-            patches::Matrix{SkyPatch},
-            active_sources::Vector{Int})
-    N = length(images)
-    S = length(vp)
-    @assert length(tile_source_map) == N
-    @assert size(patches, 1) == S
-    @assert size(patches, 2) == N
-    ElboArgs(S, N, images, vp, tile_source_map, patches, active_sources)
-end
+export ElboArgs, ActivePixel
```
rgiordan, line 15: Why are you importing from `Model` and then exporting from here?

bin/celeste_mcmc.jl
```diff
@@ -0,0 +1,166 @@
+#!/usr/bin/env julia
```
rgiordan, line 1: Let's keep things in `bin` as thin wrappers around re-usable (and testable) functionality that is kept in modules.  Maybe an `CelesteMCMC` module?

src/Model.jl
```diff
@@ -56,4 +55,7 @@ include("wcs_utils.jl")
 typealias VariationalParams{NumType <: Number} Vector{Vector{NumType}}
 typealias FreeVariationalParams{NumType <: Number} Vector{Vector{NumType}}
 
+# log prob uses VariationalParams object --- included last
```
rgiordan, line 58: We can just change the order and define the `typealias` before all the includes.  That's probably better style.

src/elbo_kl.jl
```diff
@@ -5,6 +5,7 @@
 # Note that KL divergences may be between parameters of two different types,
 # e.g. if the prior is a float and the parameter is a dual number.
 
+import ..SensitiveFloats: SensitiveFloat
```
rgiordan, line 8: If this is necessary, let's put this `import` in `ElboDeriv.jl` -- dependencies should be declared in the modules, and included files like this should just be organizational devices.  

I notice that the existing file `param_set.jl` doesn't follow this rule, either.  Could you please fix that, too, while you're at it?

src/log_prob.jl
```diff
@@ -0,0 +1,438 @@
+# Functions to compute the log probability of star parameters and galaxy
+# parameters given pixel data
+
+using Distributions
+import ..SensitiveFloats: SensitiveFloat, zero_sensitive_float, clear!
```
rgiordan, line 5: Same as above -- please declare dependencies in the module, not in included files.

src/log_prob.jl
```diff
@@ -0,0 +1,438 @@
...
+
+include("bivariate_normals.jl")
+include("log_prob_util.jl")
+EPS = 1e-6
```
rgiordan, line 9: Please document.  What is this used for?  Do you mean it to be constant?

src/log_prob.jl
```diff
@@ -0,0 +1,438 @@
...
+function make_star_logpdf(images::Vector{TiledImage},
+                          active_pixels::Vector{ActivePixel},
+                          ea::ElboArgs)
+
+    # define star prior log probability density function
+    prior    = construct_prior()
+    subprior = prior.star
+
+    function star_logprior(state::Vector{Float64})
+        brightness, colors, u = state[1], state[2:5], state[6:end]
```
rgiordan, line 42: I think it's much better to define named types to get elements of a vector as we did for the VB parameters.  Referring to parameters by number can get quite error-prone in the long run.

You can also avoid memory allocation by simply using state[mcmc_ids.bright] as a function argument.

src/log_prob.jl
```diff
@@ -0,0 +1,438 @@
...
+        return color_logprior(brightness, colors, prior, true)
+    end
+
+    # define star log joint probability density function
+    function star_logpdf(state::Vector{Float64})
+        ll_prior = star_logprior(state)
+        brightness, colors, position = state[1], state[2:5], state[6:end]
+        dummy_gal_shape = [.1, .1, .1, .1]
```
rgiordan, line 50: I suspect you'll get memory allocation performance gains by defining a constant `dummy_gal_shape` outside the function.

src/log_prob.jl
```diff
@@ -0,0 +1,438 @@
...
+function make_galaxy_logpdf(images::Vector{TiledImage},
+                            active_pixels::Vector{ActivePixel},
+                            ea::ElboArgs)
+
+    # define star prior log probability density function
+    prior    = construct_prior()
+    subprior = prior.galaxy
+
+    function galaxy_logprior(state::Vector{Float64})
+        brightness, colors, u, gal_shape =
```
rgiordan, line 89: Same here as above.

src/log_prob.jl
```diff
@@ -0,0 +1,438 @@
...
+
+Args:
+  - is_star: bool describing the type of source
+  - brightness: log r-band value
+  - colors: array of colors
+  - position: ra/dec of source
+  - gal_shape: vector of galaxy shape params (used if is_star=false)
+  - images: vector of TiledImage types (data for likelihood)
+  - active_pixels: vector of ActivePixels over which the ll is summed
+  - ea: ElboArgs object that maintains params for all sources
```
rgiordan, line 126: If you're not using `active_sources`, please mention here that inference is only done for the first source in the `ElboArgs`.  (Though probably better to just use `active_sources`, also see comment below.)

src/log_prob.jl
```diff
@@ -0,0 +1,438 @@
...
+                              ea::ElboArgs)
+    # TODO: cache the background rate image!! --- does not need to be recomputed at each ll eval
+
+    # convert brightness/colors to fluxes for scaling
+    fluxes = colors_to_fluxes(brightness, colors)
+
+    # make sure elbo-args reflects the position and galaxy shape passed in for
+    # the first source in the elbo args (first is current source, the rest are
+    # conditioned on)
+    ea.vp[1][ids.u[1]]    = position[1]
```
rgiordan, line 149: `active_sources` instead

src/log_prob.jl
```diff
@@ -0,0 +1,438 @@
...
+        pixel_band   = tile.b
+
+        # compute the unit-flux pixel values
+        populate_fsm_vecs!(elbo_vars, ea, tile_sources, tile,
+                           pixel.h, pixel.w,
+                           gal_mcs_vec[pixel.n], star_mcs_vec[pixel.n])
+
+        # compute the background rate for this pixel
+        background_rate = tile.epsilon_mat[pixel.h, pixel.w]
+        background_sources = tile_sources[tile_sources.!=1]
```
rgiordan, line 188: If you want to be consistent with the semantics of the VB code, maybe use active_sources here instead of hard-coding that the active source is `1` together with an assert that the length of the active sources vector is one.

Also you can do this outside the loop.

src/log_prob.jl
```diff
@@ -0,0 +1,438 @@
...
+    is_star::Bernoulli
+    star::StarPrior
+    galaxy::GalaxyPrior
+end
+
+
+"""
+Construct a `Prior` object, contains StarPrior and GalaxyPrior objects
+"""
+function construct_prior()
```
rgiordan, line 242: It would be julionic (julicious?) to define this as the constructor for the `Prior` type, perhaps with `pp` as an argument with `load_prior()` as a default.

src/log_prob.jl
```diff
@@ -0,0 +1,438 @@
...
+# transformation of fluxes = (u, g, r, i, z) in nanomaggies to  #
+# (bright,color) pairs                                          #
+#################################################################
+
+
+"""
+Translate between fluxes (nmgy rates) and colors (on which
+LogNormal priors are placed)
+"""
+function fluxes_to_colors(fluxes::Vector{Float64})
```
rgiordan, line 298: Let's add a unit test that verifies that this and `colors_to_fluxes` match `SourceBrightness` when the variational variance is very small.

Extra nice might be to put these functions in `source_brightness.jl` so it's clear that if we change one we should change the other.

src/log_prob.jl
```diff
@@ -0,0 +1,438 @@
...
+
+#######################################################################
+# galaxy shape transformation (unconstrained => constrained and back) #
+#######################################################################
+
+
+function constrain_gal_shape(unc_gal_shape::Vector{Float64})
+    gdev, gaxis, gangle, gscale = unc_gal_shape
+    constr_shape    = Array(Float64, 4)
+    constr_shape[1] = clamp(sigmoid(gdev), EPS, 1-EPS)
```
rgiordan, line 333: This is fine, but just in case you want to avoid the nonlinearity of `clamp`, you could use `unbox_parameter`.

src/log_prob.jl
```diff
@@ -0,0 +1,438 @@
...
+    shared_params = elbo_args_vp_to_star_state(vp)
+    gal_shape = [vp[ids.e_dev]  ,
+                 vp[ids.e_axis] ,
+                 vp[ids.e_angle],
+                 vp[ids.e_scale]]
+    return [shared_params; unconstrain_gal_shape(gal_shape)]
+end
+
+
+function variational_params_to_fluxes(i::Int, vs::Array{Array{Float64,1}})
```
rgiordan, line 395: Let's not duplicate code -- let's use `SourceBrightness`, possibly with an added option not to calculate the expected squares.

src/log_prob.jl
```diff
@@ -0,0 +1,438 @@
...
+#TODO is there a better place for this generic function? --- acm
+function logsumexp(a::Vector{Float64})
+    a_max = maximum(a)
+    out = log(sum(exp(a - a_max)))
+    out += a_max
+    return out
+end
+
+
+function logit(a::Float64)
```
rgiordan, line 431: Let's just use the versions in `Transform.jl`.

src/log_prob.jl
```diff
@@ -0,0 +1,438 @@
...
+    return exp(ret)
+end
+
+
+#######################################################################
+# galaxy shape transformation (unconstrained => constrained and back) #
+#######################################################################
+
+
+function constrain_gal_shape(unc_gal_shape::Vector{Float64})
```
rgiordan, line 330: Maybe for consistency put these in `Transform.jl`.

src/log_prob.jl
```diff
@@ -0,0 +1,438 @@
...
+    return ll
+end
+
+
+###########################################################################
+# priors - for computing prior log probs of star/gal parameters           #
+###########################################################################
+
+
+type StarPrior
```
rgiordan, line 215: Better organization would be to call these `MCMCStarPrior` &c and put them in a `priors.jl` file with the other prior types, which are now in `light_source_model.jl`, probably by historical accident.

src/log_prob_util.jl
```diff
@@ -0,0 +1,490 @@
+###################################################################
```
rgiordan, line 1: Once you move the functions from `log_prob.jl` to live with their VB counterparts, there won't be much left, and you can just combine what's left with these functions in a single `log_prob.jl` file.  This contains more than utilities, IMO -- it's much of the log likelihood calculation and types.

test/test_log_prob.jl
```diff
@@ -0,0 +1,118 @@
+using Celeste: Model, SensitiveFloats, ElboDeriv, Infer
+using Base.Test
+"""
+test log_prob.jl and log_prob_util.jl
+"""
+
+####################
+# Helper functions #
+####################
+function true_star_init()
```
rgiordan, line 10: Let's not duplicate code.  You can put these helper functions in `DerivativeTestUtils` and maybe call it `TestUtils` instead if you like.

rgiordan (changes requested):
Hi Andy, this looks pretty good!  Here are some purely organizational comments to start.  The big things are:
- I think we need a test for the some actual MCMC samples.
- We should not duplicate code, particularly in the API, but also in some of the brightness functions.
- Let's get MCMC parameters out of vectors the same way we get variational parameters out of vectors, with a ParamSet.  I think it will be more readable and much less brittle in the long run.
- When possible, let's group the MCMC code (priors, transforms, etc) with the corresponding VB code.

I'll give it a deeper look over the next day or so.

Ryan

bin/celeste_mcmc.jl
jeff-regier: That's my fault, I just wanted to get Andy start quickly. But yeah, I think an `MCMC` module is the way to go.

bin/celeste_mcmc.jl
jeff-regier: Or alternatively we could just include the log probability function, and unit tests for it, in this PR. Either way. But I suppose if we're going to have the MCMC code in this PR, it should be committed with tests.

bin/celeste_mcmc.jl
jeff-regier: `Celeste.MIN_FLUX`

src/log_prob.jl
jeff-regier: Non-constant globals really slow things down in julia, for some reason. Try `const EPS = 1e-6`.

src/log_prob.jl
jeff-regier: Or by making it a tuple--those are stack allocated and vectors are heap allocated.

src/log_prob.jl
```diff
@@ -0,0 +1,438 @@
...
+            the rest of the args
+"""
+function state_log_likelihood(is_star::Bool,
+                              brightness::Float64,
+                              colors::Vector{Float64},
+                              position::Vector{Float64},
+                              gal_shape::Vector{Float64},
+                              images::Vector{TiledImage},
+                              active_pixels::Vector{ActivePixel},
+                              ea::ElboArgs)
```
jeff-regier, line 140: Probably `ElboArgs` shouldn't appear in `log_prob.jl`, or really any mention of the ELBO. The `ElboDeriv` module (and the forthcoming `MCMC` module) imports `Model`, not the other way around.

test/test_log_prob.jl
```diff
@@ -0,0 +1,118 @@
...
+    ea.vp[1][ids.a] = [ 1.0 - 1e-4, 1e-4 ]
+    ea.vp[1][ids.r2] = 1e-4
+    ea.vp[1][ids.r1] = log(sample_star_fluxes[3]) - 0.5 * ea.vp[1][ids.r2]
+    #ea.vp[1][ids.r1] = sample_star_fluxes[3] ./ ea.vp[1][ids.r2]
+    ea.vp[1][ids.c2] = 1e-4
+    blob, ea, body
+end
+
+
+function test_that_star_truth_is_most_likely()
```
jeff-regier, line 22: Looks good!

bin/celeste_mcmc.jl
jeff-regier: I agree, and probably `api.jl` should call the mcmc module rather than the other way around, to avoid spaghetti code.

src/log_prob.jl
jeff-regier: I'd put them somewhere in the `Model` module, iff they're getting used elsewhere too.

rgiordan:
Andy, are you making a new PR?  Any updates?

PR #302, opened 2016-09-16
 Simplex fixes

rgiordan:
Fix enforce_bounds for simplexes and make all simplex parameters matrices to avoid some code duplication.


jeff-regier:
Oh man, the `ids` data structure is getting even more complicated here. There's got to be a better way to do this....

jeff-regier:
I'll ask the Tony about how to do the indexing into the various types of parameter vectors, if I can think of how to phrase the question. I've never been happy with `ids`.

Glad `enforce_bounds!` is working again.

PR #309, opened 2016-09-21
added unit tests for score.jl

### LINKED ISSUE #282: Make score.jl a module and add unit tests for it

jeff-regier:
Closes #282.

@kpamnany -- I found and fixed an off-by-one error in `api.jl:one_node_infer()` while writing these unit tests. Without the fix, the last light source on each node won't get optimized, so please merge the fix into your `src_div_par` branch.


kpamnany:
This is correct in `source_division_inference.jl`.

PR #312, opened 2016-09-23
Refactor maximize elbo

jeff-regier:


jeff-regier:
@rgiordan -- It's already merged but if you have changes to suggest please let me know any I'll make them. I was pretty happy about this PR turned out.

PR #313, opened 2016-09-24
now list_rcfs.jl takes a bounding box

jeff-regier:
I modified `list_rcfs.jl` to take a bounding box as arguments, rather than returning all the rcfs every time. (Before I just needed a list of all rcfs for staging, but now I needed to get a dataset to Kiran for testing.)


rgiordan (commented):
Already done, but LGTM

PR #314, opened 2016-09-25
fix missing var in optimization verbose printing

gostevehoward:
Looks like a function was inlined and this parameter name wasn't updated. This is a minimal fix.


jeff-regier:
Looks good, thanks.

PR #315, opened 2016-09-26
Disallow epsilon = 0, and try to catch NaNs earlier

gostevehoward:
@jeff-regier What do you think of these? Does the check on epsilon seem reasonable? Is this a good place to check it? Are the asserts going to be too much of a performance hit? I'm not sure if Julia optimizes out asserts under certain conditions or what...I can research it.

### DIFF
--- a/src/Exceptions.jl
+++ b/src/Exceptions.jl
@@ -1,0 +1,14 @@
+"""
+Custom exception types used throughout Celeste.
+"""
+module Exceptions
+
+"""
+Some parameter to a function has invalid values. The message should explain what parameter is
+invalid and why.
+"""
+type InvalidInputError <: Exception
+    message::String
+end
+
+end
--- a/src/Celeste.jl
+++ b/src/Celeste.jl
@@ -1,6 +1,7 @@
 module Celeste
 
 # submodules
+include("Exceptions.jl")
 include("Log.jl")
 
 include("SensitiveFloats.jl")
--- a/src/DeterministicVI.jl
+++ b/src/DeterministicVI.jl
@@ -3,6 +3,7 @@ Calculate value, gradient, and hessian of the variational ELBO.
 """
 module DeterministicVI
 
+import ..Exceptions
 using ..Model
 import ..Model: BivariateNormalDerivatives, BvnComponent, GalaxyCacheComponent,
                 GalaxySigmaDerivs,
@@ -16,7 +17,6 @@ import DataFrames
 import Optim
 
 export ElboArgs
-
 
 """
 ElboArgs stores the arguments needed to evaluate the variational objective
@@ -59,6 +59,13 @@ function ElboArgs{NumType <: Number}(
     @assert length(tile_source_map) == N
     @assert size(patches, 1) == S
     @assert size(patches, 2) == N
+    for tiled_image in images
+        for tile in tiled_image.tiles
+            if any(tile.epsilon_mat .<= 0)
+                throw(Exceptions.InvalidInputError("You must set all values of epsilon_mat > 0"))
+            end
+        end
+    end
     ElboArgs(S, N, psf_K, images, vp, tile_source_map, patches,
              active_sources, num_allowed_sd)
 end
--- a/src/deterministic_vi/elbo.jl
+++ b/src/deterministic_vi/elbo.jl
@@ -693,6 +693,7 @@ function process_active_pixels!{NumType <: Number}(
         # the optimization convergence criterion, so I will leave it in for now.
         elbo_vars.elbo.v[1] -= lfact(this_pixel)
     end
+    assert_all_finite(elbo_vars.elbo)
 end
 
 
@@ -813,4 +814,12 @@ function elbo{NumType <: Number}(
     # TODO: subtract the kl with the hessian.
     subtract_kl!(ea, elbo, calculate_derivs=calculate_derivs)
     elbo
+end
+
+# If Infs/NaNs have crept into the ELBO evaluation (a symptom of poorly conditioned optimization),
+# this helps catch them immediately.
+function assert_all_finite{ParamType}(sf::SensitiveFloat{ParamType, Float64})
+    @assert all(isfinite(sf.v)) "Value is Inf/NaNs"
+    @assert all(isfinite(sf.d)) "Gradient contains Inf/NaNs"
+    @assert all(isfinite(sf.h)) "Hessian contains Inf/NaNs"
 end


jeff-regier:
@gostevehoward -- This code looks useful for debugging, but I don't think we should include it in master.

First, `epsilon=0` shouldn't be a problem for the code, because that just means there's no background noise---an easy case, not a bad one. If it is, a problem, we should fix that rather than check for it. (Does epsilon show up in a log term something?) 

There should never be NaNs (or Infs for that matter) in the elbo. We should always throw an exception right away in cases where either is inevitable, though I wouldn't expect either to be inevitable except in the case of pathological data. Ideally we'd through an exception for pathological data when we load the data, not once we're processing it unsuccessfully. Also, as you suspected, adding anything to those functions in elbo.jl is likely a bad performance hit.

I'm making some revisions to the `Transform` module now, that may catch some inevitable numerical issues at the right places. Also, I got the optimizer to be more robust to poorly conditioned hessians in my recent PR: JuliaOpt/Optim.jl#284  I don't think a release with this change is tagged yet, but you could try `Pkg.checkout("Optim", "master")`.

gostevehoward:
@jeff-regier 

Regarding `epsilon=0`: what happens is that in my test images, which (in simple cases) are dark except for a small, bright source at the middle, the expected intensity `E_G.v[1]` for certain pixels gets driven towards zero. This eventually causes a blowup: currently, the breaking point seems to be [dividing by `E_G.v[1]^4`](https://github.com/jeff-regier/Celeste.jl/blob/master/src/deterministic_vi/elbo.jl#L783), which causes an Inf in `combine_hess`, which propagates into a NaN in `elbo_log_term.h` (though I'm pretty sure the version I was working with on Sunday was breaking elsewhere, but the specific breaking point isn't really important, because there are many possible ones). When epsilon > 0, this is [prevented from happening](https://github.com/jeff-regier/Celeste.jl/blob/master/src/deterministic_vi/elbo.jl#L741).

Regarding assertions: I'm sure you guys already have opinions on this, so let me lay out my philosophy to make sure we're on the same page:

I think of assertions as being used to check conditions that should always be true in the code unless there is a bug in the code itself. Assertion failure => there is definitely a bug in our code. (You already have tons of assertions like this so I'm assuming you already take this approach too.) So checking for NaNs/Infs seems like a perfect use for assertions: it should never happen, but it will, because inevitably the code will have bugs from time to time (be they bugs in the optimization routine or bugs in the input validation), and when it does, we _do_ want the whole program to crash ASAP, and assertions will save developers many hours of tracking down root causes.

Input validation, or error handling more generally, is a completely different beast, something we do expect to happen in bug-free code when the user passes invalid input. In this case we raise an exception the user can handle however she chooses. Like you said, ideally we'd catch every possible input problem up front in validation, but inevitably there will be corner cases we miss, hence the assertions.

All of this is to say, I think the fact that we should catch the root causes in input validation doesn't mean we shouldn't also assert on key symptoms down the road. What is important is that an assertion firing means someone has to fix the code in some way, it doesn't just mean we change our inputs and move on.

Regarding performance: I was hoping Julia might do something like the old-fashioned C `#assert`, which would get compiled out on `gcc -O` and have zero performance impact. I did a little benchmark and Julia does not do that right now, even with `julia -O3`. However, there are options: we could define a global `const DEBUG = false`, and
- If you write a function ala
  
  ``` julia
  @inline function my_debug_check()
    if DEBUG
      # do expensive checks
    end
  end
  ```
  
  then it seems to have zero performance impact when `DEBUG == false`. (I found I did have to specify `@inline`, even when the check function was quite simple.)
- Or, if you're down with macros, a nicer option is along the lines of
  
  ``` julia
  macro my_assert(ex)
    if DEBUG
      return :( $ex ? nothing : throw(AssertionError($(string(ex)))) )
    else
      return :( nothing )
    end
  end
  ```

Obviously we could define different debug levels to trade of granularity of assertions and performance.

Now there is still the issue of all these assertions making the code ugly and harder to read, which is fair, but I think if used judiciously they could add a lot of value, especially for common failure modes. I got the impression from Ryan that this failure mode isn't uncommon.

If you're curious, here's the script I was using to play with assertion performance:

``` julia
const DEBUG = true

function fill_fibonacci!(array)
    array[1] = array[2] = 1
    for i in 3:length(array)
        array[i] = array[i-1] + array[i-2]
    end
end

macro my_assert(ex)
    if DEBUG
        return :( $ex ? nothing : throw(AssertionError($(string(ex)))) )
    else
        return :( nothing )
    end
end

function check_valid_fibonacci(array, max_index)
    for i in 3:max_index
        if array[i] != array[i-1] + array[i-2]
            return false
        end
    end
    return true
end

@inline function assert_valid_fibonacci(array, max_index)
    if DEBUG
        @assert check_valid_fibonacci(array, max_index)
    end
end

function fill_fibonacci_with_assertions!(array)
    array[1] = array[2] = 1
    for i in 3:length(array)
        array[i] = array[i-1] + array[i-2]
        assert_valid_fibonacci(array, i)
        #@my_assert check_valid_fibonacci(array, i)
    end
end

function main(size)
    values = fill(0, size)

    fill_fibonacci!(values)
    @time fill_fibonacci!(values)

    fill_fibonacci_with_assertions!(values)
    @time fill_fibonacci_with_assertions!(values)
end

main(1000000)
```

gostevehoward:
Also, I realized my own input validation (on epsilon) in this patch is using an @assert, and ought to throw some other kind of exception, according to my own advice :). I will make that change if we keep it.

rgiordan:
> @jeff-regier https://github.com/jeff-regier
> 
> Regarding epsilon=0: what happens is that in my test images, which (in
> simple cases) are dark except for a small, bright source at the middle, the
> expected intensity E_G.v[1] for certain pixels gets driven towards zero.
> This eventually causes a blowup: currently, the breaking point seems to be dividing
> by E_G.v[1]^4
> https://github.com/jeff-regier/Celeste.jl/blob/master/src/deterministic_vi/elbo.jl#L783,
> which causes an Inf in combine_hess, which propagates into a NaN in
> elbo_log_term.h (though I'm pretty sure the version I was working with on
> Sunday was breaking elsewhere, but the specific breaking point isn't really
> important, because there are many possible ones). When epsilon > 0, this is prevented
> from happening
> https://github.com/jeff-regier/Celeste.jl/blob/master/src/deterministic_vi/elbo.jl#L741
> .

That's interesting.  Ultimately, this is a failure of the poisson
approximation, which I continue to say we should really get rid of.  Our
model should not break if we evaluate it with a low expected intensity
where there is in fact light, we should just return a very low objective
function.  Unfortunately, this means passing around variance estimates,
which would increase the size of the image data structure quite a lot.
That's not a good idea right now before IPDPS, but I have it on my roadmap.
 (Many other unrelated benefits will accrue from switching to a normal
model.)

In the meantime, epsilon being zero is not the problem, the model is the
problem.  I do not think we should make an assert that is only there to
prevent unrelated numerical pathologies in the model.  I like the idea of
failing before the NaN gets passed to Optim, but I don't think asserting
that epsilon be greater than zero is the right fix.  (If you have this
problem, you should get a NaN, and increase epsilon if you think that's the
right way to hack around the model's shortcomings in your particular case.)

jeff-regier:
@rgiordan 

Are you pretty sure it's essential we divide by `E_G.v[1]`, or can we reorder the computations to avoid this? (I'm not sure either way, just asking here.) Does the true Hessian have, in the limit, infinite entries as the rate parameter goes to 0? If that's the case, reordering won't help, but if otherwise there should be a work around.

Regardless, there's always a non-zero change that a photon will land in a pixel, so I don't think it's a huge problem with the model if we require `epsilon > 0`. Not that there aren't other reasons from switching from a Poisson to a Gaussian model.

jeff-regier:
@gostevehoward 

I agree with your views on assertions: when used carefully, in places where they reveal problems much earlier, they're well worth it, even though they add clutter to the code. The `epsilon` check in the PR makes sense in light of that, well, as an exception rather than assertion, because it's a problem with the input rather than the code.

If you'd move the call to `assert_all_finite(elbo_vars.elbo)` to just after the loop over pixels, rather than calling it within the loop, I don't think we'd need to worry about performance. (Generally the loops over pixels are the hotspots.) Would that provide feedback early enough about the problem? The debug macro is a decent solution too, but it might be worth hearing about a problem slightly later in order to avoid introducing yet another flag.

I'd rather fix the code than add assertions of course, but maybe that's not the choice. I don't think these NaNs have to be constant problem, or even a problem that keeps reappearing as the code changes. Please go ahead merge this PR whenever you're happy with it.

rgiordan:
The term in the Hessian in question has the form `Var(G) / E[G] ^ 4`.  Here, `G` is modeled as a lognormal.  The lognormal's variance is proportional to the mean squared.  So yes, our parameterization allows this term to go to infinity.

This seems to make some intuitive sense to me.  We are approximating the distribution observed pixels with a Poisson distribution whose variance is the same as its mean.  Its mean cannot be driven to zero without any non-zero observation having zero probability (and associated extreme curvature in any direction that makes it nonzero).  But we don't really believe this, and there are basically two ways to resolve it:
-  Say no observation is truly zero (keep epsilon > 0)
-  Say our model never truly has zero variance (use a normal rather than a Poisson model, or otherwise add "dark variance")

It's not obvious to me that Celeste cannot work with epsilon == 0, only that in some cases you get numerical problems with a poorly modeled object when it is.  (Please correct me if I'm mistaken.)  As such, I would not want to enforce epsilon == 0 in the code but rather let the user set it to something other than zero to improve misfits.  But if that's a minority opinion I'll defer to you guys.

jeff-regier:
Steve and I just talked. Another problem we thought of is that if the
Poisson rate parameter is 0, but some x_nbm > 0 (e.g. 1 photon), then the
model log probability is negative infinity. So I think we're leaning
towards requiring `epsilon > 0` (or better yet, for numerical stability,
`epsilon > 1e-4`). That condition is sufficient (but not necessary, as you
note) to keep the elbo well defined.

rgiordan:
On Wed, Sep 28, 2016 at 10:43 AM, Jeffrey Regier notifications@github.com
wrote:

> Steve and I just talked. Another problem we thought of is that if the
> Poisson rate parameter is 0, but some x_nbm > 0 (e.g. 1 photon), then the
> model log probability is negative infinity.

Yes, this is what I was trying to say.

> So I think we're leaning
> towards requiring `epsilon > 0` (or better yet, for numerical stability,
> `epsilon > 1e-4`). That condition is sufficient (but not necessary, as you
> note) to keep the elbo well defined.

...and yet I came to the opposite conclusion.  :)  But like I said, I'll
defer if you both agree, I don't think it's that important to allow epsilon
= 0.  (There is a flag now that allows you to render images without it by
simply not adding it to the expected brightness, which is good enough for
my purposes.)

gostevehoward:
Here's a cleaned up version of those commits. I am enforcing epsilon > 0 per discussion with @jeff-regier . I think the logic was that maybe the model should be changed, but the model as it stands is liable to crash with epsilon == 0, so we should disallow it.

Maybe trimming tiles is actually the right answer, and once you trim tiles, the problem goes away. I'm learning about that now.

I tried one-off timed runs with the assertion in place and commented out and didn't observe a slowdown, but one-off timed runs are unreliable. Is there a more reliable perf benchmark to check such things?

gostevehoward:
FWIW, small data point, just adding a call to trim_source_tiles!() does not avert the NaNs when epsilon is set to zero. Not surprising, I think, since the threshold in trim_source_tiles! involves [multiplying by epsilon](https://github.com/jeff-regier/Celeste.jl/blob/master/src/Infer.jl#L246), but I can't say I completely understand that function.

jeff-regier:
Right now it looks like bright pixels are those that exceed some fraction of `epsilon` (the sky intensity). @rgiordan -- Wouldn't it make more sense for them to exceed some multiple of `sqrt(epsilon)`, since this is a Poisson model? e.g., 2 photons when the sky intensity is 1 isn't very bright, but 2000 photons when sky level is 1000 is clearly a star or a galaxy.

We could also add a bit of some smoothing inside the square root, e.g. + 1 or +10, at the cost of making the code a bit less intuitive.

src/Exceptions.jl
```diff
@@ -0,0 +1,14 @@
+"""
+Custom exception types used throughout Celeste.
+"""
+module Exceptions
+
+"""
+Some parameter to a function has invalid values. The message should explain what parameter is
+invalid and why.
+"""
+type InvalidInputError <: Exception
```
jeff-regier, line 10: Could we just define this type around where it's thrown, rather than creating a new file? A new module with just one type seems like a bit of a heavy weight solution.

PR #318, opened 2016-09-28
Set psf k

rgiordan:
These are a couple small changes that are necessary for using convolution to model the PSF.

It is possible that we can get substantial performance gains even with the model as-is by setting `num_allowed_sd` to 2 or 3, but I haven't explored it systematically yet so I'll leave it as `Inf` for now.

### DIFF
--- a/bin/celeste.jl
+++ b/bin/celeste.jl
@@ -39,11 +39,8 @@ function main()
 #    set_logging_level(args["--logging"])
     stagedir = ENV["CELESTE_STAGE_DIR"]
     if args["infer-box"]
-        ramin = parse(Float64, args["<ramin>"])
-        ramax = parse(Float64, args["<ramax>"])
-        decmin = parse(Float64, args["<decmin>"])
-        decmax = parse(Float64, args["<decmax>"])
-        box = BoundingBox(ramin, ramax, decmin, decmax)
+        box = BoundingBox(args["<ramin>"], args["<ramax>"],
+                          args["<decmin>"], args["<decmax>"])
         outdir = args["<outdir>"]
         infer_box(box, stagedir, outdir)
     else
--- a/src/DeterministicVI.jl
+++ b/src/DeterministicVI.jl
@@ -16,16 +16,23 @@ export ElboArgs
 
 """
 ElboArgs stores the arguments needed to evaluate the variational objective
-function
+function.
 """
 type ElboArgs{NumType <: Number}
     S::Int64
     N::Int64
+
+    # The number of components in the point spread function.
+    psf_K::Int64
     images::Vector{TiledImage}
     vp::VariationalParams{NumType}
     tile_source_map::Vector{Matrix{Vector{Int}}}
     patches::Matrix{SkyPatch}
     active_sources::Vector{Int}
+
+    # The number of standard deviations away below which a bivariate normal
+    # will be calculated.  Set to Inf to evaluate bivariate normals at all points.
+    num_allowed_sd::Float64
 end
 
 
@@ -34,13 +41,18 @@ function ElboArgs{NumType <: Number}(
             vp::VariationalParams{NumType},
             tile_source_map::Vector{Matrix{Vector{Int}}},
             patches::Matrix{SkyPatch},
-            active_sources::Vector{Int})
+            active_sources::Vector{Int},
+            psf_K::Int,
+            num_allowed_sd::Float64)
     N = length(images)
     S = length(vp)
+
+    @assert psf_K > 0
     @assert length(tile_source_map) == N
     @assert size(patches, 1) == S
     @assert size(patches, 2) == N
-    ElboArgs(S, N, images, vp, tile_source_map, patches, active_sources)
+    ElboArgs(S, N, default_psf_K, images, vp, tile_source_map, patches,
+             active_sources, num_allowed_sd)
 end
 
 
--- a/src/DeterministicVI.jl
+++ b/src/DeterministicVI.jl
@@ -30,8 +30,12 @@ type ElboArgs{NumType <: Number}
     patches::Matrix{SkyPatch}
     active_sources::Vector{Int}
 
-    # The number of standard deviations away below which a bivariate normal
-    # will be calculated.  Set to Inf to evaluate bivariate normals at all points.
+    # Bivarite normals will not be evaulated at points further than this many
+    # standard deviations away from their mean.  See its usage in the ELBO and
+    # bivariate normals for details.
+    #
+    # If this is set to Inf, the bivariate normals will be evaluated at all points
+    # irrespective of their distance from the mean.
     num_allowed_sd::Float64
 end
 
@@ -41,9 +45,9 @@ function ElboArgs{NumType <: Number}(
             vp::VariationalParams{NumType},
             tile_source_map::Vector{Matrix{Vector{Int}}},
             patches::Matrix{SkyPatch},
-            active_sources::Vector{Int},
-            psf_K::Int,
-            num_allowed_sd::Float64)
+            active_sources::Vector{Int};
+            psf_K::Int=2,
+            num_allowed_sd::Float64=Inf)
     N = length(images)
     S = length(vp)
 
@@ -51,7 +55,7 @@ function ElboArgs{NumType <: Number}(
     @assert length(tile_source_map) == N
     @assert size(patches, 1) == S
     @assert size(patches, 2) == N
-    ElboArgs(S, N, default_psf_K, images, vp, tile_source_map, patches,
+    ElboArgs(S, N, psf_K, images, vp, tile_source_map, patches,
              active_sources, num_allowed_sd)
 end
 
--- a/src/deterministic_vi/elbo.jl
+++ b/src/deterministic_vi/elbo.jl
@@ -222,7 +222,8 @@ Args:
     - bmc: The component to be added
     - x: An offset for the component in pixel coordinates (e.g. a pixel location)
     - wcs_jacobian: The jacobian of the function pixel = F(world) at this location.
-    - calculate_derivs: Whether to calculate derivatives.
+    - is_active_source: Whether it is an active source, (i.e. whether to
+                        calculate derivatives if requested.)
 
 Returns:
     Updates elbo_vars.fs0m_vec[s] in place.
@@ -233,17 +234,17 @@ function accum_star_pos!{NumType <: Number}(
                     bmc::BvnComponent{NumType},
                     x::Vector{Float64},
                     wcs_jacobian::Array{Float64, 2},
-                    calculate_derivs::Bool)
+                    is_active_source::Bool)
     eval_bvn_pdf!(elbo_vars.bvn_derivs, bmc, x)
 
-    if elbo_vars.calculate_derivs && calculate_derivs
+    if elbo_vars.calculate_derivs && is_active_source
         get_bvn_derivs!(elbo_vars.bvn_derivs, bmc, true, false)
     end
 
     fs0m = elbo_vars.fs0m_vec[s]
     fs0m.v[1] += elbo_vars.bvn_derivs.f_pre[1]
 
-    if elbo_vars.calculate_derivs && calculate_derivs
+    if elbo_vars.calculate_derivs && is_active_source
         transform_bvn_ux_derivs!(
             elbo_vars.bvn_derivs, wcs_jacobian, elbo_vars.calculate_hessian)
         bvn_u_d = elbo_vars.bvn_derivs.bvn_u_d
@@ -277,7 +278,8 @@ Args:
     - gcc: The galaxy component to be added
     - x: An offset for the component in pixel coordinates (e.g. a pixel location)
     - wcs_jacobian: The jacobian of the function pixel = F(world) at this location.
-    - calculate_derivs: Whether to calculate derivatives.
+    - is_active_source: Whether it is an active source, (i.e. whether to
+                        calculate derivatives if requested.)
 
 Returns:
     Updates elbo_vars.fs1m_vec[s] in place.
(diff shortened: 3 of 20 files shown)


bin/celeste.jl
```diff
@@ -39,11 +39,8 @@ function main()
 #    set_logging_level(args["--logging"])
     stagedir = ENV["CELESTE_STAGE_DIR"]
     if args["infer-box"]
-        ramin = parse(Float64, args["<ramin>"])
-        ramax = parse(Float64, args["<ramax>"])
-        decmin = parse(Float64, args["<decmin>"])
-        decmax = parse(Float64, args["<decmax>"])
-        box = BoundingBox(ramin, ramax, decmin, decmax)
+        box = BoundingBox(args["<ramin>"], args["<ramax>"],
```
rgiordan, line 42: That it picked up this change of yours may be an artifact of the `rebase -i` squash that I don't really understand.

src/DeterministicVI.jl
```diff
@@ -16,16 +16,23 @@ export ElboArgs
...
 ElboArgs stores the arguments needed to evaluate the variational objective
-function
+function.
 """
 type ElboArgs{NumType <: Number}
     S::Int64
     N::Int64
+
+    # The number of components in the point spread function.
+    psf_K::Int64
```
jeff-regier, line 26: Isn't the number of psf components already known from the PSF itself? It's good to avoid declaring the same constant 2 places, so there's no potential for disagreement.

bin/celeste.jl
jeff-regier: Yeah `rebase` is still something of a mystery to me too.

src/DeterministicVI.jl
```diff
@@ -16,16 +16,23 @@ export ElboArgs
...
     images::Vector{TiledImage}
     vp::VariationalParams{NumType}
     tile_source_map::Vector{Matrix{Vector{Int}}}
     patches::Matrix{SkyPatch}
     active_sources::Vector{Int}
+
+    # The number of standard deviations away below which a bivariate normal
+    # will be calculated.  Set to Inf to evaluate bivariate normals at all points.
+    num_allowed_sd::Float64
```
jeff-regier, line 39: Not sure I understand the comment. This is about truncating bvnormal distributions?

jeff-regier: Is dynamically truncating the Gaussians going to lead to a non-continuous objective function? Currently we're only truncating before the optimization runs, to decide what pixels to consider for each source. That way the objective is continuous once we're optimizing.

src/DeterministicVI.jl
jeff-regier: Isn't the PSF already fixed before the elbo is called?

src/deterministic_vi/elbo.jl
```diff
@@ -235,11 +235,12 @@ function accum_star_pos!{NumType <: Number}(
                     x::Vector{Float64},
                     wcs_jacobian::Array{Float64, 2},
                     calculate_derivs::Bool)
+
     eval_bvn_pdf!(elbo_vars.bvn_derivs, bmc, x)
 
-    # TODO: Also make a version that doesn't calculate any derivatives
-    # if the object isn't in active_sources.
-    get_bvn_derivs!(elbo_vars.bvn_derivs, bmc, true, false)
+    if elbo_vars.calculate_derivs && calculate_derivs
```
jeff-regier: Why are there 2 different `calculate_derivs` booleans? Can we use just one?

src/DeterministicVI.jl
jeff-regier: I see, you're preparing for learning the psf within the optimization procedure. Are you adding the rest of the psf parameters to ElboArgs then too?

src/DeterministicVI.jl
```diff
@@ -34,13 +41,18 @@ function ElboArgs{NumType <: Number}(
             vp::VariationalParams{NumType},
             tile_source_map::Vector{Matrix{Vector{Int}}},
             patches::Matrix{SkyPatch},
-            active_sources::Vector{Int})
+            active_sources::Vector{Int},
+            psf_K::Int,
```
jeff-regier: Do we need a default value for `psf_K`? If so, would you make `psf_K` a keyword argument with a default value, rather than having a constant named `default_psfK`?

test/test_slow_derivatives.jl
```diff
@@ -66,7 +66,7 @@ function test_real_image()
     cat_local = vcat(catalog[sa], catalog[neighbors])
     vp = Vector{Float64}[init_source(ce) for ce in cat_local]
     patches, tile_source_map = Infer.get_tile_source_map(tiled_images, cat_local)
-    ea = ElboArgs(tiled_images, vp, tile_source_map, patches, [1])
+    ea = ElboArgs(tiled_images, vp, tile_source_map, patches, [1], default_psf_K, Inf)
```
jeff-regier: How about making `Inf` the default for an optional keyword argument to the  `ElboArgs` constructor, in addition to `default_psf_k`?

test/test_elbo_values.jl
```diff
@@ -86,7 +86,7 @@ function test_derivative_flags()
     elbo = DeterministicVI.elbo(ea)
 
     elbo_noderiv = DeterministicVI.elbo(ea; calculate_derivs=false)
-    @test_approx_eq elbo.v[1] elbo_noderiv.v
+    @test_approx_eq elbo.v[1] elbo_noderiv.v[1]
```
jeff-regier, line 89: Why is `.v` an array?

src/model/psf_model.jl
```diff
@@ -1,5 +1,5 @@
-# The number of Gaussian components in the PSF.
-const psf_K = 2
+# The default number of Gaussian components in the PSF.
+const default_psf_K = 2
```
jeff-regier: By putting the default in the constructor to `ElboArgs`, there's no risk that code will call `default_psf_K` rather than `ea.psf_K`--potentially a difficult to detect bug.

src/deterministic_vi/elbo.jl
```diff
@@ -396,6 +397,7 @@ function populate_fsm_vecs!{NumType <: Number}(
                     h::Int, w::Int,
                     gal_mcs::Array{GalaxyCacheComponent{NumType}, 4},
                     star_mcs::Array{BvnComponent{NumType}, 2})
+
```
jeff-regier: We usually start the code right away after the function declaration

src/deterministic_vi/elbo.jl
```diff
@@ -406,20 +408,27 @@ function populate_fsm_vecs!{NumType <: Number}(
         calculate_hessian =
             elbo_vars.calculate_hessian && elbo_vars.calculate_derivs && active_source
         clear!(elbo_vars.fs0m_vec[s], calculate_hessian)
-        for k = 1:psf_K # PSF component
-            accum_star_pos!(
-                elbo_vars, s, star_mcs[k, s], x, wcs_jacobian, active_source)
+        for k = 1:ea.psf_K # PSF component
+            if (ea.num_allowed_sd == Inf ||
+                check_point_close_to_bvn(star_mcs[k, s], x, ea.num_allowed_sd))
```
jeff-regier, line 412: Isn't this going to lead to discontinuities? I think we want to continue to fix the subimage around each celestial body before the optimization.

src/DeterministicVI.jl
rgiordan: Yes, you could get away without this.  You could, for example, get the number of psf components from each patch in `fit_object_psfs!`, get it from the dimension of `star_mcs` and `gal_mcs` in `populate_fsm_vecs`, pass it into `fit_object_psfs!` as an argument, &c.  But then, the same could be said of `ElboArgs.S` and `ElboArgs.N`.  This is readable and consistent with the paradigm that model parameters go in `ElboArgs`.

rgiordan: In fact, this is preparing for representing the PSF with a convolution, which requires a PSF_free image of the sky that can be convolved with a pixelated PSF.  Furthermore, in the PSF-free image, a star cannot be rendered as a single point, since then any `u` derivative will be zero (the image as a function of the location of the star will be discontinuous).  So I'm rendering the PSF-free image with a single PSF component that is about a half-pixel wide and truncating the BVN evaluations at a small number of standard deviations.

This is all entirely separate from inferring the PSF from the image.  I will do that in a future CL with a different data structure for shared parameters.

src/DeterministicVI.jl
rgiordan: In theory it results in a discontinuous objective function.  In practice everything works fine even with pretty severe truncations (e.g. two standard deviations).  Above, e.g., 5 standard deviations, it becomes numerically continuous since the contribution of the remaining pixels are below the tolerance of floating point arithmetic.

Anyhow, this kind of concern is why I allowed the user to specify it and set it to `Inf`.

src/deterministic_vi/elbo.jl
rgiordan: No.  If you look at how `calculate_derivs` is defined, it is determined by whether this is an active source.  The variable `elbo_vars.calcualte_derivs` determines whether we are calculating derivatives at all in this ELBO evaluation.  Please compare with the identical usage five lines later.

src/DeterministicVI.jl
rgiordan: Yes, we could do it that way if you prefer.

test/test_slow_derivatives.jl
rgiordan: No problem.

test/test_elbo_values.jl
rgiordan: Recall that when I went through the ELBO very carefully last semester looking for everywhere that memory was being allocated, every time the `Float64` value `v` was being set in a sensitive float, it was allocating memory.  Changing to an array prevented quite a lot of memory allocation.

I agree that this is strange and maybe it's not the case anymore.

src/model/psf_model.jl
rgiordan: Agreed.

src/deterministic_vi/elbo.jl
rgiordan: Done

src/deterministic_vi/elbo.jl
rgiordan: See my other comment.

src/DeterministicVI.jl
rgiordan: Does the updated comment in the new push look good to you?

rgiordan:
@jeff-regier , did you mean to open a review?  (What I mean to ask is, is it ok if I merge?)

src/DeterministicVI.jl
jeff-regier: Are you sure two standard deviations is enough? Back when I was using `NLopt` I had serious convergence problems with discontinuities in the objective created by dynamically truncating the PSF. Now, NLopt was not particularly robust, but nonetheless, this seems really risky to me. If it works, it only works for a particular optimizer, not optimizers in general. Some are very sensitive to discontinuities. Why would we voluntarily introduce discontinuities? What's wrong with fixing the active pixels a priori?

src/DeterministicVI.jl
jeff-regier: OK

src/deterministic_vi/elbo.jl
jeff-regier: Could we change the name of the local `calculate_derivs` to something like `is_active_source`?

jeff-regier (commented):
The PR looks good, but please review my comment about discontinuities.

src/DeterministicVI.jl
jeff-regier: Yes, please.

src/DeterministicVI.jl
jeff-regier: I don't think I see the new comment. It's the "away below which" part that's confusing.

rgiordan: > Why would we voluntarily introduce discontinuities?
> Because it saves a considerable amount of computation.
> 
> What's wrong with fixing the active pixels a priori?
> We do fix the active pixels a priori.  Maybe you mean to have a set of active pixels per source?  That's a good idea, but a more intrusive change.  Even with either of these plans, this is not a no-op, since a single multivariate Gaussian may not contribute meaningfully to a large number of pixels.

Re: the number of standard deviations and different optimizers, these questions can only be answered by experience.  That is why you can set it in ElboArgs and it is Inf by default.  Suffice to say unit tests pass (and pass faster) when I set this to four.

src/deterministic_vi/elbo.jl
rgiordan: "calculate_derivs" means what it says.  As used now, we determine whether to calculate_derivs based on whether it is an active source.  That's not necessarily the case.

This is the way it's always been.  Why do you want to change it now?

jeff-regier: As is, the derivative isn't necessarily calculated when the local  `calculate_derivs` is true. So it's a bad or at least misleading variable name. I like `is_active_source`, because it explains why you're overriding `elbo_vars.calculate_derivs`. Other names could work too.

src/DeterministicVI.jl
jeff-regier: I'm wary of settings that give developers the option of creating difficulty-to-track-down optimization bugs, but I do see the benefit. How much fast are the unit tests? Could we get the same benefit from more aggressive a prior trimming?

Another idea: I wonder if we could add some smoothing to a truncated Gaussian, that interpolates between a few pixels on either side of the truncation boundary. Then we'd still have a continuous function, but finite support. We might then be able to set `num_allowed_sd` very aggressively, since the brightest pixels are usually way more informative than the farther out pixels.

rgiordan: These are all interesting ideas.

But as-is, this is a non-breaking, unit-tested experimental feature that leaves default functionality the same.  It is a potential fix to a potential problem that I just want to investigate as part of a bigger (and more pressing) problem.  I am not sure why you're pushing back so hard.  I think the potential downsides are clear as are the potential upsides and it is worth investigating.

Could I please just try it out?  Or should I just be forking Celeste so I can try stuff like this?

src/deterministic_vi/elbo.jl
rgiordan: Ok, changed.

jeff-regier (approved):

PR #319, opened 2016-09-28
Refactor tranforms3

jeff-regier:
This PR isn't very substantive. I spent awhile trying to rid of the `Dict` in the `ParamBounds` structure, but that didn't lead to a huge reduction in memory use, so I haven't included it here. I'm just including some changes I made early while getting familiar with `Transform` that make it a bit more readable.

### DIFF
--- a/src/Transform.jl
+++ b/src/Transform.jl
@@ -25,23 +25,11 @@ function inv_logit{NumType <: Number}(x::NumType)
 end
 
 
-function inv_logit{NumType <: Number}(x::Array{NumType})
-    @assert(all(x .>= 0))
-    @assert(all(x .<= 1))
-    -log(1.0 ./ x - 1)
-end
-
-
 """
 Convert x in R to lie in the unit interval.
 """
 function logit{NumType <: Number}(x::NumType)
     1.0 / (1.0 + exp(-x))
-end
-
-
-function logit{NumType <: Number}(x::Array{NumType})
-    1.0 ./ (1.0 + exp(-x))
 end
 
 
--- a/src/Transform.jl
+++ b/src/Transform.jl
@@ -39,8 +39,9 @@ the last entry implicitly has the untransformed value 1.
 """
 function constrain_to_simplex{NumType <: Number}(x::Vector{NumType})
     if any(x .== Inf)
-        # Is there a bug here? If more than 1 entry in x is Inf, that just means
-        # that the last entry in z is 0, not the both entries in z are the same.
+        # If more than 1 entry in x is Inf, it may be because the
+        # the last entry in z is 0. Here we set all those entries to the
+        # same value, though that may not be strictly correct.
         z = NumType[ x_entry .== Inf ? one(NumType) : zero(NumType) for x_entry in x]
         z ./ sum(z)
         push!(z, 0)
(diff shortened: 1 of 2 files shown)


src/Transform.jl
```diff
@@ -25,13 +25,6 @@ function inv_logit{NumType <: Number}(x::NumType)
 end
 
 
-function inv_logit{NumType <: Number}(x::Array{NumType})
```
jeff-regier, line 28: You can get a vectorized version of inv_logit(x) by calling the scalar version of the dot notation: i.e. `inv_logit.(x)`

src/Transform.jl
```diff
@@ -40,17 +33,14 @@ function logit{NumType <: Number}(x::NumType)
...
-
-
 """
 Convert an (n - 1)-vector of real numbers to an n-vector on the simplex, where
 the last entry implicitly has the untransformed value 1.
 """
 function constrain_to_simplex{NumType <: Number}(x::Vector{NumType})
     if any(x .== Inf)
+        # Is there a bug here? If more than 1 entry in x is Inf, that just means
+        # that the last entry in z is 0, not the both entries in z are the same.
```
jeff-regier: Oh yeah, I wanted to ask you about this.

src/Transform.jl
rgiordan: Good to know!

src/Transform.jl
rgiordan: Since infinity is unordered, the only other solution is to throw an error.  Since you can't always control what an optimization algorithm does, that's not very desirable.  What do you think would be better?

rgiordan (changes requested):
This all looks good.  Let's resolve the comment about simplexifying multiple infinities before committing though.

src/Transform.jl
jeff-regier: I think `Inf`s are always bad news, even just one, a `NaN` waiting to happen downstream. Even if `Inf` is the "right" answer, you've got to worry about numerical stability. Two `Inf`s here would be especially problematic. Maybe we should enforce box constraints (for both simplex variables and others) by excluding out-of-bounds regions of the domain from the trust region. We could continue to use the `simplex_box` transform. Exponentiation leads to some really poorly conditioned parts of the domain. What do you think?

rgiordan: I think you are proposing a pre-filtering step that maps unbounded floats into some box of legal values.  That is the kind of thing that causes many optimization routines to crash (e.g. BFGS) since it effectively produces zero gradients where the routine expects nonzero gradients.  It says that the ELBO is constant outside a certain range with a derivative discontinuity at the edge of the box.  If you want to give that a try, why not do it in a separate CL and test on a bunch of objects?

Here's my reasoning about the current logic.  When an optimization routine passes a pair of `Inf``s, what should Celeste do?  I think it should gracefully choose a legal constrained parameter value, pass it to the optimizer, and let the model say that it is an extremely bad set of parameters.  That is what this does (and what it has been doing, just fine, for many months.

One potential problem with `Inf` is that finite gradient steps cannot get away from them.  If we find that this is a real problem, it seems to me that there are more graceful solutions than a bounding box (like letting the rest of the model be fit and then restarting with finite values).  I would probably cross this bridge when we come to it, but if you feel like working on it now, feel free.

jeff-regier: I was proposing modifying `Optim.jl` to enforce box constraints itself, for the Newton trust region method, then no longer using doing the `vp_to_unconstrain` stuff (for box constrains, the simplex constraints are another matter). I think box constraints can be enforced by excluding out-of-bounds points from the trust region at every iteration. We've probably got more pressing things to work on now though...

rgiordan (approved):

PR #326, opened 2016-09-29
Logprob function

andymiller:
Hi all -- I separated out all of the `logprob` stuff from the `mcmc` stuff.  This PR adds a star and galaxy log prob function.  I refactored some of `fsm`-related code from the `DeterministicVI` module, but left appropriately typed wrappers in their place.

### DIFF
--- a/src/model/fsm_util.jl
+++ b/src/model/fsm_util.jl
@@ -1,0 +1,406 @@
+#TODO RE Document this function
+"""
+Convolve the current locations and galaxy shapes with the PSF.  If
+calculate_derivs is true, also calculate derivatives and hessians for
+active sources.
+
+Args:
+ - psf: A vector of PSF components
+ - ea: The current ElboArgs
+ - b: The current band
+ - calculate_derivs: Whether to calculate derivatives for active sources.
+
+Returns:
+ - star_mcs: An array of BvnComponents with indices
+    - PSF component
+    - Source (index within active_sources)
+ - gal_mcs: An array of BvnComponents with indices
+    - PSF component
+    - Galaxy component
+    - Galaxy type
+    - Source (index within active_sources)
+  Hessians are only populated for s in ea.active_sources.
+"""
+function load_bvn_mixtures{NumType <: Number}(
+                    S::Int64,
+                    patches::Matrix{SkyPatch},
+                    vp::VariationalParams{NumType},
+                    active_sources::Vector{Int},
+                    b::Int;
+                    calculate_derivs::Bool=true,
+                    calculate_hessian::Bool=true)
+
+    star_mcs = Array(BvnComponent{NumType}, psf_K, S)
+    gal_mcs = Array(GalaxyCacheComponent{NumType}, psf_K, 8, 2, S)
+
+    # TODO: do not keep any derviative information if the sources are not in
+    # active_sources.
+    for s in 1:S
+        psf = patches[s, b].psf
+        vs = vp[s]
+
+        world_loc = vs[[ids.u[1], ids.u[2]]]
+        m_pos = Model.linear_world_to_pix(patches[s, b].wcs_jacobian,
+                                          patches[s, b].center,
+                                          patches[s, b].pixel_center, world_loc)
+
+        # Convolve the star locations with the PSF.
+        for k in 1:psf_K
+            pc = psf[k]
+            mean_s = [pc.xiBar[1] + m_pos[1], pc.xiBar[2] + m_pos[2]]
+            star_mcs[k, s] =
+              BvnComponent{NumType}(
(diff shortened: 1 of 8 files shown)


src/model/fsm_util.jl
```diff
@@ -0,0 +1,432 @@
...
+    # Vectors of star and galaxy bvn quantities from all sources for a pixel.
+    # The vector has one element for each active source, in the same order
+    # as ea.active_sources.
+
+    # TODO: you can treat this the same way as E_G_s and not keep a vector around.
+    fs0m_vec::Vector{SensitiveFloat{StarPosParams, NumType}}
+    fs1m_vec::Vector{SensitiveFloat{GalaxyPosParams, NumType}}
+
+    # Brightness values for a single source
+    E_G_s::SensitiveFloat{CanonicalParams, NumType}
```
jeff-regier: Does `Model` use `E_G_s`? Please remove the unused fields.

src/model/fsm_util.jl
```diff
@@ -0,0 +1,432 @@
...
+
+    # Pre-allocated memory for the gradient and Hessian of combine functions.
+    combine_grad::Vector{NumType}
+    combine_hess::Matrix{NumType}
+
+    # A placeholder for the log term in the ELBO.
+    elbo_log_term::SensitiveFloat{CanonicalParams, NumType}
+
+    # The ELBO itself.
+    elbo::SensitiveFloat{CanonicalParams, NumType}
```
jeff-regier: Is `elbo` used? If so, maybe rename it to something like `logprob`?

src/model/log_prob.jl
```diff
@@ -0,0 +1,443 @@
...
+#  - images: Vector of TiledImage types (data for log_likelihood)
+#  - active_pixels: Vector of ActivePixels on which the log_likelihood is based
+#  - ea: ElboArgs book keeping argument
+#Returns:
+#  - gal_logpdf  : unnormalized logpdf function handle that takes in a flat,
+#                 unconstrained array as parameter
+#  - gal_logprior: star param log prior function handle that takes in same
+#                 flat, unconstrained array as parameter
+#"""
+#function make_galaxy_logpdf(images::Vector{TiledImage},
```
jeff-regier: Please omit the commented out code from the commit.

src/model/log_prob.jl
```diff
@@ -0,0 +1,443 @@
...
+    shared_params = elbo_args_vp_to_star_state(vp)
+    gal_shape = [vp[ids.e_dev]  ,
+                 vp[ids.e_axis] ,
+                 vp[ids.e_angle],
+                 vp[ids.e_scale]]
+    return [shared_params; unconstrain_gal_shape(gal_shape)]
+end
+
+
+function variational_params_to_fluxes(i::Int, vs::Array{Array{Float64,1}})
```
jeff-regier, line 417: Can you initialize this from the `CatalogEntry` variables rather than the variational parameters? We really should move all this `vp`/`vs` stuff to within the `DeterministicVI` module.

jeff-regier (changes requested):
Looks good. The Model is still somewhat mixed up with the VI stuff though. Can you avoid using `ea` in the tests and `vp` and `vs` in the `Model` code?

src/model/fsm_util.jl
andymiller: Good call -- done.

src/model/fsm_util.jl
andymiller: Not needed - removed.

src/model/log_prob.jl
andymiller: I forgot about this, I will incorporate this function and add the corresponding test.

andymiller:
I used `VariationalParams` because the type was defined in `Model.jl`.  For the `log_prob` stuff, I think I just need to remove the dependency of `load_bvn_mixtures` on `vp`, and then I can directly operate on the fluxes (and jettison the `variational_params_to_fluxes` method I've been using).

jeff-regier:
I just added issue about moving `VariationalParams` to `DeterministicVI`. Sorry for not catching that sooner.

andymiller:
The `load_bvn_mixtures` function takes in vp, but I think it only needs the location + galaxy params.  Should I create a new `Shape` type that keeps track of [u1, u2, e_dev, e_axis, e_angle, e_scale] for each source?  Then a variational param could either just contain one of these `Shape` types, or we can convert between them when needed.

andymiller:
@jeff-regier, I think removing all of the references to variational parameters will take a little more work.  Can we merge this as is, and then I'll work on scrubbing that dependency?

jeff-regier:
Sure, sounds good.

Issue #251, opened 2016-07-27
performance regression testing


jeff-regier:
We need reliable tests of Celeste's run times, and how runtimes have evolved since ICML. In the past `benchmark_elbo.jl` worked well until it didn't, with the addition of the Hessian code, around late February 2016. 

Back then we expected `benchmark_elbo.jl` to take ~25 seconds, but after the Hessian commit, it took around ~125 seconds. It would be good to understand what's happening. In `benchmark_elbo.jl`, though, the number of sources `S` is 100. Going forward `S=1` is more important.

Tests should also measure ELBO (the objective function) performance with/without gradients, and with/without Hessians.

[julia/test/perf](https://github.com/JuliaLang/julia/tree/master/test/perf) may be a good place to look for examples of how performance testing is done.

PR #328, opened 2016-10-03
Benchmark elbo

### LINKED ISSUE #125: Understand benchmark_elbo.jl run time

jeff-regier:
This PR restores the old `benchmark_elbo.jl` script, that tests `DeterministicVI.elbo` on 100 light sources. It used to take 24 seconds. When I first restored it, it took 32 seconds. Then I found a couple of commands in `SensitiveFloats` that were taking too long, and changed them. Now it's back to 24 seconds. So I think this  PR effectively closes #125. 

I'd leave #251 open because I think we still need more benchmarks. Probably a benchmark with a setup something like `test_infer.jl` would do it.

### DIFF
--- a/src/SensitiveFloats.jl
+++ b/src/SensitiveFloats.jl
@@ -161,22 +161,15 @@ function combine_sfs!{ParamType <: ParamSet,
         sf_result::SensitiveFloat{ParamType, T1},
         v::T1, g_d::Vector{T2}, g_h::Matrix{T3};
         calculate_hessian::Bool=true)
-
-    # TODO: time consuming **************
-
-    # TODO: this line is allocating a lot of memory and I don't know why.
-    # Commenting this line out attributes the same allocation to the next line.
-    # Is memory being allocated lazily or misattributed?
-    @assert g_h[1, 2] == g_h[2, 1]
-
     # You have to do this in the right order to not overwrite needed terms.
     if calculate_hessian
         combine_sfs_hessian!(sf1, sf2, sf_result, g_d, g_h)
     end
 
-    for ind in eachindex(sf_result.d)
-        sf_result.d[ind] = g_d[1] * sf1.d[ind] + g_d[2] * sf2.d[ind]
-    end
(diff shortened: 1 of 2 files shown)


src/SensitiveFloats.jl
```diff
@@ -161,22 +161,15 @@ function combine_sfs!{ParamType <: ParamSet,
         sf_result::SensitiveFloat{ParamType, T1},
         v::T1, g_d::Vector{T2}, g_h::Matrix{T3};
         calculate_hessian::Bool=true)
-
-    # TODO: time consuming **************
-
-    # TODO: this line is allocating a lot of memory and I don't know why.
-    # Commenting this line out attributes the same allocation to the next line.
-    # Is memory being allocated lazily or misattributed?
-    @assert g_h[1, 2] == g_h[2, 1]
```
jeff-regier, line 170: I deleted the assertion since you don't usually want to test floating point numbers for exact equality.

rgiordan: In case you wonder, Hessians can be constructed enforcing exact symmetry, by populating mirror elements with exactly the same float.  Everything I've written does it this way.  That's why this is ok (notice it hasn't failed yet).

Incidentally, the second TODO can be deleted, as it was due to a bug in the profiler.

jeff-regier: OK I'll put the assertion back. Did we ever consider the Hermitian matrix type? Does it not support some operations that we need or something?

rgiordan:
That's a good question.  I seem to recall there being something since we
were forced to sort the eigenvalues in the trust region code.  (By which: I
mean I recall trying to use a symmetric matrix type and not being able to
and having to sort the eigenvalues instead, but I don't remember what the
problem was.)  But I don't think I tried very hard, and it's probably worth
revisiting.

PR #331, opened 2016-10-04
remove Exceptions module; define error type where it's used

gostevehoward:
Following up on [comment here](https://github.com/jeff-regier/Celeste.jl/pull/315).

The logic was that, conceptually, this exception has nothing to do with DeterministicVI, and could be used elsewhere in the code. But practically, if it's only used in one place right now, with no immediate plans to use it elsewhere, I'm happy to define it where it's used and it can easily be extracted to a standalone module later if it gets reused.

(The counterargument is that putting it in a shared place to start with encourages reuse by making it easier for developers to find. But it's speculative, and I'm not concerned.)


jeff-regier:
Thanks Steve!

PR #339, opened 2016-10-10
lower memory usage

jeff-regier:


andreasnoack:
Do you create many `BvnComponent`s? It seems like an obvious candidate for `FixedSizeArrays`. That would completely avoid GC interactions.

`sqrt(maximum([ the_cov[1, 1], the_cov[2, 2] ]))` -> `sqrt(max( the_cov[1, 1], the_cov[2, 2]))` will also save you an allocation.

jeff-regier:
Good idea. So far we have other lines that get run more often, but this would be worth doing once the others are fixed.

src/bivariate_normals.jl
```diff
@@ -126,14 +126,18 @@ immutable BvnComponent{NumType <: Number}
...
+      precision[1, 2] = -the_cov[1, 2] / the_det
+      precision[2, 1] = -the_cov[2, 1] / the_det
+
       if calculate_siginv_deriv
         # Derivatives of Sigma^{-1} with respect to sigma.  These are the second
         # derivatives of log|Sigma| with respect to sigma.
         # dsiginv_dsig[a, b] is the derivative of sig^{-1}[a] / d sig[b]
         dsiginv_dsig = Array(ThisNumType, 3, 3)
 
-        precision = the_cov^-1
```
jeff-regier, line 135: @andreasnoack -- Any idea why inverting a small matrix allocates so much memory?

andreasnoack: Yes. Many temporaries are created in the process. First, because you are calling the general matrix power function, what is calculated is

``` jl
A^-1 = inv(A^1)
```

and `A^1 = copy(A)` so that is one copy. Then

``` jl
inv(A) = inv(lufact(A)) = inv(lufact!(copy(A))) = inv!(copy(lufact!(copy(A))))
```

so three copies. We could avoid one of the copies by using `inv!` instead of `inv` since we already know that we have copied `A`. The LU factorization allocates a vector for the permutations and finally the intermediate `LU` type is heap allocated because it contains heap references. Avoiding the last part is work in progress but it's not clear when that work is done.

This is very inefficient for a 2x2 matrix so the way to go here is to use `FixedSizeArrays` or something similar. I'll push a branch soon where I've changed some of the bivariate normal code to use `FixedSizeArrays`. I tried it on `benchmark_inter` and it reduces the allocations a bit but it didn't really speed things up but I guess fixed the change can also be used for the derivatives so there might be more to gain here.

jeff-regier: Sounds like a useful PR you're working on, regardless of what `benchmark_infer.jl` reports, because we'll deal with the current bottlenecks soon and then this will be a major use of memory. As @kpamnany suggests, maybe `StaticArrays` would be better if it alone is actively maintained?

andreasnoack: I've changes the branch to use `StaticArrays` instead of `FixedSizeArrays` and tests are now passing (with a small patch to `StaticArrays`). The timing output from `benchmark_inter` goes from

```
24.936787 seconds (20.48 M allocations: 1.339 GB, 3.82% gc time)
10.575892 seconds (8.19 M allocations: 866.467 MB, 7.65% gc time)
```

on master to

```
23.577065 seconds (19.39 M allocations: 1.255 GB, 4.11% gc time)
10.268357 seconds (6.31 M allocations: 752.464 MB, 7.73% gc time)
```

on `anj/fixedsize`. So the effect is modest at this point.

jeff-regier: That's a really good reduction in memory allocations. Thanks for working it. Would you submit it as a PR? @kpamnany -- Should we include this in the IPDPS runs? (I don't see any risky changes and I'm verifying it on stripe 82 now.)

kpamnany: Sure. You pick the code freeze point. :)

andreasnoack: My branch would probably require https://github.com/JuliaArrays/StaticArrays.jl/pull/51 depending on which functions in Celeste that are exercised. I'll try to finish up the PR to `StaticArrays` now.

jeff-regier:
@kpamnany I wouldn't delay your scaling runs, but if Andreas gets a PR merged beforehand let's use it. Andreas, looks like some of the tests for your branch are failing on Travis

andreasnoack:
It fails because of the missing patch to `StaticArrays`. I've updated the PR there so it should be ready to merge.

andreasnoack:
The patch has been merged into `StaticArrays` so now the test pass https://travis-ci.org/jeff-regier/Celeste.jl/builds/167525154. Temporarily, Travis is checking out the master version of `StaticArrays`. A new version will probably make it into METADATA within 24 hours.

kpamnany:
The scaling runs are underway, so we won't use the PR-to-be unless you guys think it will help a lot, or if I have to rerun.

PR #340, opened 2016-10-10
Active pixels

jeff-regier:
Removes `trim_tile...` and moves finding active pixels into pre-processing. Way less memory allocation.

### DIFF
--- a/src/DeterministicVI.jl
+++ b/src/DeterministicVI.jl
@@ -77,17 +77,6 @@ function ElboArgs{NumType <: Number}(
         end
     end
 
-    active_pixels = ActivePixel[]
-    for i in 1:N
-        tiles = images[i].tiles
-        for j in 1:length(tiles)
-            W, H = size(tiles[j].pixels)
-            for w in 1:W, h in 1:H
-                push!(active_pixels, ActivePixel(i, j, h, w))
-            end
-        end
-    end
-
     ElboArgs(S, N, psf_K, images, vp, tile_source_map, patches,
              active_sources, num_allowed_sd, ActivePixel[])
 end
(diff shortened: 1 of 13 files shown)


test/runtests.jl
```diff
@@ -28,15 +28,15 @@ run(`make RUN=4263 CAMCOL=5 FIELD=119`)
 cd(wd)
 
 # Check whether to run time-consuming derivatives tests.
-test_derivatives_flag = "--test_derivatives"
-test_detailed_derivatives = test_derivatives_flag in ARGS
-test_files = setdiff(ARGS, [ test_derivatives_flag ])
+long_running_flag = "--long-running"
```
jeff-regier, line 31: @rgiordan -- FYI, I renamed `--test_derivatives` flag to `--long-running`. I thought it `--test_derivatives` was confusing because you can test some derivatives without that flag, and I thought `--long-running` could be used for other long running tests too.

src/DeterministicVI.jl
```diff
@@ -66,19 +68,17 @@ function ElboArgs{NumType <: Number}(
...
-        for tile in tiled_image.tiles
-            for ep in tile.epsilon_mat
-                if ep <= 0.0
-                    throw(InvalidInputError(
-                        "You must set all values of epsilon_mat > 0 for all images included in ElboArgs"
-                    ))
-                end
-            end
+
+    for img in images, tile in img.tiles, ep in tile.epsilon_mat
```
jeff-regier, line 72: @gostevehoward -- I changed the epsilon check to loop over the values. It turns out that `all(isnan...)` allocates a lot of memory, I guess it's forming any array of booleans for each pixel before running `all`.

gostevehoward:
Good point, but now I wonder if the call I added to assert_all_finite() in
process_active_pixels() is a large source of wasted allocations? i.e.,
because of elbo_args.jl line 167
<(https://github.com/jeff-regier/Celeste.jl/blob/master/src/deterministic_vi/elbo_args.jl#L167>,
where we call isfinite() on the hessian? (I didn't know allocations were a
big concern.)

jeff-regier:
I think it's fine. There are a lot more epsilons than elements in the Hessian.

PR #342, opened 2016-10-11
Preallocate elbo vars

jeff-regier:


test/test_images.jl
```diff
@@ -216,7 +221,7 @@ end
 
 
 test_blob()
-test_stamp_get_object_psf()
-test_get_tiled_image_source()
-test_local_source_candidate()
-test_set_patch_size()
+#test_stamp_get_object_psf()
```
rgiordan, line 219: You're disabling a lot of tests here -- would you like to make an issue to ensure that they get enabled?  This has bitten us in the past.

jeff-regier: OK, good idea

PR #350, opened 2016-10-14
Calculate one fsm at a time, possibly outside ElboArgs

rgiordan:
This allows me to save some copies when populating a matrix of fsm sensitive floats.

### DIFF
--- a/src/SensitiveFloats.jl
+++ b/src/SensitiveFloats.jl
@@ -84,6 +84,18 @@ function zero_sensitive_float{ParamType <: ParamSet}(
 end
 
 
+function zero_sensitive_float_array{ParamType <: ParamSet}(
+    ::Type{ParamType}, NumType::DataType, local_S::Int, d::Integer...)
+
+    sf_array = Array(SensitiveFloat{ParamType, NumType}, d)
+    for ind in 1:length(sf_array)
+        sf_array[ind] = zero_sensitive_float(ParamType, NumType, local_S)
+    end
+
+    sf_array
+end
+
+
 function clear!{ParamType <: ParamSet, NumType <: Number}(
             sp::SensitiveFloat{ParamType, NumType})
     clear!(sp, true)
--- a/src/deterministic_vi/elbo_objective.jl
+++ b/src/deterministic_vi/elbo_objective.jl
@@ -117,33 +117,37 @@ and then updated in place.
 Args:
     - elbo_vars: Elbo intermediate values, with updated fs1m_vec and fs0m_vec.
     - ea: Model parameters
+    - E_G_s: The expectation  and variance of the brightnesses of this source
+          at this pixel, updated in place.
+    - fs0m, fs1m: The star and galaxy shape parameters for this source at
+          this pixel.
     - sbs: Source brightnesses
     - s: The source, in 1:ea.S
     - b: The band
 
 Returns:
-    Updates elbo_vars.E_G_s and elbo_vars.var_G_s in place with the brightness
-    for this sourve at this pixel.
+    Updates E_G_s and var_G_s in place with the brightness
+    for this source at this pixel.
 """
 function accumulate_source_brightness!{NumType <: Number}(
                     elbo_vars::ElboIntermediateVariables{NumType},
                     ea::ElboArgs{NumType},
-                    sbs::Vector{SourceBrightness{NumType}},
-                    s::Int, b::Int)
-    # E[G] and E{G ^ 2} for a single source
-    E_G_s = elbo_vars.E_G_s
+                    E_G_s::SensitiveFloat{CanonicalParams, NumType},
+                    var_G_s::SensitiveFloat{CanonicalParams, NumType},
+                    fs0m::SensitiveFloat{StarPosParams, NumType},
+                    fs1m::SensitiveFloat{GalaxyPosParams, NumType},
+                    sb::SourceBrightness{NumType},
+                    b::Int, s::Int,
+                    active_source::Bool)
+
     E_G2_s = elbo_vars.E_G2_s
 
     clear_hessian = elbo_vars.calculate_hessian && elbo_vars.calculate_derivs
     clear!(E_G_s, clear_hessian)
     clear!(E_G2_s, clear_hessian)
 
-    sb = sbs[s]
-
-    active_source = (s in ea.active_sources)
-
     for i in 1:Ia # Stars and galaxies
-        fsm_i = (i == 1) ? elbo_vars.fs0m_vec[s] : elbo_vars.fs1m_vec[s]
+        fsm_i = (i == 1) ? fs0m : fs1m
         a_i = ea.vp[s][ids.a[i, 1]]
 
         lf = sb.E_l_a[b, i].v[1] * fsm_i.v[1]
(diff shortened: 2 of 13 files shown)


src/deterministic_vi/elbo_objective.jl
```diff
@@ -127,12 +63,54 @@ function populate_fsm_vecs!{NumType <: Number}(
...
+"""
+function populate_fsm!{NumType <: Number}(
+                    elbo_vars::ElboIntermediateVariables{NumType},
+                    ea::ElboArgs{NumType},
+                    fs0m::SensitiveFloat{StarPosParams, NumType},
+                    fs1m::SensitiveFloat{GalaxyPosParams, NumType},
+                    s::Int,
+                    b::Int,
+                    x::Vector{Float64},
+                    active_source::Bool,
```
jeff-regier: Maybe call this `is_active_source`, to indicate that it's a boolean, and for consistency.

jeff-regier (approved):
lgtm

rgiordan:
@jeff-regier , PTAL?  I added some more changes to this PR as long as I was adding back in the zero_sensitive_float_array.

src/deterministic_vi/elbo_objective.jl
```diff
@@ -127,45 +127,91 @@ function populate_fsm_vecs!{NumType <: Number}(
...
+    @assert 1 <= b <= B
+    Model.populate_fsm!(elbo_vars.bvn_derivs,
+                        fs0m, fs1m,
+                        elbo_vars.calculate_derivs,
+                        elbo_vars.calculate_hessian,
+                        s, x, is_active_source,
+                        ea.num_allowed_sd,
+                        ea.patches[s, b].wcs_jacobian,
+                        gal_mcs, star_mcs)
+end
```
jeff-regier: Would you just call `Model.populate_fsm!` directly, without this wrapper function? (I suspect this was left over from previous PR, though it shows up as new code, but would mind removing it as part of this one? It adds a lot of lines to the code base without doing anything.)

src/deterministic_vi/elbo_objective.jl
```diff
@@ -127,45 +127,91 @@ function populate_fsm_vecs!{NumType <: Number}(
...
+          this pixel.
+    - sb: Source brightnesse
     - s: The source, in 1:ea.S
     - b: The band
 
 Returns:
-    Updates elbo_vars.E_G_s and elbo_vars.var_G_s in place with the brightness
-    for this sourve at this pixel.
+    Updates E_G_s and var_G_s in place with the brightness
+    for this source at this pixel.
```
jeff-regier, line 152: I don't typically think of in place updates as being return values. Have you seen comments in other Julia packages, where they refer in place updates as returned values? Anyway, feel free to ignore this...I prefer more minimalist comments but that's a personal preference and not really about this PR.

src/SensitiveFloats.jl
```diff
@@ -84,6 +84,18 @@ function zero_sensitive_float{ParamType <: ParamSet}(
...
+function zero_sensitive_float_array{ParamType <: ParamSet}(
+    ::Type{ParamType}, NumType::DataType, local_S::Int, d::Integer...)
+
+    sf_array = Array(SensitiveFloat{ParamType, NumType}, d)
+    for ind in 1:length(sf_array)
+        sf_array[ind] = zero_sensitive_float(ParamType, NumType, local_S)
+    end
+
+    sf_array
+end
```
jeff-regier, line 96: I shouldn't have deleted it and you're planning to use it shortly, so let's get this included. In general, I think we should aim to keep unreachable code (code that doesn't get called from `bin`) out of the code base, and commit it only once it is reachable from `bin`. I mean, there's no need to be too strict about this, but unreachable code is pretty confusing for anyone reading the code who doesn't know it's history, in my experience.

src/deterministic_vi/elbo_objective.jl
```diff
@@ -315,7 +361,26 @@ function accumulate_source_brightness!{NumType <: Number}(
...
+        elbo_vars,
+        ea,
+        elbo_vars.E_G_s,
+        elbo_vars.var_G_s,
+        elbo_vars.fs0m_vec[s],
+        elbo_vars.fs1m_vec[s],
+        sbs[s],
+        b, s,
+        s in ea.active_sources)
 end
```
jeff-regier, line 342: Same here, I'd be great to not have this wrapper.

src/model/fsm_util.jl
```diff
@@ -184,22 +184,65 @@ end
 
 
 """
-Populate fs0m_vec and fs1m_vec for all sources for a given pixel.
+Populate fs0m and fs1m for source s in the a given pixel.
 
-Args:
-    - model_vars: Model intermediate values.
```
jeff-regier, line 190: `tmp_model_vars` might be an event better name for `model_vars`, or `tmp_vars`, just because `model_vars` isn't so clearly different from model parameters, but up to you of course.

test/test_log_prob.jl
```diff
@@ -122,5 +110,5 @@ end
 test_sigmoid_logit()
 test_gal_shape_constrain()
 test_color_flux_transform()
-#test_that_star_truth_is_most_likely()
+#test_that_star_truth_is_most_likely_log_prob()
```
jeff-regier, line 113: Ah, thanks for renaming that...I'd seen that warning a lot

jeff-regier (approved):
Lgtm.  Will you run `bin/benchmark_infer.jl` before committing it to make sure runtime and memory allocations aren't changed?

src/deterministic_vi/elbo_objective.jl
rgiordan: No, I haven't seen that elsewhere.  I'd be happy to conform to some style guide if we had one.  I do think it's pretty useful for functions to document what variables they change, and I know we disagree about this.  I think the way we've been doing it where I just document functions I care about has been working ok, so if you don't object, I'm just going to keep doing it.

src/deterministic_vi/elbo_objective.jl
rgiordan: Sure.  At some point we will have to do a substantial reorganization, since the idea of passing variables around in a single argument type is gradually eroding.  But this is easy to pull out.

src/deterministic_vi/elbo_objective.jl
rgiordan: Unlike the other, this wrapper does the service of referring to the appropriate pre-allocated memory in elbo_vars, and so seems worth having around.

src/model/fsm_util.jl
rgiordan: Maybe it would be easier to fit these renaming requests into less substantive PRs.

src/SensitiveFloats.jl
rgiordan: This makes sense, and I'll try to follow that guideline in the future.  But the PSF is a substantive change, and though I'm doing most of the work in a separate repo, in the interest of smaller PRs I think it makes sense to do some things like this first.

src/deterministic_vi/elbo_objective.jl
jeff-regier: Yeah, I noticed that erosion. Not a big deal though, extra arguments is fine while the code is in flux.

src/deterministic_vi/elbo_objective.jl
jeff-regier: That's fine. I may try to make my case to you in person someday :-)

src/deterministic_vi/elbo_objective.jl
jeff-regier: Doesn't seem worth it to me, unless the wrapper is called from a lot of different places. Extra indirection is confusing, especially for someone who didn't write the code. But as you wish.

rgiordan:
I ran `bin/benchmark_infer.jl` and there were no major changes.

Before I merge, @jeff-regier , if I were to do an IPDPS code freeze I would probably not include this PR, especially since we don't really understand what's breaking now and this will add a little noise.  It would be convenient for me to have this in so I don't have to keep doing merges.  However, IPDPS is more important and I would understand if you would prefer that I wait for a week before merging.  What do you think?

jeff-regier:
Lets merge it, we'll make any further changes to the `src_div_par` branch directly

PR #351, opened 2016-10-14
Use StaticArrays for small matrices and vectors to avoid heap allocations.

andreasnoack:
The time `@time` in `benchmark_infer` are

``` jl
24.936787 seconds (20.48 M allocations: 1.339 GB, 3.82% gc time)
10.575892 seconds (8.19 M allocations: 866.467 MB, 7.65% gc time)
```

on master and

``` jl
23.577065 seconds (19.39 M allocations: 1.255 GB, 4.11% gc time)
10.268357 seconds (6.31 M allocations: 752.464 MB, 7.73% gc time)
```

with this PR.

It might be possible to push this further. E.g. for the derivatives and I'm also wondering if the PSF parameters could be stored in an immutable and thereby become stack allocated. However, those changes would require more restructuring. In this PR I've just made the changes that didn't require any restructuring.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ before_install:
 script:
   - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
   - julia -e 'Pkg.clone(pwd()); Pkg.build("Celeste")'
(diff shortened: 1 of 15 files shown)


jeff-regier:
Looks great, thanks for reducing the memory allocations.

PR #361, opened 2016-10-20
Logprob refactor

### LINKED ISSUE #327: Move `VariationalParams` from `Model` to `DeterministicVI`
Also, `FreeVariationalParams`.

andymiller:
Removes reference to `VariationalParams` and `FreeVariationalParams` from the `Model` module, and places them in the `Transform` module.  

Creates `LatentStateIndexes` in `model/param_set.jl`, which indexes into a flat array that holds the state of star and galaxy parameters.  I add a conversion function for `CatalogEntry` objects to the `LatentStateParams`-ordered arrays.  

This closes #327

### DIFF
--- a/src/model/fsm_util.jl
+++ b/src/model/fsm_util.jl
@@ -8,7 +8,7 @@ Args:
  - patches (formerly from ElboArgs)
  - vp: (formerly from ElboArgs)
  - active_sources (formerly from ElboArgs)
- - psf: A vector of PSF components
+ - psf_K: Number of psf components (psf from patches object)
  - b: The current band
  - calculate_derivs: Whether to calculate derivatives for active sources.
  - calculate_hessian
--- a/src/model/fsm_util.jl
+++ b/src/model/fsm_util.jl
@@ -27,7 +27,7 @@ Returns:
 function load_bvn_mixtures{NumType <: Number}(
                     S::Int64,
                     patches::Matrix{SkyPatch},
-                    vp::VariationalParams{NumType},
+                    source_params::Vector{Vector{NumType}},
                     active_sources::Vector{Int},
                     psf_K::Int64,
                     b::Int;
@@ -40,9 +40,9 @@ function load_bvn_mixtures{NumType <: Number}(
     # active_sources.
     for s in 1:S
         psf = patches[s, b].psf
-        vs = vp[s]
+        sp  = source_params[s]
 
-        world_loc = vs[[ids.u[1], ids.u[2]]]
+        world_loc = sp[lidx.u]
         m_pos = Model.linear_world_to_pix(patches[s, b].wcs_jacobian,
                                           patches[s, b].center,
                                           patches[s, b].pixel_center, world_loc)
@@ -59,58 +59,15 @@ function load_bvn_mixtures{NumType <: Number}(
         # Convolve the galaxy representations with the PSF.
         for i = 1:2 # i indexes dev vs exp galaxy types.
             e_dev_dir = (i == 1) ? 1. : -1.
-            e_dev_i = (i == 1) ? vs[ids.e_dev] : 1. - vs[ids.e_dev]
+            e_dev_i = (i == 1) ? sp[lidx.e_dev] : 1. - sp[lidx.e_dev]
 
             # Galaxies of type 1 have 8 components, and type 2 have 6 components.
             for j in 1:[8,6][i]
                 for k = 1:psf_K
                     gal_mcs[k, j, i, s] = GalaxyCacheComponent(
                         e_dev_dir, e_dev_i, galaxy_prototypes[i][j], psf[k],
-                        m_pos, vs[ids.e_axis], vs[ids.e_angle], vs[ids.e_scale],
-                        calculate_derivs && (s in active_sources),
-                        calculate_hessian)
-                end
-            end
-        end
-    end
-
-    star_mcs, gal_mcs
-
-
-    star_mcs = Array(BvnComponent{NumType}, psf_K, S)
-    gal_mcs = Array(GalaxyCacheComponent{NumType}, psf_K, 8, 2, S)
-
-    # TODO: do not keep any derviative information if the sources are not in
-    # active_sources.
-    for s in 1:S
-        psf = patches[s, b].psf
-        vs = vp[s]
-
-        world_loc = vs[[ids.u[1], ids.u[2]]]
-        m_pos = Model.linear_world_to_pix(patches[s, b].wcs_jacobian,
-                                          patches[s, b].center,
-                                          patches[s, b].pixel_center, world_loc)
-
-        # Convolve the star locations with the PSF.
-        for k in 1:psf_K
-            pc = psf[k]
-            mean_s = [pc.xiBar[1] + m_pos[1], pc.xiBar[2] + m_pos[2]]
-            star_mcs[k, s] =
-              BvnComponent{NumType}(
-                mean_s, pc.tauBar, pc.alphaBar, calculate_siginv_deriv=false)
-        end
-
-        # Convolve the galaxy representations with the PSF.
-        for i = 1:2 # i indexes dev vs exp galaxy types.
-            e_dev_dir = (i == 1) ? 1. : -1.
-            e_dev_i = (i == 1) ? vs[ids.e_dev] : 1. - vs[ids.e_dev]
-
-            # Galaxies of type 1 have 8 components, and type 2 have 6 components.
-            for j in 1:[8,6][i]
-                for k = 1:psf_K
-                    gal_mcs[k, j, i, s] = GalaxyCacheComponent(
-                        e_dev_dir, e_dev_i, galaxy_prototypes[i][j], psf[k],
-                        m_pos, vs[ids.e_axis], vs[ids.e_angle], vs[ids.e_scale],
+                        m_pos,
+                        sp[lidx.e_axis], sp[lidx.e_angle], sp[lidx.e_scale],
                         calculate_derivs && (s in active_sources),
                         calculate_hessian)
                 end
--- a/src/model/log_prob.jl
+++ b/src/model/log_prob.jl
@@ -3,7 +3,7 @@
 using Distributions
 import ..SensitiveFloats: SensitiveFloat, zero_sensitive_float, clear!
 
-EPS = 1e-6
+const eps_prob_a = 1e-6
 
 ###################################
 # log likelihood function makers  #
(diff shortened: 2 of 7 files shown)


src/model/log_prob.jl
```diff
@@ -206,13 +200,15 @@ function state_log_likelihood(is_star::Bool,
...
         # compute the background rate for this pixel
-        background_rate = tile.epsilon_mat[pixel.h, pixel.w]
+        background_rate    = tile.epsilon_mat[pixel.h, pixel.w]
         background_sources = tile_sources[tile_sources.!=1]
         for s in background_sources
-            flux_s = variational_params_to_fluxes(s, vp)
-            rate_s = is_star ? model_vars.fs0m_vec[s].v : model_vars.fs1m_vec[s].v
-            rate_s *= flux_s[pixel_band]
-            background_rate += rate_s
+            println("background s: ", s)
```
jeff-regier: We need to avoid print statements, not just for clutter...they actually cause Julia to crash when it has multiple threads. I wonder when they'll have a thread-safe println statement.

src/model/log_prob.jl
```diff
@@ -206,13 +200,15 @@ function state_log_likelihood(is_star::Bool,
...
+            #TODO: compute background rate conditional on source_params
+            #flux_s = variational_params_to_fluxes(s, vp)
+            #rate_s = is_star ? model_vars.fs0m_vec[s].v : model_vars.fs1m_vec[s].v
+            #rate_s *= flux_s[pixel_band]
+            #background_rate += rate_s
         end
```
jeff-regier, line 212: I'd rather exclude commented out code from the commit.

src/model/log_prob.jl
```diff
@@ -395,44 +391,40 @@ function init_galaxy_state(entry::CatalogEntry)
...
-function variational_params_to_fluxes(i::Int, vs::Array{Array{Float64,1}})
-    ret = Array(Float64, 5)
-    ret[3] = exp(vs[ids.r1[i]] + 0.5 * vs[ids.r2[i]])
-    ret[4] = ret[3] * exp(vs[ids.c1[3, i]])
-    ret[5] = ret[4] * exp(vs[ids.c1[4, i]])
-    ret[2] = ret[3] / exp(vs[ids.c1[2, i]])
-    ret[1] = ret[2] / exp(vs[ids.c1[1, i]])
-    ret
+function extract_star_state(ls::Array{Float64, 1})
+    return [[ls[lidx.r[1]]]; ls[lidx.c[:, 1]]; ls[lidx.u]]
```
jeff-regier, line 418: Can you return a tuple here rather than a vector? Vectors allocate memory on the heap each time....very slow potentially. Tuples are on the stack.

src/model/log_prob.jl
```diff
@@ -395,44 +391,40 @@ function init_galaxy_state(entry::CatalogEntry)
...
-    ret[5] = ret[4] * exp(vs[ids.c1[4, i]])
-    ret[2] = ret[3] / exp(vs[ids.c1[2, i]])
-    ret[1] = ret[2] / exp(vs[ids.c1[1, i]])
-    ret
+function extract_galaxy_state(ls::Array{Float64, 1})
+    gal_shape = [ls[lidx.e_dev]  ,
+                 ls[lidx.e_axis] ,
+                 ls[lidx.e_angle],
+                 ls[lidx.e_scale]]
+    return [[ls[lidx.r[2]]]; ls[lidx.c[:, 2]]; ls[lidx.u];
```
jeff-regier, line 427: Same for these two lines...small vectors seem to create a huge number of allocations. Also, please try running `./benchmark_infer.jl` with the changes to confirm that runtime and memory allocations haven't increased.

src/model/param_set.jl
```diff
@@ -93,6 +93,30 @@ getids(::Type{CanonicalParams}) = ids
...
+            reshape((7+Ia):(7+Ia+(B-1)*Ia-1), (B-1, Ia)),  # c
+            reshape((7+Ia+(B-1)*Ia):(7+2Ia+(B-1)*Ia-1), (Ia, 1)),  # a
+            reshape((7+2Ia+(B-1)*Ia):(7+2Ia+(B-1)*Ia+D*Ia-1), (D, Ia))) # k
+end
+
+const lidx = LatentStateIndexes()
+getlidx(::Type{LatentStateIndexes}) = lidx
+length(::Type{LatentStateIndexes}) = 22 #6 + 3*Ia + 2*(B-1)*Ia + D*Ia
+
+
```
jeff-regier, line 119: Looks good!

test/test_log_prob.jl
```diff
@@ -110,5 +116,5 @@ end
 test_sigmoid_logit()
 test_gal_shape_constrain()
 test_color_flux_transform()
-#test_that_star_truth_is_most_likely_log_prob()
+test_that_star_truth_is_most_likely_log_prob()
 #test_that_gal_truth_is_most_likely()
```
jeff-regier: Should we have `test_that_gal_truth_is_most_likely_log_prob()` test too?

src/model/fsm_util.jl
```diff
@@ -27,7 +27,7 @@ Returns:
 function load_bvn_mixtures{NumType <: Number}(
```
jeff-regier, line 27: Some of the `Args` listed in the comment are out of date. What about documenting some of the recurring arguments, e.g. `S` and `b` in a comment for the whole `Model` module, rather than in each function that uses them?

src/model/log_prob.jl
```diff
@@ -3,8 +3,6 @@
 using Distributions
 import ..SensitiveFloats: SensitiveFloat, zero_sensitive_float, clear!
 
-#include("../bivariate_normals.jl")
-#include("log_prob_util.jl")
 EPS = 1e-6
```
jeff-regier: EPS should be maked `const`...this can make a big different for runtime.

I haven't seen the all caps convention used in Julia for constants.

Also, `eps` could be confused with `espilon_mat` in the Celeste model. Is there another way to do this?

jeff-regier (changes requested):
Looks good, the main thing is verifying the `./benchmark_infer.jl` doesn't show that allocations and runtime changing (on the second executing--the first executing is to compile the code.)

src/model/log_prob.jl
andymiller: updated - made `EPS` into `const eps_log_prob`.

src/model/log_prob.jl
andymiller: ahhh my bad -- removed!

src/model/log_prob.jl
andymiller: Updated!  Mirrors the pixel rate calculation for all sources.

src/model/log_prob.jl
andymiller: How do you string out a few vectors into a single tuple?  The goal of this function is to return a flat vector of parameters (or tuple) that can be used in a generic MCMC sampler routine.

jeff-regier: The call it splatting I think.

```
julia> x = randn(5)
5-element Array{Float64,1}:
  0.677473 
 -1.24358  
  0.85705  
  0.0845632
 -0.944667 

julia> y = (1.1, 2.2, x...)
(1.1,2.2,0.6774731377985451,-1.243582815614421,0.8570502787317705,0.08456323283057172,-0.9446668464656754)
```

test/test_log_prob.jl
jeff-regier: or `test_that_gal_truth_maximizes_log_prob`, that name should be unique. Ryan added `_log_prob` just to avoid naming conflicts with the elbo tests.

src/model/log_prob.jl
andymiller: Thanks!  

One issue with substituting tuples here is that MCMC and MAP estimation will be doing vector/matrix computations with the parameter vector.  This conversion won't be called at each iteration --- just to start, so it might be premature to rewrite code to optimize this before I benchmark mcmc routines.

src/model/fsm_util.jl
andymiller: Ok --- i updated the comment above `load_bvn_mixtures` to account for `psf_K`.  I pattern matched on this refactor -- I'm not super sure what all of the inputs from `ElboArgs` were.

andymiller:
This commit shouldn't change results from `benchmark_infer.jl`, but below is a comparison of this branch to master on my machine

Run on this branch

```
ti profile:
prep: 1.9193 (1919304278)
fork: 5.184e-06 (5.184e-06 - 5.184e-06)
user: 26.4687 (26.4687 - 26.4687)
join: 1.766e-06 (1.766e-06 - 1.766e-06)
```

Run on master

```
ti profile:
prep: 1.9485 (1948498076)
fork: 5.344e-06 (5.344e-06 - 5.344e-06)
user: 27.1303 (27.1303 - 27.1303)
join: 1.916e-06 (1.916e-06 - 1.916e-06)
```

test/test_log_prob.jl
andymiller: Great call --- this one slipped by.  I'll write this up shortly!

jeff-regier:
It's too bad `benchmark_infer.jl` prints out so much output. The line we're interested looks like:

```
 22.654070 seconds (6.34 M allocations: 740.662 MB, 7.17% gc time)
```

It's the second printed line that has this form. (The first one also compiles the code.)

The 22.65 seconds is specific to my old computer. The 70.66 MB in allocations should match up with what you see for master.

test/test_log_prob.jl
andymiller: Added!

andymiller:
Reran `benchmark_infer.jl`

logprob-refactor branch:

```
ti profile:
prep: 1.97917 (1979171610)
fork: 5.022e-06 (5.022e-06 - 5.022e-06)
user: 19.5808 (19.5808 - 19.5808)
join: 1.401e-06 (1.401e-06 - 1.401e-06)
  7.837405 seconds (5.07 M allocations: 667.170 MB, 9.47% gc time)
```

master: 

```
ti profile:
prep: 1.93458 (1934577849)
fork: 4.839e-06 (4.839e-06 - 4.839e-06)
user: 19.9811 (19.9811 - 19.9811)
join: 1.547e-06 (1.547e-06 - 1.547e-06)
  8.059243 seconds (6.34 M allocations: 740.664 MB, 9.39% gc time)
```

Do those look good, @jeff-regier ?

jeff-regier (approved):

jeff-regier:
Looks great! Please merge this whenever you're ready.

jeff-regier:
Also, if you add the statement "This closes #327" to your initial PR description, it'll close the issue automatically once you merge and link the issue and the PR. Kind of a slick move.

PR #363, opened 2016-10-24
added a test for load_active_pixels!

### LINKED ISSUE #355: load_active_pixels! only works with a single active source, unit tests for Infer.jl
`load_active_pixels!` only gets the active pixels for the first active source.  This probably slipped past because it and the other functions in `Infer.jl`, are also missing tests.

jeff-regier:
I added a basic test for `load_active_pixels!`. The thing about testing this function is that it isn't well defined what the function is supposed to do, beyond what the code that implements it says it does. Sometimes really dim pixels are informative because we expect them to be bright (based on initial parameters); sometimes really bright pixels are informative because we expect them to be dim (based on our initial estimates of the parameters). We might want to think more about how to select active pixels. Maybe something about the log probability changes, given the initial estimates of the parameters, if a particular pixel (or set of pixels) is censored. Perhaps subsampling pixels for large galaxies could be a special case of our procedure for selecting active pixels.

This closes #355.

### DIFF
--- a/src/Infer.jl
+++ b/src/Infer.jl
@@ -17,8 +17,10 @@ import WCS
 
 using ..Model
 import ..PSF
-using ..DeterministicVI
 import ..Log
+
+using ..DeterministicVI
+import ..DeterministicVI.tile_predicted_image
 
 
 """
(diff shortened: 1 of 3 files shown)


rgiordan:
I think at least this functionality of the old test should still ber here: that we test
(1) most of an image's brightness was captured by the active pixels and
(2) that min_pixel_radius works as advertised
Maybe you could just find and restore the old test.
I really disagree with simply asserting that you only have one active source.  Even if you are only currently using one source at a time for your work, it wastes of a lot of work I did.  It would not take much effort to fix this.  But we should maybe talk about this in person.

jeff-regier:
@rgiordan -- PTAL. Tests are failing currently only because `data.sdds3.org` is down.

jeff-regier:
`benchmark_infer.jl` gets 10% slower and allocates a bunch more memory :-1:

jeff-regier:
This last commit prevents most of the new memory allocation.

`benchmark_infer.jl` on master:
 19.562068 seconds (5.07 M allocations: 667.168 MB, 6.74% gc time)

`benchmark_infer.jl` with this PR:
 18.438190 seconds (5.07 M allocations: 667.219 MB, 7.01% gc time)

The change in times isn't meaningful because I had some other programs open. The slight increase in memory allocations is acceptable.

src/Infer.jl
```diff
@@ -203,43 +205,77 @@ Arguments:
...
+
+            # TODO; use log_prob.jl in the Model module to get the
+            # get the expected brightness, not variational inference
+            pred_tile_pixels = tile_predicted_image(tile,
+                                                    ea,
+                                                    local_active_sources,
+                                                    include_epsilon=false)
+
+            for h in tile.h_range, w in tile.w_range
+                # The pixel location in the rendered image.
```
rgiordan, line 257: This is the pixel location in the tile, not the image, right?

rgiordan (approved):
Looks good to me, modulo the comment.  Thanks for maintaining the option of multiple sources.

PR #372, opened 2016-10-26
First pass at GalSim benchmarks

gostevehoward:
Seems to be at a reasonable point for a first merge.

The CSV will probably get replaced by a pure Python description since it's not critical anymore (truth values are passed in FITS headers) and is less pretty than Python, and adding things like multi-source test cases would be really ugly in the CSV. Let me know if you'd prefer I do that before merging.

Let me know if you have other questions or want to see docstrings and/or unit tests.


jeff-regier (approved):
This code looks nice, Steve. We may want to move the Julia part to the `src/` directory in the future so that 1) it gets run by unit tests, and 2) so that unit tests can be written based on it. Maybe to a `GalsimBenchmarks` module? But that's for a future PR. Glad to have this committed.

PR #377, opened 2016-10-28
Single Node Multiple Iteration Optimization

agnusmaximus:
- Uses the cyclades partitioning algorithm to partition sources into non-conflicting batches for multi iter optimization
- Target source elbo args share underlying variational parameters, which are updated multiple times in the process_sources function (by multiple threads). This function uses maximize_f to optimize the variational parameters.
- Modify one_node_infer and one_node_infer_multi to also return array of objective values
- Test case to ensure that cyclades is partitioning updates to non-conflict batches
- Test case to ensure that when optimizing sources, multi_iter gets to a higher sum objective values than both multi_iter with fewer iterations, and one_node_infer
- Todo: Add more test cases with overlapping sources to make sure multi_iter gets to a higher objective value


src/ParallelRun.jl
```diff
@@ -296,8 +299,388 @@ function load_images(rcfs, stagedir)
     images
 end
 
+function union_find(i, components_tree)
```
jeff-regier: We've been trying to use the "bang" notation for functions that modify their arguments. Since `components_tree` is being mutated here, perhaps call this `union_find!`?

src/ParallelRun.jl
```diff
@@ -296,8 +299,388 @@ function load_images(rcfs, stagedir)
...
+        if !haskey(components, index)
+            components[index] = Vector{Int64}()
+        end
+        push!(components[index], src_to_target_index[index_to_source[i-source_start+1]])
+    end
+end
+
+"""
+Partitions sources via the cyclades algorithm. 
+TODO max - refactor into different source file (E.G: Cyclades.jl)
```
jeff-regier: I agree, let's put the cyclades stuff in its own file (but not it's own module) and include it at the top of `ParallelRun.jl` if possible. We've being going all lower case for included files that aren't modules, e..g `src/cyclades.jl`

src/ParallelRun.jl
```diff
@@ -296,8 +299,388 @@ function load_images(rcfs, stagedir)
...
+    end
+    @assert n_sources_assigned == n_sources
+    Log.info("Finished basic source partitioning. Elapsed time: $(toq()) seconds")
+    thread_sources_assignment
+end
+
+"""
+Like one_node_infer, uses multiple threads on one node to fit the Celeste 
+model over numerous iterations. 
+TODO max - refactor into different source file (E.G: Cyclades.jl)? Maybe also rename?
```
jeff-regier: I agree, `multi_iter` is a bit vague. How about `one_node_joint_infer`, since Cyclades optimizes all the light sources jointly, in contrast to `one_node_infer`? Or `one_node_infer_cyclades`?

src/ParallelRun.jl
```diff
@@ -296,8 +299,388 @@ function load_images(rcfs, stagedir)
...
+
+Returns:
+- An array of vectors representing the workload of each thread ([thread][batch][sources]). 
+  Note for partition equally, there is only 1 batch.
+
+"""
+function partition_equally(nprocthreads, n_sources)
+    Log.info("Starting basic source partitioning...")
+    tic()
+    n_sources_per_thread = convert(Int64, floor(n_sources / nprocthreads))
```
jeff-regier: A bit cleaner: `floor(Int64, n_sources / nprocthreads)`

test/test_infer_multi_iter.jl
```diff
@@ -0,0 +1,138 @@
...
+    @assert sum_multi > sum_single
+    @assert sum_multi > sum_two
+end
+
+"""
+helper edges_between_sources
+Helper function to determine if there are edges between 2 sets of sources
+"""
+function edges_between_sources(a, b, map)
+    for v_a in a
```
jeff-regier: One line `for` statements are a bit more compact: `for v_a in a, v_b in b`

test/test_infer_multi_iter.jl
```diff
@@ -0,0 +1,138 @@
...
+    target_sources = [6,7,8,9,10,11,12,13,14,15,16,17,18,19,20,21]
+
+    # 3 threads, 16 sources, batch_size = 4
+    source_assignment = ParallelRun.partition_cyclades(3, target_sources, neighbor_map, batch_size=4)
+
+    # Gather the entities across all batches.
+    # 1. They should sum up to 16 exactly.
+    # 2. They should be numbered from 1 - 16.
+    all_sources =  Vector{Int64}()
+    n_threads = length(source_assignment)
```
jeff-regier: Though the unit tests often just get run with 1 thread. Maybe call this `n_simulated_threads` rather than `n_threads`? And rename `thread_id`  to `simulated_thread_id`, so no one gets confused?

test/test_infer_multi_iter.jl
```diff
@@ -0,0 +1,138 @@
...
+        @assert n_batches == length(source_assignment[thread_id])
+        for batch=1:n_batches
+            for source_id in source_assignment[thread_id][batch]
+                push!(all_sources, source_id)
+            end
+        end
+    end
+    @assert length(all_sources) == 16
+    for source_id=1:16
+        @assert source_id in all_sources
```
jeff-regier: We generally use `@test` rather than `@assert` in the unit tests.

test/test_infer_multi_iter.jl
```diff
@@ -0,0 +1,138 @@
...
+    println("Single iter objective value: $(sum_single)")
+    @assert sum_multi > sum_single
+    @assert sum_multi > sum_two
+end
+
+"""
+helper edges_between_sources
+Helper function to determine if there are edges between 2 sets of sources
+"""
+function edges_between_sources(a, b, map)
```
jeff-regier: `map` is a built-in function in Julia, so it would be more clear to use a variable different name.

test/test_infer_multi_iter.jl
```diff
@@ -0,0 +1,138 @@
...
+                        13=>[9,12,17],
+                        14=>[11,15,18],
+                        15=>[11,14,16,19],
+                        16=>[12,15,17,20],
+                        17=>[13,16,21],
+                        18=>[14,19],
+                        19=>[18,15,20],
+                        20=>[16,19,21],
+                        21=>[17,20])
+
```
jeff-regier: This looks like a really thorough test.

test/test_infer_multi_iter.jl
```diff
@@ -0,0 +1,138 @@
...
+test infer multi iter obj overlapping.
+This makes sure test_infer_multi_iter achieves sum objective value lest than single iter on non overlapping sources. 
+"""
+function test_infer_multi_iter_obj_overlapping()
+
+    # This bounding box has overlapping stars. (neighbor map is not empty)
+    box = ParallelRun.BoundingBox(164.39, 164.41, 39.11, 39.13)
+    field_triplets = [RunCamcolField(3900, 6, 269),]
+    result_multi, obj_values_multi = ParallelRun.one_node_infer_multi_iter(field_triplets, datadir; box=box, n_iters=100)
+    result_multi, obj_values_two = ParallelRun.one_node_infer_multi_iter(field_triplets, datadir; box=box, n_iters=1)
```
jeff-regier: Should this be `n_iters=2`? Even if not, it seems a bit confusing, though maybe I'm just not thinking about it the right way.

src/ParallelRun.jl
```diff
@@ -296,8 +299,388 @@ function load_images(rcfs, stagedir)
...
+    
+    # Read all primary objects in these fields.
+    tic()
+    duplicate_policy = primary_initialization ? :primary : :first
+    catalog = SDSSIO.read_photoobj_files(rcfs, stagedir,
+                        duplicate_policy=duplicate_policy)
+    timing.read_photoobj = toq()
+    Log.info("$(length(catalog)) primary sources")
+
+    reserve_thread[] && thread_fun(reserve_thread)
```
jeff-regier: Are all these `reserve_thread[] && ...` things necessary? @kpamnany --- Does Dtree need a free thread or something or what's the idea here? Presumably we'd rather use all the threads.

src/ParallelRun.jl
```diff
@@ -296,8 +299,388 @@ function load_images(rcfs, stagedir)
...
+    Log.info("Neighbors found in $(toq()) seconds")
+
+    reserve_thread[] && thread_fun(reserve_thread)
+
+    # Partition the sources
+    n_sources = length(target_sources)
+    Log.info("Optimizing $(n_sources) sources")
+    if cyclades_partition
+        # Convert neighbormap to cyclades map (map from source_id -> [neighbor source_ids])
+        cyclades_neighbor_map = Dict{Int64, Vector{Int64}}()
```
jeff-regier: Would it work to make `neighbor_map` a `Vector{Vector{Int64}}` rather than a `Dict`, by giving the sources consecutive integers as their names? And then storing the mapping for new names (i.e. the consecutive integers) to original names in a `Vector{Int64}`? We're try to avoid using `Dicts` whenever possible because they allocate a lot of memory. On the other hand, if this isn't a bottleneck if might not be worth changing.

src/ParallelRun.jl
```diff
@@ -296,8 +299,388 @@ function load_images(rcfs, stagedir)
...
+        for (index, neighbors) in enumerate(neighbor_map)
+            source_id = target_sources[index]
+            cyclades_neighbor_map[source_id] = neighbors
```
jeff-regier: Oh, I see. Could Cyclades use `neighbors` directly without copying it? It's nice to avoid storing the same information twice, to simplify the code.

jeff-regier: Also maybe easier to think about the multi-node case with fewer data structures.

src/ParallelRun.jl
jeff-regier: Another thing about `Dicts` is the we don't have multi-node versions of them. But for arrays, we can create "global" or "distributed" versions that can be accessed from any node of the supercomputer.

src/ParallelRun.jl
```diff
@@ -296,8 +299,388 @@ function load_images(rcfs, stagedir)
...
+            @assert length(ea.active_pixels) > 0
+            model[cur_source_index] = ea
+        end
+    end
+
+    # Initialize elboargs in parallel
+    tic()
+    thread_initialize_sources_assignment = partition_equally(nprocthreads, n_sources)
+    
+    Threads.@threads for i=1:nprocthreads
```
jeff-regier: Do you how much of a speedup you get here for using 2,4, and 8 threads, rather than 1 thread? One issue we've had is that Julia's GC is, for now, single-threaded, so the speedup often isn't as good as we'd like. The overhead is particularly bad for parts of the code that allocate and free a lot of memory.

src/ParallelRun.jl
```diff
@@ -296,8 +299,388 @@ function load_images(rcfs, stagedir)
...
+            # Process every batch of every iteration with nprocthreads
+            Threads.@threads for i = 1:nprocthreads
+                process_sources(thread_sources_assignment[i][batch])
+            end
+        end
+    end
+    Log.info("Done fitting elboargs. Elapsed time: $(toq())")    
+    
+    # Return add results to dictionary
+    # TODO max: parallelize?
```
jeff-regier: I wouldn't, should be very fast single-threaded. Also, I think in the `src_div_par` branch we've stopped using a `Dict` for the results.

jeff-regier (changes requested):
@agnusmaximus -- This all looks great, not just that it works, but that the code is so readable. I've requested some changes inline, mostly very minor. Thanks for the contribution!

agnusmaximus:
Great! I'll fix these up sometime today and update all this

test/test_infer_multi_iter.jl
agnusmaximus: oops yeah it should be 2

src/ParallelRun.jl
agnusmaximus: Done

src/ParallelRun.jl
agnusmaximus: Done

src/ParallelRun.jl
agnusmaximus: Done

src/ParallelRun.jl
agnusmaximus: Done, using one_node_joint_infer

src/ParallelRun.jl
agnusmaximus: I see, it seems like a vector of InferResult is used instead of a map. Should I wait until that branch is merged with master to change it?

test/test_infer_multi_iter.jl
agnusmaximus: Done

test/test_infer_multi_iter.jl
agnusmaximus: Done

test/test_infer_multi_iter.jl
agnusmaximus: That's cool! Did not know you could do that. Done

test/test_infer_multi_iter.jl
agnusmaximus: Done

test/test_infer_multi_iter.jl
agnusmaximus: Done

src/ParallelRun.jl
agnusmaximus: Hm, it seems like doing the partitioning is relatively very fast compared to allocation and optimization (0.000236101 seconds vs 209.690328562s for 104 sources). I'm thinking maybe we won't need to parallelize the connected components code for single node because of this. What do you think?

src/ParallelRun.jl
agnusmaximus: Regarding the multi-node case, new code has to be written to right?

agnusmaximus: I think the neighbors array isn't copied, but a reference to it is stored in cyclades_neighbor_map

agnusmaximus:
Let's not merge this just yet, I get a seg fault with 8/12 threads for some reason (4 threads seems to work fine... I'll need to check this out later)

agnusmaximus:
Wait hmmm.... Running with 8 threads on the login nodes succeeds... I think it might be the way I'm running the batch script on the regular nodes that might be causing the seg fault... I'll look into that

agnusmaximus:
Hm yeah, not sure why but sometimes with > 4 threads I get a seg fault: http://pastebin.com/eNw3nTxE. Has anything like this been seen before? It seems to be related to the garbage collector.

kpamnany:
Do you get that segfault on a single node Max? How repeatable is it?

gostevehoward:
In the scaling runs two weeks ago we saw a lot of segfaults but they only happened in 64+ node runs as I understand so we were never able to track down the root cause. If you have access to our slack channel take a look at the history for some more context. If you can provide detailed instructions on how to reliably reproduce the segfault with one (or few) nodes that might be of great value in tracking down the problem.

agnusmaximus:
The segfaults seem to happen pretty often when I run with > 4 threads on the joint_infer test case. If you check out the multi_iter branch, and set num threads to be something like 12 ("export JULIA_NUM_THREADS=12") then "julia ./test/runtests.jl joint_infer" it should produce the error after running for like 10-15 minutes I think... (They happen pretty often, I have not gotten 12 threads to finish executing without running into the error that I linked in pastebin)

agnusmaximus:
Update: Hmmm 12 threads just finished for me... weird. So it's not super reliable...
Update 2: Dang 6 threads just finished for me (on the login node), let me try it out using the batch script

kpamnany:
I _think_ this is a race condition somewhere in Julia's runtime. And clearly, the particular interleaving that causes a problem doesn't happen frequently; the more the threads, the higher the probability. Good to know there's a single-node case that can produce it though. Please snapshot/tag the commit where you know it happens (and don't delete this branch please).

@tkelman: would you see if you can reproduce this/have Jameson take a look?

jeff-regier:
@agnusmaximus  That segfault is a nightmare, at least if it turns out the be the same one we were stuck on during the IPDPS submission. @tkelman and @vtjnash are really good at fixing these segfaults if we can given them simple test cases that reliably exhibit the problem. So far we haven't had much luck isolating it though. If you have a single-node test case that might be the closest we've come.

agnusmaximus:
I've pushed bin/batch_joint_infer.sl which just runs the joint_infer method test case on 8, and 12 (piping output to infer_multi_%d_thread). Usually one of these fails. Not sure how helpful this batch script will be in locating the segfault

agnusmaximus:
Also regarding the segfault, I think Kiran is right in that it's some sort of race condition when accessing shared memory from multiple threads. If I remember right, when I used separate data per thread in multiple iteration training via cyclades (so threads did not share data) running on 1,8,12 threads finished fine. 

Edit: Actually, I have no idea... haha

jeff-regier (approved):

jeff-regier:
Thanks for the revisions Max.

agnusmaximus:
For batch_joint_infer.jl to run right, you need to add "const dt_nodeid = 1" on line 26 of ParallelRun.jl. Sorry about this!

PR #389, opened 2016-11-02
make GalSim PSF larger to avoid pdf point sampling problems

### LINKED ISSUE #369: Debug why GalSim stars are predicted to be galaxies

gostevehoward:
Fixes #369. See comments in that issue for all the details.

Also making the colors used in GalSim match the prior modes rather than prior medians.

I uploaded a new FITS file to the portal.nersc site.

### DIFF
--- a/benchmark/galsim/GalsimBenchmark.jl
+++ b/benchmark/galsim/GalsimBenchmark.jl
@@ -15,7 +15,7 @@ const FILENAME = "output/galsim_test_images.fits"
 function make_psf()
     alphaBar = [1.; 0.]
     xiBar = [0.; 0.]
-    tauBar = [1. 0.; 0. 1.]
+    tauBar = [16. 0.; 0. 16.]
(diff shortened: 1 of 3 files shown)


jeff-regier:
Colors are mvnormal, so mode = median. (Color is on the log scale: the log ratio of fluxes in adjacent bands.) But for the fluxes, mode != median. Are you using the fluxes' mode now?

gostevehoward:
Oh I see, I had been thinking of "color" as referring to just the ratio of
fluxes, and having it modeled as lognormal. For reference band flux, I
don't use the exact mean or median or mode, I just use values that are
reasonably typical for stars/galaxies. But for colors I was trying to set
it to the exact mode while I was debugging the issues. I guess the new
colors I'm using aren't necessarily sensible then, but I also think it
shouldn't affect results much. Probably worth changing though just for
clarity's sake.

PR #392, opened 2016-11-03
Update Celeste to Julia master syntax

andreasnoack:
This also enables Travis testing against Julia master.

PR #394, opened 2016-11-04
Adjust to upcoming changes in Optim

andreasnoack:
They finally correct the use of `minimum` which breaks code that explicitly accesses the field. Using accessor functions works on both the released version and the master branch of Optim.

~~Notice, this PR includes commits that are also in #392 so that PR should get merged before this one.~~ Rebased to only include the `Optim` fixes.


jeff-regier:
Did one of your commits get excluded from both PRs? Before you rebased, I noticed some changes where you switched the order of iterations, so loop over rows in the inner loop rather than columns. I'd like to include those changes too.

andreasnoack:
I just removed it from this PR. I'm doing some general profiling right now and I'l open a new PR soon.

PR #397, opened 2016-11-06
Reduce the number of allocations

andreasnoack:
I've been looking into reducing the number of allocations and thereby hopefully reduce the GC work. My idea was that the output from `--track-allocation` might be a little distorted since it is dominated by large data allocations. I was more interested in finding the small allocations that are repeated many times so I tweaked `--track-allocation` to report `1` whenever a positive allocation happened. I'll post the results below but first the overall results by commit for single threaded runs

#### Current master
10.174977 seconds (4.08 M allocations: 623.774 MB, 7.60% gc time)
10.149520 seconds (4.08 M allocations: 623.774 MB, 7.68% gc time)

#### This PR
9.980523 seconds (2.83 M allocations: 543.700 MB, 6.86% gc time)
10.027899 seconds (2.83 M allocations: 543.700 MB, 7.42% gc time)

So the total number of allocations is reduced quite a bit with these commits. Unfortunately, the effect on the running time is not that big but I still hope that this will help for larger runs. I think most of these changes are harmless but you can judge the changes one-by-one to see if you are okay with them.

There are still some big allocators left in the code. The allocations from master are shown below. Among the most frequently allocating lines, I've only fixed the lines that uses `intersect`. It will require more substantial changes to the code if we want to get rid of the rest. I think the challenge is frequent use of complicated data structures. E.g. an array of composite types. There the gc will have to register all the elements, but when extracting an element there is a new reference to a heap allocated object and the gc will have to register it again. Julia will probably become better in handling such situations over time but we might also want to think about how we structure the data.

```
    10872     positive_constraint = (pb.ub == Inf)
    11200     sig_sf.j .*= gc.nuBar
    11200       sig_sf.t .*= gc.nuBar
    12648   cos_sin = cos(e_angle)sin(e_angle)
    12648   GalaxySigmaDerivs(j, t)
    12648   j = Array(NumType, 3, length(gal_shape_ids))
    12648   t = Array(NumType, 3, length(gal_shape_ids), length(gal_shape_ids))
    16056                 jac, hess = Transform.box_derivatives(
    19992   bmc = BvnComponent{NumType}(
    19992     e_dev_dir::Float64
    20792         var_G.v[1] += elbo_vars.var_G_s.v[1]
    22868       ThisNumType = promote_type(T1, T2, T3)
    24084                     diagm(hessian_values .* sf.d[:])
    37944     j[i, gal_shape_ids.e_angle] =
    37944     j[i, gal_shape_ids.e_axis] =
    37944     j[i, gal_shape_ids.e_scale] = (2XiXi ./ e_scale)[[1, 2, 4][i]]
    37944       t[i, gal_shape_ids.e_angle, gal_shape_ids.e_angle] =
    37944       t[i, gal_shape_ids.e_angle, gal_shape_ids.e_axis] =
    37944       t[i, gal_shape_ids.e_axis, gal_shape_ids.e_axis] =
    37944       t[i, gal_shape_ids.e_scale, gal_shape_ids.e_scale] =
    38623     pixels = img.pixels[h_range, w_range]
    38625     epsilon_mat = img.epsilon_mat[h_range, w_range]
    38625     ImageTile(img.b,
    59976   const empty_sig_sf =
    79048         combine_sfs!(
    79048         elbo_vars.E_G.v[1] += tile.epsilon_mat[h, w]
    81288     if calculate_hessian
   117836     x = @SVector Float64[tile.h_range[h], tile.w_range[w]]
   154500     @assert length(patch_ctrs) == length(patch_radii_px)
   154500         idx = get_local_sources(image_tiles[hh, ww],
   154500             if length(intersect(target_sources, tile_source_map)) > 0
   154500         out[hh, ww] = cands[idx]
   154500     ret = Int[]
   154500     tile_source_map = Int[]
```

### DIFF
--- a/src/Infer.jl
+++ b/src/Infer.jl
@@ -154,6 +154,39 @@ function get_tile_source_map(images::Vector{TiledImage},
     patches, tile_source_map
 end
 
+"""
+Test if the there are any intersections between two vectors.
+"""
+function isintersection_sorted(x::AbstractVector, y::AbstractVector)
+    nx, ny = length(x), length(y)
+    if nx == 0 || ny == 0
+        return false
+    end
+    @inbounds begin
+        i = j = 1
+        xi, yj = x[1], y[1]
+        while true
+            if xi < yj
+                if i == nx
+                    return false
+                else
+                    i += 1
+                    xi = x[i]
+                end
+            elseif xi > yj
+                if j == ny
+                    return false
+                else
+                    j += 1
+                    yj = y[j]
+                end
+            else
+                return true
+            end
+        end
+    end
+end
+
 
 """
 Updates patches in place with fitted psfs for each active source.
@@ -178,7 +211,7 @@ function fit_object_psfs!{NumType <: Number}(
         for tile_source_map in ea.tile_source_map[i]
             # check if *any* of this tile's sources are a target, and
             # if so, add *all* the tile sources to the output.
-            if length(intersect(target_sources, tile_source_map)) > 0
+            if isintersection_sorted(target_sources, tile_source_map)
                 relevant_sources = union(relevant_sources, tile_source_map)
             end
         end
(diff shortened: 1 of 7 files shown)


jeff-regier:
Looks good, thanks for the PR, I think these small changes will add up. I'm going to try to remove some of those larger allocations today.

Issue #406, opened 2016-11-10
restore unit tests


jeff-regier:
The following unit test are commented out currently. They should be restored or replaced.

- [ ] test_derivatives.jl:test_combine_pixel_sources()
- [ ] test_derivatives.jl:test_add_log_term()
- [x] test_derivatives.jl:test_e_g_s_functions()
- [ ] test_derivatives.jl:test_transform_sensitive_float()
- [x] test_derivatives.jl:test_elbo()
- [ ] test_elbo.jl:test_that_variance_is_low()
- [ ] test_images.jl:test_stamp_get_object_psf()
- [ ] test_images.jl:test_set_patch_size()
- [ ] test_optimization.jl:test_real_stamp_optimization()

Issue #382, opened 2016-10-31
Don't recompute `tile_source_map` for each source


jeff-regier:
The worst 3 lines for memory allocations from processing a whole RCF (not just `benchmark_infer.jl`) are all related to creating the `tile_source_map`. 

```
12533040000     ret = Int[]
12534015792             if length(intersect(target_sources, tile_source_map)) > 0
28088728944     tile_source_map = Int[]
```

Currently `tile_source_map` is recomputed for every light source. Then the map only includes itself and its neighbors, which is nice. But it's too slow.

Issue #346, opened 2016-10-12
SkyPatch -> SubImage


jeff-regier:
Add `epsilon`, `pixels`, and `is_active_pixel` (a bitmap) to `SkyPatch`, and rename `SkyPatch` to `SubImage` (or `Stamp`?). In the elbo and log_prob, never access the `TiledImage`--only get data from the `SubImage`.

When evaluating the elbo or the log prob, for each pixel in the sub-image, we check whether it's active and skip it if it isn't. For each active pixel, we then check each of its neighbors, to see if they are active in that pixel too.

Alternatively, rather than checking each neighbor, we could add a `background` matrix to `SubImage`, that stores the combined brightness from neighbors, thus addressing #280. But the `background` would have to be kept up-to-date so this might be a pain.

We can remove tiles (and `TiledImage`, and `ImageTile`, etc.) from the code base one `SubImage` replaces `SkyPatch`, i.e., #316.

Issue #310, opened 2016-09-21
Use log probability to determine the active pixels, not the elbo


jeff-regier:

jeff-regier:
Following PR #408, using log probability to determine active pixels doesn't seem so essential. It's probably better to focus on #157 -- initializing active pixels without an existing catalog.

PR #408, opened 2016-11-10
Remove tiles

### LINKED ISSUE #316: Get rid of tiles
For a rasterized psf (https://github.com/jeff-regier/Celeste.jl/issues/87), tiles don't make a lot of sense, mostly because the light from pixels in a particular tile extend beyond that tile after convolution with the PSF.

Right now the main reason we have tiles is to limit the sources that we process (via `tile_source_map`).  Instead of this, we could just have a vector of vectors from sources to the sources that are nearby according to the same logic used to decide whether sources affect a tile.

In case sources affect a lot of tiles, we could additionally add some logic in all the functions that evaluate values and derivatives of `BvnComponent`s and `GalaxyCacheComponents` to return zero automatically if the point is some number of standard deviations away from its mean.  That is probably a good idea in its own right.

This could result in considerable simplification of the `Image` structure and pre-processing, too.

@jeff-regier , @gostevehoward , any thoughts or things that I have missed?

jeff-regier:
This PR removes everything related to tiles, e.g., `TiledImage`, `tile_source_map`, and `ImageTile`, to close #316, #346, and #382. These data structures were some of the least memory efficient, as shown by `benchmark_infer.jl`.

On master
```
  5.167481 seconds (3.28 M allocations: 546.604 MB, 15.05% gc time)
```

On the notiles branch (this PR):
```
  3.855813 seconds (1.80 M allocations: 316.182 MB, 5.49% gc time)
```

Memory allocation falls by ~40% and gc time falls by ~60%.

But `benchmark_infer.jl` may overstate the importance of not copying/tiling each image, because it only optimizes 4 light sources. However, `benchmark_rcf.jl` shows that the improvement persists when we process a whole RCF, albeit one with relatively few sources.

On master
```
152.511854 seconds (283.88 M allocations: 19.706 GB, 20.77% gc time)
```

On the notiles branch:
```
105.759341 seconds (94.08 M allocations: 8.588 GB, 4.33% gc time)
```

The new code is cleaner too: tiles were a convoluted way to partition images. This PR removes nearly 1000 lines of code in net from the codebase, including some of the most confusing code. In the old code, images were divided into 20 pixel x 20 pixel tiles. Each light source only "influenced" a subset of the tiles.  We had a map from each light source to its tiles, and from each tile to its light sources. The indexes of pixels within these tiles that were "active" were recorded in a vector.

The new code maps each light source to the rectangle of pixels in each region it might influence. For each pixel in this rectangle, we maintain a bitmap, true if the pixel is "active" (i.e. relevant to the objective function) and false otherwise. There's no longer an explicit map stored from regions of the images to light sources. Rather, where we need this map in the code, we iterated through all sources and test whether they list the pixel as active. This loop is very fast in practice because we only process nearby light sources (neighbors) together, and because looping doesn't allocate any memory.

Though this PR moves us forward, it introduces a few regressions. 1) I ran out of time before updating every unit test, so I left some commented out and opened issue #406. 2) I refactored `elbo_objective.jl` because I was struggling to understand what was going on. And while I think the new version is cleaner and more efficient, it creates work for anyone familiar with, or basing code on, how it was. 3) While I stopped using the elbo to determine active pixels (#310), I used kind of a hack. I make the pixels observe to be bright the active pixels. That biases the results. But `validate_on_stripe82.jl` holds up, so I think it's OK for now, though we'll want to explore subsampling pixels more systematically soon.


gostevehoward:
Epic change. Nice work.

PR #411, opened 2016-11-11
GalSim test cases w/ misspecified PSF, also update to remove TiledImage

gostevehoward:
I added an optional `loc_width` param to infer_source() to make it reusable from GalsimBenchmark, does that seem OK?

### DIFF
--- a/benchmark/galsim/GalsimBenchmark.jl
+++ b/benchmark/galsim/GalsimBenchmark.jl
@@ -202,7 +202,7 @@ function make_catalog_entries(header::FITSIO.FITSHeader)
     num_sources = header["CL_NSRC"]
     for source_index in 1:num_sources
         if num_sources == 1
-            initial_position = [36, 36]
+            initial_position = [19, 19]
         else
             initial_position = [
                 header[string("CL_X", source_index)],
--- a/benchmark/galsim/GalsimBenchmark.jl
+++ b/benchmark/galsim/GalsimBenchmark.jl
@@ -5,7 +5,7 @@ import FITSIO
 import StaticArrays
 import WCS
 
-import Celeste: Infer, Model, DeterministicVI
+import Celeste: Model, DeterministicVI
 
 const FILENAME = "output/galsim_test_images.fits"
 
@@ -46,26 +46,23 @@ function read_fits(filename; read_sdss_psf=false)
     extensions, wcs
 end
 
-function make_tiled_images(band_pixels, psf, wcs, epsilon, iota)
+function make_images(band_pixels, psf, wcs, epsilon, iota)
     # assume dimensions equal for all images
     height, width = size(band_pixels[1])
     [
-        Model.TiledImage(
-            Model.Image(
-                height,
-                width,
-                band_pixels[band],
-                band,
-                wcs,
-                psf,
-                0, # SDSS run
-                0, # SDSS camcol
-                0, # SDSS field
-                fill(epsilon, height, width),
-                fill(iota, height),
-                Model.RawPSF(Array(Float64, 0, 0), 0, 0, Array(Float64, 0, 0, 0)),
-            ),
-            tile_width=48,
+        Model.Image(
+            height,
+            width,
+            band_pixels[band],
+            band,
+            wcs,
+            psf,
+            0, # SDSS run
+            0, # SDSS camcol
+            0, # SDSS field
+            fill(epsilon, height, width),
+            fill(iota, height),
+            Model.RawPSF(Array(Float64, 0, 0), 0, 0, Array(Float64, 0, 0, 0)),
         )
         for band in 1:5
     ]
@@ -202,7 +199,8 @@ end
 
 function make_catalog_entries(header::FITSIO.FITSHeader)
     catalog_entries = Model.CatalogEntry[]
-    for index in 1:header["CL_NSRC"]
+    num_sources = header["CL_NSRC"]
+    for source_index in 1:num_sources
         if num_sources == 1
             initial_position = [36, 36]
         else
@@ -216,22 +214,7 @@ function make_catalog_entries(header::FITSIO.FITSHeader)
     catalog_entries
 end
 
-
-# TODO: this code seems to be obsolete?
-# this code is very close to Infer.infer_source() but avoids PSF fitting
-function infer_source(band_images::Vector{Model.TiledImage},
-                      catalog_entries::Vector{Model.CatalogEntry},
-                      verbose::Bool)
-    vp = Vector{Float64}[Model.init_source(entry) for entry in catalog_entries]
-    patches, tile_source_map = Infer.get_tile_source_map(band_images, catalog_entries)
-    elbo_args = DeterministicVI.ElboArgs(band_images, vp, tile_source_map, patches, [1])
-    Infer.load_active_pixels!(elbo_args)
-    DeterministicVI.maximize_f(DeterministicVI.elbo, elbo_args, verbose=verbose, loc_width=9.0)
-    variational_parameters::Vector{Float64} = vp[1]
-    variational_parameters
-end
-
-function main(; test_case_name=Nullable{String}(), verbose=false)
+function main(; test_case_name=Nullable{String}())
     all_benchmark_data = []
     psf = make_psf()
     extensions, wcs = read_fits(FILENAME)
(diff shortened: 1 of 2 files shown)


jeff-regier:
Rather than setting `loc_width`, how about always using the default value of `loc_width` in benchmarks/unit tests, and adjusting the initializations and the galsim images' WCS transforms to make the default appropriate? That might make for a more realistic testing/benchmarking setup. In practice we won't initialize to more 2 or 3 pixels from the true location of a light source, and arcseconds per pixel is always 0.396 in SDSS. I'm also glad to just merge this as is if it's a pain to modify the benchmarks, though in general I prefer to keep function interfaces simple.

gostevehoward:
No I have no problem with that. But the default `loc_width` value
https://github.com/jeff-regier/Celeste.jl/blob/master/src/deterministic_vi/maximize_elbo.jl#L157
is
0.0015, or .0015/.396 = 0.0038 pixels, which seems tiny. I must be
misunderstanding the units of `loc_width`?

jeff-regier:
It may be in degrees instead. 3600 arcseconds per degree. Yes, that seems about right.

gostevehoward:
I changed the pixel scale to 0.396 but `loc_width` does seem to be 0.0015 arcsec, not 0.0015 degrees = 5.4 arcsec:

```
│ Row │ label         │ field                        │ expected │ actual   │
├─────┼───────────────┼──────────────────────────────┼──────────┼──────────┤
│ 1   │ "simple_star" │ "X center (world coords)"    │ 19.206   │ 19.0015  │
│ 2   │ "simple_star" │ "Y center (world coords)"    │ 19.206   │ 19.0015  │
```

The (world-coords) position (in arcsec, right?) was initialized as (19, 19) and note the inferred position is exactly (19.0015, 19.0015).

jeff-regier:
The only place `loc_width` gets used (as opposed to passed around) is on line 672 of `Transform.jl`. On that line, `loc_width` is added to and subtracted from `vp[s][ids.u]`. So the units of `loc_width` are the same as the units of the variational parameters. For actual data, the units of `vp[s][ids.u]` are always in degrees. Let's stick with that convention for benchmarks and unit tests too if possible. That keeps our tests and actual use cases from diverging. (Or am I misunderstanding the issue?)

gostevehoward:
I have no problem with that, but I must be misunderstanding some important thing. Maybe I need to learn to understand the WCS better. Let me dig around and get back to this thread.

jeff-regier:
Figure 1 might help: http://www.atnf.csiro.au/people/mcalabre/WCS/wcs.pdf

It's pretty much just an affine transformation though: `y = Ax + b`, where `y` is world coordinates and `x` is pixel coordinates. Here `x`, `y`, and `b` are vectors of length 2, and `A` is a 2x2 matrix.

I suspect that the A matrix, given to you by GalSim, has the wrong scale for the Celeste code base. The SDSS images may be higher resolution than the implicit resolution of the images generated by GalSim. (Resolution meaning pixels per square arcsecond.) Just a thought though.

gostevehoward:
OK, now I understand that the "world coordinates" given by a WCS do not have any inherent units associated with them. You set the scaling of your WCS to put the world coordinates in your desired units. So GalSim's convention, which I was following since I naively assumed it was standard, was to have the WCS work with world coords in arcsec. I guess the WCS for SDSS images uses world coords in degrees, so this is what Celeste generally assumes? This didn't actually break almost anything, because most all coordinates pass through the WCS so you can drop in an image with a WCS scaled 3600x different and everything just adapts...except that hard-coded `loc_width` value. (The Celeste PSF is specified in pixel coords, so that side-stepped the issue.)

Of course there is one other coordinate-scale value in the model, the half-light radius, which is now wildly off between Celeste's output and the test case expectation. I would have thought the half-light radius in `vp` would be in world coords and would scale with the WCS...maybe it's in pixel coords? I'll look into it.

jeff-regier:
The half light radius (`e_scale`) is in arc seconds. That's kind of confusing given our whole discussion about how location coordinates are in degrees, but it's the way the existing SDSS catalogs report scales.

gostevehoward:
Ah...I've been crawling through the code and was just about to conclude it's in pixel coords. How can Celeste put it in arcsec if it doesn't know the units of the WCS world coords (per my comment above)? Is there a 1/3600 hardcoded somewhere?

As far as I can tell, `e_scale` is really a scaling factor applied to the standard deviations of the galaxy prototype components (in `get_galaxy_prototypes()`). These eventually are used to form bivariate normals and fill in `fs1m` which is used to determine expected pixel brightness. When computing `fs1m` (in `populate_fsm_vecs!()`), we are given `w` and `h`, which seem to be pixel coords, are are eventually passed to `eval_bvn_pdf!()` (as the 2-vector `x`), where they are used to evaluate the above-mentioned BVN. I haven't been able to find where this value would get converted from pixel to world coords in some way. I could use a hint at this point :-/

jeff-regier:
It's hard coded. `grep -r 0.396 .` in `src` shows the lines. It's mainly the one in `SDSSIO.jl`.

To evaluate the objective function, I'm not sure anything is ever converted from pixel coordinates to world coordinates. It goes the other direction. The parameters start in world coordinates (or, more generally, physical units). In the process of computing the expected brightness at each pixel, we convert repeatedly from physical units to image-specific units.

The derivatives and Hessians that are computed along with the objective function might contain some calculations that look like inverting the WCS->pixel transformation. It's usually enough to know that the derivative/Hessian calculations are correct though.

jeff-regier:
To clarify, 0.396 is arc seconds per pixel. It's for `e_scale`, which is stored in arc second rather than degrees. The location `u` is stored in degrees, but I wouldn't expect to see something like `1 / 3600` in the code, because that conversion is read from the image FITS files. It's implicit in the WCS data that comes with each image.

gostevehoward:
@jeff-regier I'm still confused about `e_scale`. It still seems like it's in pixel coordinates to me, both from reading the code, and from these two pieces of evidence:

First, in the [one line of `SDSSIO.jl` where 0.396 is hardcoded](https://github.com/jeff-regier/Celeste.jl/blob/master/src/SDSSIO.jl#L559), it seems to be converting a radius in arcsec (from the SDSS catalog) to a radius in pixels, which it then stores in the `CatalogEntry`. So it seems `CatalogEntry.gal_scale` holds a radius in pixels, not arcsec. That `gal_scale` value is then [used to initialize `e_scale` in the variational params](https://github.com/jeff-regier/Celeste.jl/blob/master/src/model/light_source_model.jl#L180).

Second, I tried running a GalSim test case with 0.396 arcsec/pixel, as in SDSS, and a single galaxy with a half-light radius of 3 * .396 = 1.19 arcsec, or 3 pixels. Celeste infers `e_scale = 4.5`:

```
 1   │ "small_galaxy" │ "X center (world coords)"    │ 0.005335 │ 0.005335 │
│ 2   │ "small_galaxy" │ "Y center (world coords)"    │ 0.005335 │ 0.005335 │
│ 3   │ "small_galaxy" │ "Minor/major axis ratio"     │ 0.4      │ 0.406235 │
│ 4   │ "small_galaxy" │ "Angle (degrees)"            │ 0        │ 0.0      │
│ 5   │ "small_galaxy" │ "Half-light radius (arcsec)" │ 1.188    │ 4.5051   │
│ 6   │ "small_galaxy" │ "Brightness (nMgy)"          │ 10       │ 9.98041  │
│ 7   │ "small_galaxy" │ "Probability of galaxy"      │ 1        │ 0.995    │
```

I then tried changing the WCS to 100 arcsec/pixel and giving the galaxy a half-light radius of 300 arcsec = 3 pixels. Celeste infers the same `e_scale = 4.5`:

```
 1   │ "small_galaxy" │ "X center (world coords)"    │ 1.34722  │ 1.34722  │
│ 2   │ "small_galaxy" │ "Y center (world coords)"    │ 1.34722  │ 1.34722  │
│ 3   │ "small_galaxy" │ "Minor/major axis ratio"     │ 0.4      │ 0.406235 │
│ 4   │ "small_galaxy" │ "Angle (degrees)"            │ 0        │ 0.0      │
│ 5   │ "small_galaxy" │ "Half-light radius (arcsec)" │ 300      │ 4.5051   │
│ 6   │ "small_galaxy" │ "Brightness (nMgy)"          │ 10       │ 9.98041  │
│ 7   │ "small_galaxy" │ "Probability of galaxy"      │ 1        │ 0.995    │
```

What am I missing?

jeff-regier:
@gostevehoward  You're right, I remembered incorrectly. `e_scale` is indeed in pixel coordinates. Sorry about that. Nice job figuring it out.

PR #417, opened 2016-11-14
Various performance enhancements by making Julia code more idiomatic

jrevels:
This PR makes a bunch of small changes and a few bigger changes in order to make the Julia code more idiomatic, and thus increase performance. Here's a summary of the changes:

- refactor some allocating methods to reuse memory when possible
- refactor some methods to avoid redundant computations
- avoid using types like classes (e.g. remove `Function` fields)
- manually hoist pointer loads (Julia's compiler isn't smart enough yet to determine when a binding won't get de-aliased, so you have to manually do this if you don't want to duplicate loads)
- the `v` field of the `SensitiveFloat` type is now `::Base.RefValue{NumType}` instead of `::Vector{NumType}`. I initially tried just making it `NumType`, which would be the idiomatic thing to do, but that ended up making things very slow for some unknown reason (I believe @jeff-regier already knows about this).

I mainly touched `PSF.jl`, `Transform.jl`, and the files in `src/deterministic_vi`. The modifications to most of the other files are due to find-and-replace (e.g. `.v[1]` --> `.v[]`). There are other places in the code that could've benefited from the above changes as well, but this PR was already really large and I didn't want to go further overboard.

I ran `benchmark_infer.jl`, and it seems this PR gives a ~25% speedup with a single thread, and a ~30% speedup with 4 threads (running on Julia v0.6):

```
on master (1 thread): 4.432900 seconds (1.78 M allocations: 316.157 MB, 6.07% gc time)
on master (4 threads): 3.076258 seconds (1.69 M allocations: 312.785 MB, 16.95% gc time)

on jr/idiomatic2 (1 thread): 3.526573 seconds (1.95 M allocations: 284.098 MB, 7.39% gc time)
on jr/idiomatic2 (4 threads): 2.358118 seconds (1.80 M allocations: 279.312 MB, 11.32% gc time)
```

I haven't made much effort in getting tests to match the new code, so they probably won't pass until I fix them. At the very least, `benchmark_infer.jl` is giving me close to the same output on both this branch and master, so hopefully it'll just be a matter of code churn rather than bug fixing.

### DIFF
--- a/src/Transform.jl
+++ b/src/Transform.jl
@@ -289,7 +289,7 @@ Members:
 """
 type TransformDerivatives{NumType <: Number}
     dparam_dfree::Matrix{NumType}
-    d2param_dfree2::Array{Matrix{NumType}}
+    d2param_dfree2::Vector{Matrix{NumType}}
     Sa::Int
 
     # TODO: use sparse matrices?
@@ -497,18 +497,10 @@ Args:
 Returns:
  - Update vp_free in place.
 """
-function array_to_free_vp!{NumType <: Number}(
-    xs::Matrix{NumType}, vp_free::FreeVariationalParams{NumType},
-    omitted_ids::Vector{Int})
-
-    left_ids = setdiff(1:length(UnconstrainedParams), omitted_ids)
-    P = length(left_ids)
-    S = length(vp_free)
-    @assert size(xs) == (P, S)
-
-    for s in 1:S, p1 in 1:P
-        p0 = left_ids[p1]
-        vp_free[s][p0] = xs[p1, s]
+function array_to_free_vp!{T}(xs::Array{T}, vp_free::FreeVariationalParams{T},
+                              kept_ids::Vector{Int}, omitted_ids::Vector{Int})
+    for s in 1:length(vp_free), p1 in 1:length(kept_ids)
+        vp_free[s][kept_ids[p1]] = xs[p1, s]
     end
 end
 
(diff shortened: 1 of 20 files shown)


jeff-regier:
Hi Jarrett, Great changes, thanks for the contribution. I ran `validate_on_stripe82.jl` to confirm that none of them change the scientific results. (They don't.) I ran `benchmark_rcf.jl` to get another perspective on how this will change runtime.

On master:

```
104.746012 seconds (95.70 M allocations: 8.689 GB, 2.39% gc time)
```

With this PR:

```
 90.919500 seconds (103.66 M allocations: 8.699 GB, 4.73% gc time)
```

The 15% decrease in run time is a big deal. That should save us around 50,000 cpu hours later this week when I do a large scale run.

Any idea why the number of allocations increase with this PR? Perhaps a lot of really small allocations somewhere? (The number of allocations grows by 10%, but the amount allocated grows by only 1%.)

src/deterministic_vi/elbo_objective.jl
```diff
@@ -36,39 +36,54 @@ function calculate_source_pixel_brightness!{NumType <: Number}(
     @inbounds for i in 1:Ia # Stars and galaxies
         fsm_i = (i == 1) ? fs0m : fs1m
         a_i = ea.vp[s][ids.a[i, 1]]
+        sb_E_l_a_b_i = sb.E_l_a[b, i]
```
andreasnoack, line 39: Did you try to benchmark the effect of these loads? I tried a few without where I couldn't see an effect and in contrast to most of the other changes which make the code nicer, this makes it quite a bit uglier.

jrevels: Yeah, I did most of them in one fell swoop so there might be some inconsequential ones in there, especially since some of these loads are via fields of an immutable. If you find ones that don't matter, let me know and I'll revert them.

andreasnoack: I haven't tried on this PR. I just tried a few in my own pass through the code last week where I couldn't see an effect.

jrevels:
> Any idea why the number of allocations increase with this PR? Perhaps a lot of really small allocations somewhere? (The number of allocations grows by 10%, but the amount allocated grows by only 1%.)

I noticed that as well. Before I pulled `master` and rebased, this PR reduced allocations vs. `master` from a couple days ago - now it seems the tides have turned. I probably slipped up somewhere, I'll see if I can pinpoint where the extra allocations are coming from.

src/Transform.jl
```diff
@@ -538,121 +530,117 @@ active_sources: The sources that are being optimized.    Only these sources'
...
-    function to_vp!{NumType <: Number}(
-            vp_free::FreeVariationalParams{NumType}, vp::VariationalParams{NumType})
-        @assert length(vp_free) == active_S
-        for si=1:active_S
-            s = active_sources[si]
-            free_to_vp!(vp_free[si], vp[s], bounds[si])
-        end
-    end
+function from_vp{T}(dt::DataTransform, vp::VariationalParams{T})
+    vp_free = fill(zeros(T, length(ids_free)), length(dt.active_sources))
```
jeff-regier: This may be the source of the problem: `fill(zeros(42), 5)` creates 5 references to the same array of 42 zeros, whereas `[zeros(42) for i in 1:5]` creates 5 references to 5 different arrays of zeros. It's the latter we want. Each light source has different parameters.

src/Transform.jl
```diff
@@ -538,121 +530,117 @@ active_sources: The sources that are being optimized.    Only these sources'
...
-                zeros(NumType, length(CanonicalParams)) for s = 1:S]
-            to_vp!(vp_free, vp)
-            vp
-    end
+function to_vp{T}(dt::DataTransform, vp_free::FreeVariationalParams{T})
+    n_active_sources = length(dt.active_sources)
+    @assert(n_active_sources == dt.S,
+            string("to_vp is not supported when active_sources is a ",
+                   "strict subset of all sources."))
+    vp = fill(zeros(T, length(ids_free)), n_active_sources)
```
jeff-regier: This could be the source of the problem too. In addition to creating reference to arrays,  `length(ids_free) != length(CanonicalParams)`. We had `length(CanonicalParams)` here before.

jeff-regier: You could also replace `length(ids_free)` with `length(ids)`, since

```
length(Celeste.Model.ids) == length(Celeste.Model.CanonicalParams)
```

test/test_transforms.jl
```diff
@@ -87,7 +89,7 @@ function test_parameter_conversion()
 	transform1 = Transform.get_mp_transform(ea1.vp, ea1.active_sources)
 
 	@assert transform1.S == ea.S
-	@assert transform1.active_S == 1
+	@assert length(transform.active_sources) == 1
```
jeff-regier: This should be `transform1` rather than `transform`.

src/deterministic_vi/elbo_objective.jl
```diff
@@ -173,14 +188,27 @@ function calculate_source_pixel_brightness!{NumType <: Number}(
...
+            E_G2_s_h[ids_u_1, ids_u_1] = E_G2_s_hsub_vec_1_u_u[1, 1] + E_G2_s_hsub_vec_2_u_u[1, 1]
+
+            E_G_s_h[ids_u_1, ids_u_2] = E_G_s_hsub_vec_1_u_u[1, 2] + E_G_s_hsub_vec_2_u_u[1, 2]
+            E_G2_s_h[ids_u_1, ids_u_2] = E_G2_s_hsub_vec_1_u_u[1, 2] + E_G2_s_hsub_vec_2_u_u[1, 2]
+
+            E_G_s_h[ids_u_2, ids_u_1] = E_G_s_hsub_vec_1_u_u[2, 1] + E_G_s_hsub_vec_2_u_u[2, 1]
+            E_G2_s_h[ids_u_2, ids_u_1] = E_G2_s_hsub_vec_1_u_u[2, 1] + E_G2_s_hsub_vec_2_u_u[2, 1]
+
+            E_G_s_h[ids_u_2, ids_u_2] = E_G_s_hsub_vec_1_u_u[2, 2] + E_G_s_hsub_vec_2_u_u[2, 2]
+            E_G2_s_h[ids_u_2, ids_u_2] = E_G2_s_hsub_vec_1_u_u[2, 2] + E_G2_s_hsub_vec_2_u_u[2, 2]
```
jeff-regier: We're going to be increasing `Ia` to `4` soon. We need to include "quasar" and "off" modes in addition to "star" and "galaxy". How much of speedup do we get from unrolling the loop?

src/deterministic_vi/elbo_objective.jl
```diff
@@ -187,28 +187,28 @@ function calculate_source_pixel_brightness!{NumType <: Number}(
...
+                u_j = ids.u[j]
+                E_G_s_term = E_G_s_hsub_vec[1].u_u[i, j]
+                E_G2_s_term = E_G2_s_hsub_vec[1].u_u[i, j]
+                @inbounds for k in 2:Ia
+                    E_G_s_term += E_G_s_hsub_vec[k].u_u[i, j]
+                    E_G2_s_term += E_G2_s_hsub_vec[k].u_u[i, j]
+                end
+                E_G_s_h[u_i, u_j] = E_G_s_term
+                E_G2_s_h[u_i, u_j] = E_G2_s_term
+            end
```
jrevels: > We're going to be increasing Ia to 4 soon. We need to include "quasar" and "off" modes in addition to "star" and "galaxy". How much of speedup do we get from unrolling the loop?

@jeff-regier Is this correct for general `Ia > 1`? It doesn't cause any slowdown AFAICT.

jeff-regier:
@jrevels  The main bug is in `deterministic_vi/source_brightness.jl`. I didn't find the particular line, but I swapped files in and out until I'd isolated it. (You just have to do a find and replace setting `v[1]` to `v[]`.)

`source_brightness.jl` is not cpu intensive at all, so no need to prevent pointer lookups there. I'd propose substituting the old `source_brightness.jl` back in, with `v[1]` replaced with `v[]`.

jrevels:
> source_brightness.jl is not cpu intensive at all, so no need to prevent pointer lookups there. I'd propose substituting the old source_brightness.jl back in, with v[1] replaced with v[].

Done! Should I wait until #418 is merged to master and then rebase, or are you planning to #418 be merged after this gets merged?

jeff-regier:
I think PR #418 ended up being largely orthogonal to this one. The tests should pass on your pr without it, by rolling back `source_brightness.jl` and fixing the bugs in `Transform` I commented on. So let's merge this first. Will you figure out the extra memory allocations before merging this PR though?

src/deterministic_vi/elbo_objective.jl
jeff-regier: I think we won't know if it's correct until we do the math, but probably not. Neither "off" nor "quasar" are much like "galaxy" (`k = 2`). For this section of code, I'm prefer the way things were before this PR, if that doesn't matter much for performance.

jrevels: Ah, okay. I was following the comment above this code, I guess that's wrong then? I'll switch it back.

jeff-regier: I think the comment applies to the case where `Ia = 2`, not the future case where `Ia = 4`, but I'm a bit out of my depth here. Ryan masterminded the Hessian code. I'd just default to the former way of doing things here if the changes don't help performance.

jrevels:
It turns out that a lot of the extra allocations were due to an inference bug (ref JuliaLang/julia#15276). Fiddling around with type annotations fixes most of the additional allocations (but not all of them).

jeff-regier:
With `benchmark_rcf.jl`, now I see

```
 92.319823 seconds (102.73 M allocations: 8.677 GB, 5.06% gc time)
```

So the total size of the allocations is now lower than before the PR, but the number of allocations is still higher. The number of allocations seems to matter more for gc time in this case. Thanks for continuing to work on this PR....the last 10% of the programming always seems to take longest....

jeff-regier:
I merged this manually, to deal with all the conflicts. Memory allocation on the merged branch is really low now, for reasons I don't fully understand. With `benchmark_rcf.jl` and  one thread:
```
 54.310745 seconds (90.49 M allocations: 7.939 GB, 2.10% gc time)
```
The runtimes aren't comparable because this is on Cori rather than my vagrant install of Ubuntu, but the allocations show be.

jrevels:
Awesome, thanks for handling that!

> Memory allocation on the merged branch is really low now, for reasons I don't fully understand.

I'm not too surprised - a lot of the changes here really do decrease local memory usage significantly (demonstrable via profiling with `--track-allocation`). The fact that these local improvements didn't present an impact in the overall allocation count seemed to be due to buggy type inference and small-but-frequent allocations that hit this branch particularly hard for some reason. It's reasonable to me that changes to master enabled the benefits of this branch to "propagate" more easily by improving upstream/downstream code. Of course, this is just a guess...and Julia's type inference can be a finicky beast to make assumptions about.

PR #418, opened 2016-11-17
added a field named sources_to_visit to ElboArgs

jeff-regier:
@jrevels  -- I was trying to get to the bottom of why `test_active_pixels` fails in your PR. It turns out that that test never should have passed at all. This PR makes the test fail, and then fixes the failing test. But your PR may be breaking the test in a different way.

The only reason it was passing after my "no more tiles" PR was that every pixel in every image was active for both light sources in the unit test. I changed the test so that some pixels weren't active. It failed. It failed because in `elbo_likelihood`, I loop over only the sources listed in `active_sources`. That's fine for calculating derivatives and Hessians---only those pixels' terms have non-zero gradients wrt the sources were optimizing. However, the constant term (and thus `elbo_lik_2.v[]`) is off because we don't visit a lot of pixels. The test `test_active_pixels` relies on the elbo values matching up.

@rgiordan -- I fixed the issue by adding a new field to `ElboArgs` called `sources_to_visit`. It's a list of the sources that are relevant to calculating the elbo. I left the name `active_sources` unchanged, though I'd rather call it `sources_to_optimize` for symmetry with `sources_to_visit`.  Some day... The `sources_to_visit` controls what pixels are visited. It's the union of the active pixels for those sources. If `sources_to_visit == active_sources`, you get the old behavior, which is very efficient for optimizing because no extra pixels are visited.

@agnusmaximus -- I see you're debugging the Cyclades optimization now. You may want to confirm that Cyclades is finding a better (higher) value  of the objective function (i.e. the elbo) than single source inference. I'm not sure how you did this before: the collection of pixels in the elbo was changing based on what sources were listed as `active_sources`. Now, if after doing optimization, you set `sources_to_visit` to the complete list of sources you're optimizing (the `target sources`, not the whole catalog), that should compute the elbo correctly for either optimization technique.

### DIFF
--- a/src/Infer.jl
+++ b/src/Infer.jl
@@ -98,7 +98,7 @@ function infer_source(images::Vector{Image},
     cat_local = vcat(entry, neighbors)
     vp = Vector{Float64}[init_source(ce) for ce in cat_local]
     patches = get_sky_patches(images, cat_local)
-    ea = ElboArgs(images, vp, patches, [1])
+    ea = ElboArgs(images, vp, patches, [1], [1])
     load_active_pixels!(ea)
 
     f_evals, max_f, max_x, nm_result = maximize_f(elbo, ea)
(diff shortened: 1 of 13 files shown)


agnusmaximus:
Ah, I finally think I understand this.  Previously I was just making the maximize_f return the score achieved after doing the newton updates every iteration, then taking the average of the sum of scores achieved across all the target light sources. Does this approach work as well? Otherwise I'll modify the code to make another call to maximize_f with active_sources=target_sources, sources_to_visit=target_sources passed in.

jeff-regier:
That sum over sources will only be correct when the sources don't overlap. The objective value is a sum over images and and a sum over pixels, but the sum over sources can't be pulled out front unless the sources don't overlap. Sorry, I think I said the wrong thing about this before. For each Newton step, you should keep setting `active_sources = [1]` and `sources_to_visit = [1]`. But to evaluate the objective to see how well you did, set `sources_to_visit` to the sources you optimized and set `active_sources = []`. I mean, you could set `active_sources` to anything and it shouldn't change the value of the objective. It just changes what derivatives to compute.

PR #419, opened 2016-11-18
PSF Version of the ELBO

rgiordan:
This seems to be working, though remains a bit different from the current Celeste.  It is also currently much slower, though there is vast scope for optimization.

I would like to check this in so that it tracks changes made in the fundamental organization of Celeste to keep from clunky merges.  The FFT-based ELBO itself is not tested, though that is appropriate, since it's still pretty preliminary.

I would also like to check in some simple image rendering functions to make sure that they, too stay updated as the codebase changes.


benchmark/galsim/GalsimBenchmark.jl
```diff
@@ -216,6 +216,8 @@ function make_catalog_entries(header::FITSIO.FITSHeader)
     catalog_entries
 end
 
+
+# TODO: this code seems to be obsolete?
```
jeff-regier, line 220: I think @gostevehoward removed it in his latest PR but that PR hasn't been merged yet.

bin/render_source.jl
```diff
@@ -0,0 +1,55 @@
+using Celeste:
```
jeff-regier, line 1: I'd be better to have most of this in `src`, and make the `bin` script a thin wrapper around it. Not something that needs to be done now, but maybe open an issue? I might get to it if I start using this code to debug the `e_scale` issue.

src/CelesteEDA.jl
```diff
@@ -0,0 +1,184 @@
...
+
+        if !p.active_pixel_bitmap[h2, w2]
+            continue
+        end
+
+        clear!(ea.elbo_vars.E_G, false)
+        clear!(ea.elbo_vars.var_G, false)
+
+        populate_fsm_vecs!(
+                                 ea.elbo_vars.bvn_derivs,
```
jeff-regier: Spacing looks off. Tabs accidentally?

bin/render_source.jl
```diff
@@ -0,0 +1,55 @@
...
+
+
+###########################
+# Visualize a source at initial parameters using both classic and FFT Celeste.
+
+ea = ElboArgs(images, deepcopy(vp), patches, [1]);
+Celeste.Infer.load_active_pixels!(ea);
+
+ea_fft = ElboArgs(images, deepcopy(vp), patches, [1], psf_K=1);
+Celeste.Infer.load_active_pixels!(ea_fft, exclude_nan=false);
```
jeff-regier, line 39: Why are `NaN` values included?

src/CelesteEDA.jl
```diff
@@ -0,0 +1,184 @@
...
+import ..SensitiveFloats.clear!
+import ..DeterministicVIImagePSF:
+    FSMSensitiveFloatMatrices
+using ..DeterministicVIImagePSF:
+    clear_brightness!, populate_star_fsm_image!, populate_gal_fsm_image!,
+    accumulate_source_image_brightness!, load_gal_bvn_mixtures
+
+import Base.print
+function print(ce::CatalogEntry)
+  for field in fieldnames(ce)
```
jeff-regier: 2 spaces -> 4 spaces

src/DeterministicVIImagePSF.jl
```diff
@@ -0,0 +1,33 @@
...
+
+import ..SensitiveFloats:
+    SensitiveFloat, zero_sensitive_float, zero_sensitive_float_array,
+    multiply_sfs!, add_scaled_sfs!, clear!
+
+include("deterministic_vi_image_psf/sensitive_float_fft.jl")
+include("deterministic_vi_image_psf/lanczos.jl")
+include("deterministic_vi_image_psf/fsm_matrices.jl")
+include("deterministic_vi_image_psf/elbo_image_psf.jl")
+
```
jeff-regier, line 30: Could these files get included in `DeterministicVI` rather than starting a new module? It looks like you're already importing a lot for `DeterministicVI`.

jeff-regier: Also, do some of these includes belong in `Model`? We'd want @andymiller to have access to the new PSF model too.

src/deterministic_vi/elbo_objective.jl
```diff
@@ -1,4 +1,101 @@
...
+                    ea::ElboArgs{NumType},
+                    b::Int;
+                    calculate_derivs::Bool=true,
+                    calculate_hessian::Bool=true)
+    # call bvn loader from the Model Module
+    Model.load_bvn_mixtures(ea.S, ea.patches, ea.vp, ea.active_sources,
+                            ea.psf_K, b,
+                            calculate_derivs=calculate_derivs,
+                            calculate_hessian=calculate_hessian)
+end
```
jeff-regier: IMO these functions make the code less readable, adding almost 100 lines of code, while giving really minimal benefit to someone already very familiar with the code. It adds a whole extra layer of indirection they have to come to understand.  If these functions being wrapped have too many arguments (they do), let me work on that. In the meantime, please just call the Model functions directly. Let's not have a whole additional layer of function calls.

src/deterministic_vi/elbo_objective.jl
```diff
@@ -188,6 +285,25 @@ function calculate_source_pixel_brightness!{NumType <: Number}(
...
+        elbo_vars,
+        ea,
+        elbo_vars.E_G_s,
+        elbo_vars.var_G_s,
+        elbo_vars.fs0m_vec[s],
+        elbo_vars.fs1m_vec[s],
+        sbs[s],
+        b, s,
+        s in ea.active_sources)
+end
```
jeff-regier: I find overloading function names work well for small numbers of arguments, e.g. 2 ints or 2 floats. But that it's confusing to have the same name for functions with very different signatures. I'd rather have `wrapper` in the name of the wrapper function, or something like that that. I've also seen `do_` used as a prefix for the wrapped function. But I'd much rather get rid of the wrappers entirely. They're really not worth it, even if they making for slightly few keystrokes: code is read much more often than written.

src/model/fsm_util.jl
```diff
@@ -61,22 +61,22 @@ function load_bvn_mixtures{NumType <: Number}(
...
+                 for k = 1:psf_K
+                     gal_mcs[k, j, i, s] = GalaxyCacheComponent(
+                         e_dev_dir, e_dev_i, galaxy_prototypes[i][j], psf[k],
+                         m_pos,
+                         sp[lidx.e_axis], sp[lidx.e_angle], sp[lidx.e_scale],
+                         calculate_derivs && (s in active_sources),
+                         calculate_hessian)
+                 end
+             end
+         end
```
jeff-regier: You added a space to all these lines, but wasn't the spacing right before?

src/model/fsm_util.jl
```diff
@@ -195,7 +180,7 @@ function populate_fsm!{NumType <: Number}(
                             mv_calculate_derivs,
                             mv_calculate_hessian,
                             gal_mcs[k, j, i, s], x, wcs_jacobian,
-                            is_active_source)
+                            active_source)
```
jeff-regier: Let's use `is_active_source` here to indicate that it's a boolean value, rather than an element of `active_sources` (an array of ints, not booleans).

src/deterministic_vi_image_psf/fsm_matrices.jl
```diff
@@ -0,0 +1,175 @@
...
+        initialize_fsm_sf_matrices_band!(
+            fsm_vec[b], b, num_active_sources,
+            h_lower_vec[b], w_lower_vec[b],
+            h_upper_vec[b], w_upper_vec[b],
+            psf_image_mat)
+    end
+end
+
+
+# This is useful for debugging and exploration.  Do not delete.
```
jeff-regier: Please give this function a name like `debug_populate_fsm_vec!` so it isn't confused with the real `populate_fsm_vec!`. Also please give a date for deleting this in the comment, e.g `Do not delete before 2/2017`, so that we're not maintaining unreachable code indefinitely

jeff-regier (changes requested):
It's exciting to have the new PSF included! David is excited about the PSF model too, btw, I meant to mention that earlier. The revisions I suggest are all pretty quick, I think. If any are more involved, feel free to open issues for them rather than delaying this merge.

bin/render_source.jl
rgiordan: Yes, there are some similar functions in ParallelRun.  Seems to me that this kind of thing belongs in SDSSIO.  Done.

bin/render_source.jl
rgiordan: The way the PSF fitting code works for galaxies is that it first renders the object in the active pixels and then convolves with the PSF using a FFT.  If the active pixels have a big hole in the middle (as is usually the case for bright objects), then the pre-convolution object is hardly rendered at all.

The right solution is to have a separate set of active pixels for pre- and post-convolution rendering.  I'll leave it to you to decide how to handle that, though.

src/CelesteEDA.jl
rgiordan: Man, we really need a linter.

src/CelesteEDA.jl
rgiordan: Done.

src/DeterministicVIImagePSF.jl
rgiordan: Let's keep this siloed for now while I'm working on it.

src/model/fsm_util.jl
rgiordan: Done.

src/deterministic_vi_image_psf/fsm_matrices.jl
rgiordan: Done.

src/deterministic_vi/elbo_objective.jl
rgiordan: I think these came in accidentally from an old version of the file in the merge since I wasn't using them.  So I'll delete them.

I disagree, but more importantly, I think it would be useful to have a set of coding standards that we can all write to.  Right now we are oscillating between one style and another.  These date to the end of the summer, IIRC, when you presumably had a different opinion.

jeff-regier (approved):

jeff-regier:
Looks good, thanks for making all the revisions.

PR #423, opened 2016-11-19
Estimate box runtime

### LINKED ISSUE #421: count active pixels in an arbitrary bounding box

jeff-regier:
This PR implements a new script, `bin/estimate-box-runtime.jl`, that counts the number of active pixels in the specified bounding box. It closes #421. Pixels that are active for 2 sources are double counted, by design. Pixels that are active for a source that won't be optimized are excluded from the tally.

Here are some examples.

This is from a region where the stripes are pretty spread out, near the middle of the arcs. (See the [SDSS coverage map](http://classic.sdss.org/dr7/coverage/).)
```
$ ./estimate-box-runtime.jl 200 200.5 40 40.5
The specified region contains 3,253,336 active pixels.
```

Less spread out, but still near the center.
```
$ ./estimate-box-runtime.jl 150 150.5 40 40.5
The specified region contains 7,119,292 active pixels.
```

Near one of the "poles".
```
$ ./estimate-box-runtime.jl 240 240.5 30 30.5
The specified region contains 14,198,395 active pixels.
```

From stripe 82:
```
$ ./estimate-box-runtime.jl 0 0.5 0 0.5
The specified region contains 53,177,432 active pixels.
```

------
Some changes to the code base included with the script:
- To estimate box runtime without creating an `ElboArgs` instance, or doing anything specific to one particular time of approximate inference, this PR changes `load_active_pixels!` to take `images` and `patches` as its arguments.
- The PR moves `infer_source` from `Infer` to `DeterministicVI` so that everything in `Infer` can be used by `MCMC` just as it is used by `DeterministicVI`. `Infer` is now loaded in `Celeste.jl` immediately after `Model`, because any inference routine is loaded.
- The PR splits `one_node_infer` into several smaller functions, but preserves the original inference to `one_node_infer`.

### DIFF
--- a/bin/estimate-box-runtime.jl
+++ b/bin/estimate-box-runtime.jl
@@ -1,0 +1,21 @@
+#!/usr/bin/env julia
+
+import Celeste.ParallelRun: BoundingBox, estimate_box_runtime
+
+
+const usage_info =
+"""
+Usage:
+  estimate-box-runtime.jl <ramin> <ramax> <decmin> <decmax>
+"""
+
+const stagedir = ENV["CELESTE_STAGE_DIR"]
+
+if length(ARGS) != 4
+    println(usage_info)
+else
+    box = BoundingBox(ARGS...)
+    num_active = estimate_box_runtime(box, stagedir)
+    println("The specified region contains $num_active active pixels.")
+end
+
--- a/bin/estimate-box-runtime.jl
+++ b/bin/estimate-box-runtime.jl
@@ -16,6 +16,7 @@ if length(ARGS) != 4
 else
     box = BoundingBox(ARGS...)
     num_active = estimate_box_runtime(box, stagedir)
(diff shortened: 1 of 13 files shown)


jeff-regier:
@kpamnany @rgiordan  I ran `infer-box.jl` on the those 3 regions, using 8 threads for each. The first region, with 3.2 million active pixels, finished in 10 minutes, 22 second (13% gc time). The second region, with 7.2 million active pixels, finished in 21 minutes, 14 seconds (14.6% gc time). Third third region, with 14.2 million active pixels, finished in 54 minutes (16.8% gc time).

So runtime correlates highly with the number of active pixels, except to the extent that GC time grows with job duration. We could probably bring all these runtimes down before GB by 5x or 10x with 1) more threads (and fewer memory allocations), 2) the new PSF, 3) more careful selection of active pixels.

PR #427, opened 2016-11-22
Cyclades fix objective value comparison

agnusmaximus:
Call maximize_f again after comparison to get objective value; remove returning sum of obj values from one_node_infer and similar functions

### DIFF
--- a/test/test_joint_infer.jl
+++ b/test/test_joint_infer.jl
@@ -15,13 +15,8 @@ function compute_obj_value(results,
 
     ea = ElboArgs(images, [x["vs"] for x in results],
                   Infer.get_sky_patches(images, catalog[target_sources]),
-                  collect(1:length(target_sources)), [1])
-    iter_count, obj_value, max_x, r = DeterministicVI.maximize_f(
-        DeterministicVI.elbo,
-        ea,
-        max_iters=1)
-
-    obj_value
+                  collect(1:length(target_sources)), Int64[])
+    DeterministicVI.elbo(ea, calculate_derivs=false, calculate_hessian=false).v[]
 end
 
 """
--- a/test/test_joint_infer.jl
+++ b/test/test_joint_infer.jl
@@ -13,37 +13,9 @@ function compute_obj_value(results,
                                                                            box=box, objid="",
                                                                            target_rcfs=RunCamcolField[])
 
-    # Create a map from entry.thing_id -> variational params of the source.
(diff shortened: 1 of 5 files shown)


test/test_joint_infer.jl
```diff
@@ -1,5 +1,28 @@
...
+                           box=BoundingBox(-1000.,1000.,-1000.,1000.))
+    catalog, target_sources, neighbor_map, images = ParallelRun.infer_init(rcfs, stagedir;
+                                                                           box=box, objid="",
+                                                                           target_rcfs=RunCamcolField[])
+
+    ea = ElboArgs(images, [x["vs"] for x in results],
+                  Infer.get_sky_patches(images, catalog[target_sources]),
+                  collect(1:length(target_sources)), [1])
+    iter_count, obj_value, max_x, r = DeterministicVI.maximize_f(
+        DeterministicVI.elbo,
```
jeff-regier: Could you call `DeterministicVI.elbo` directly here to get the objective's value?

test/test_joint_infer.jl
```diff
@@ -1,5 +1,28 @@
...
+                           rcfs::Vector{RunCamcolField},
+                           stagedir::String;
+                           box=BoundingBox(-1000.,1000.,-1000.,1000.))
+    catalog, target_sources, neighbor_map, images = ParallelRun.infer_init(rcfs, stagedir;
+                                                                           box=box, objid="",
+                                                                           target_rcfs=RunCamcolField[])
+
+    ea = ElboArgs(images, [x["vs"] for x in results],
+                  Infer.get_sky_patches(images, catalog[target_sources]),
+                  collect(1:length(target_sources)), [1])
```
jeff-regier: The last two arguments may be in the wrong order. You'd all the light sources as sources to visit, I think. Does it work to use `Int64[]` as `active_sources` rather than `[1]`?

jeff-regier (changes requested):

test/test_joint_infer.jl
agnusmaximus: Great! I missed that you could do that. Thanks!

test/test_joint_infer.jl
agnusmaximus: Ah yeah, I tried [] which didn't work, but Int64[] worked.

jeff-regier (approved):
Great! Also, a relief to see that the joint inference unit tests still pass with the real objective function :-)

PR #428, opened 2016-11-23
Add try catch in joint infer parallel function; use newton steps=20 o…

agnusmaximus:
…n first iter, and new steps=1 afterwards

### DIFF
--- a/src/cyclades.jl
+++ b/src/cyclades.jl
@@ -291,7 +291,11 @@ function one_node_joint_infer(catalog, target_sources, neighbor_map, images;
                     max_iters=n_newton_steps)
             end
         catch ex
-            Log.error(string(ex))
+            if is_production_run || nthreads() > 1
+                Log.error(string(ex))
+            else
+                rethrow(ex)
+            end
         end
     end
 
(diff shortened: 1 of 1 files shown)


src/cyclades.jl
```diff
@@ -280,13 +280,18 @@ function one_node_joint_infer(catalog, target_sources, neighbor_map, images;
...
+            n_newton_steps = iter == 1 ? 20 : 1
+            for cur_source_indx in source_assignment
+                cur_entry = catalog[target_sources[cur_source_indx]]
+                iter_count, obj_value, max_x, r = DeterministicVI.maximize_f(
+                    DeterministicVI.elbo,
+                    model[cur_source_indx],
+                    max_iters=n_newton_steps)
+            end
+        catch ex
+            Log.error(string(ex))
```
jeff-regier: Perhaps better to duplicate the code from ParallelRun.jl in this catch block:
```
                    if is_production_run || nthreads() > 1
                        Log.error(string(ex))
                    else
                        rethrow(ex)
                    end
```
We generally want to halt and see full stack traces on errors, except in production runs (indicated by an environment variable) or when we're using multiple threads, and the stack trace wouldn't get displayed anyway.

Issue #293, opened 2016-09-15
accuracy benchmarks based on galsim data


jeff-regier:

gostevehoward:
Here's where I'm at right now.

My first pass at the GalSim script was generated a grid of random images. My thinking now is to define a list of benchmark cases each as a set of parameters to GalSim, mostly corresponding to parameters inferred by Celeste. So, for example, a CSV like [this](https://github.com/gostevehoward/Celeste.jl/blob/galsim_benchmark/benchmark/galsim/galsim_truth.csv), in which we start with a tiny, noiseless, bright galaxy; then, one at a time, we add rotation and ellipticity, varying size, varying flux, etc; then we try all those transformations together, then add noise. The [generate_test_image.py](https://github.com/gostevehoward/Celeste.jl/blob/galsim_benchmark/benchmark/galsim/generate_test_image.py) script reads this CSV and writes a series of FITS images, one for each test case (I know there's a way to put multiple images in a single FITS file, but is it worth the complexity?). Here's an idea of what those images look like, right now:

![Sample test images](https://cloud.githubusercontent.com/assets/1097616/18592205/a904f6da-7bea-11e6-94d6-c48b22d198b0.png)

The [benchmark script](https://github.com/gostevehoward/Celeste.jl/blob/galsim_benchmark/benchmark/galsim/GalsimBenchmark.jl) reads the same CSV, reads in each corresponding image and runs Celeste on it. In theory this can report the inferred and true values side-by-side, or in the future just report some overall accuracy metric and flag if anything's way off.

However, I keep running into [that issue where the optimization crashes](https://github.com/jeff-regier/Celeste.jl/issues/292#issuecomment-247197908). Ryan helpfully debugged the first one and gave me lots of tips (in that issue's comments thread) on how to debug them myself. So I think now is the time for me to learn more about how the optimization works, what all these parameters mean, and how to debug such issues myself.

rcthomas:
Writing multi-extension FITS files shouldn't be that complex:

http://docs.astropy.org/en/stable/generated/examples/io/create-mef.html#sphx-glr-generated-examples-io-create-mef-py

Galsim is just using pyfits or astropy.io.fits to do this.  It may do it for you already.  But the nice thing would be a single csv would go with a single FITS file at least.

gostevehoward:
[...]

PR #434, opened 2016-11-26
Change GalSim WCS so that world coords are in degrees

gostevehoward:
This matches the SDSS WCS and makes the hard-coded default `loc_width` work.

I've also changed things to account for the fact that `e_scale` is in pixel coords.


jeff-regier:
Looks really good, Steve. I like those `@galsim_test_case` decorators.

Does this PR close #293?

gostevehoward:
Sure, I closed the issue.

gostevehoward:
I uploaded the latest FITS image to portal.nersc.

PR #435, opened 2016-11-27
Joint infer galsim

agnusmaximus:
Sorry for the lateness, I've included joint_infer into the Galsim benchmarks. Running the galsim benchmark runs both the single infer_source method and also the joint infer method (with 20 iterations). Then it creates the table with the expected/infer_source/joint_infer values side by side. 

If it's better to pass in a flag to determine whether to run joint_infer vs infer_source, definitely let me know and I'll change that and update this PR.

Here are some results that I got from the benchmark:
http://pastebin.com/pe1JdEZB

I'm not sure why the non-overlapping test cases don't do so well both for infer_source and joint_infer. But on the overlapping test cases joint_infer seems to do relatively better.

Definitely let me know what you think! @jeff-regier @gostevehoward

### DIFF
--- a/benchmark/galsim/GalsimBenchmark.jl
+++ b/benchmark/galsim/GalsimBenchmark.jl
@@ -4,7 +4,7 @@ import FITSIO
 import StaticArrays
 import WCS
 
-import Celeste: Model, DeterministicVI, ParallelRun
+import Celeste: Model, DeterministicVI, ParallelRun, Infer
 
 const FILENAME = "output/galsim_test_images.fits"
 
@@ -231,15 +231,16 @@ function parallel_inference(band_images, catalog_entries; joint_infer=false)
     target_sources = collect(1:length(catalog_entries))
 
     # Create the neighbor map (everyone is a neighbor of each other)
-    #neighbor_map = Infer.find_neighbors(target_sources, catalog_entries, band_images)
-    neighbor_map = Vector{Int64}[Int64[] for s in target_sources]
-    for target_source in target_sources
-        for other_target_source in target_sources
-            if target_source != other_target_source
-                push!(neighbor_map[target_source], other_target_source)
-            end
-        end
-    end
+    neighbor_map = Infer.find_neighbors(target_sources, catalog_entries, band_images)
+    #neighbor_map = Vector{Int64}[Int64[] for s in target_sources]
+    #for target_source in target_sources
+    #    for other_target_source in target_sources
+    #        if target_source != other_target_source
+    #            push!(neighbor_map[target_source], other_target_source)
+    #        end
+    #    end
+    #end
+    println("$(neighbor_map)")
 
     # Optimize
     results = ParallelRun.parallel_infer(catalog_entries, target_sources, neighbor_map, band_images;
--- a/benchmark/galsim/GalsimBenchmark.jl
+++ b/benchmark/galsim/GalsimBenchmark.jl
@@ -131,7 +131,7 @@ function get_field(header::FITSIO.FITSHeader, key::String)
     end
 end
 
-function actual_values(params)
+function actual_values(ids, star_galaxy_index, params)
     Float64[
         params[ids.u[1]],
         params[ids.u[2]],
@@ -166,8 +166,8 @@ function benchmark_comparison_data(single_infer_params, joint_infer_params, head
             get_field(header, "CL_C45_1"),
             header["CL_TYPE1"] == "star" ? 0 : 1,
         ],
-        single_infer_actual=actual_values(single_infer_params),
-        joint_infer_actual=actual_values(joint_infer_params)
+        single_infer_actual=actual_values(ids, star_galaxy_index, single_infer_params),
+        joint_infer_actual=actual_values(ids, star_galaxy_index, joint_infer_params)
     )
 end
 
--- a/benchmark/galsim/GalsimBenchmark.jl
+++ b/benchmark/galsim/GalsimBenchmark.jl
@@ -232,15 +232,6 @@ function parallel_inference(band_images, catalog_entries; joint_infer=false)
 
     # Create the neighbor map (everyone is a neighbor of each other)
     neighbor_map = Infer.find_neighbors(target_sources, catalog_entries, band_images)
-    #neighbor_map = Vector{Int64}[Int64[] for s in target_sources]
-    #for target_source in target_sources
-    #    for other_target_source in target_sources
-    #        if target_source != other_target_source
-    #            push!(neighbor_map[target_source], other_target_source)
-    #        end
-    #    end
-    #end
-    println("$(neighbor_map)")
 
     # Optimize
     results = ParallelRun.parallel_infer(catalog_entries, target_sources, neighbor_map, band_images;
--- a/benchmark/galsim/GalsimBenchmark.jl
+++ b/benchmark/galsim/GalsimBenchmark.jl
@@ -147,7 +147,7 @@ function actual_values(params)
     ]
 end
 
-function benchmark_comparison_data(single_params, joint_infer_params, header)
+function benchmark_comparison_data(single_infer_params, joint_infer_params, header)
     ids = Model.ids
     star_galaxy_index = header["CL_TYPE1"] == "star" ? 1 : 2
     DataFrame(
--- a/benchmark/galsim/GalsimBenchmark.jl
+++ b/benchmark/galsim/GalsimBenchmark.jl
@@ -131,6 +131,22 @@ function get_field(header::FITSIO.FITSHeader, key::String)
     end
 end
 
+function actual_values(params)
+    Float64[
+        params[ids.u[1]],
+        params[ids.u[2]],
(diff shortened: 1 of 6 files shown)


agnusmaximus:
Edit: I'm currently looking into find_neighbors, which last time I checked was doing something weird. Currently I'm setting everyone to be each others neighbors, so joint_infer (in the galsim test cases) is pretty much serial right now.

agnusmaximus:
Hmm just to confirm, the dist function used in Infer.jl (https://github.com/jeff-regier/Celeste.jl/blob/master/src/Infer.jl#L54) is to approximate angular distance right? (https://en.wikipedia.org/wiki/Angular_distance)

If so, is there any reason not to use the above? ^ 

I haven't really figured out the problem yet, but there seems to be a couple of issues
1) https://github.com/jeff-regier/Celeste.jl/blob/master/src/Infer.jl#L69 - `s` should be swapped with `ts`, since `s` is the actual source id rather than the index into target sources. Edit: (nvm, I think this is ok, sorry about that!)
2) https://github.com/jeff-regier/Celeste.jl/blob/master/src/Infer.jl#L55 - `dec1` is in degrees rather than in radians, and I think cos takes in radians.

I've tried implementing https://en.wikipedia.org/wiki/Angular_distance into find_neighbors and it seems to work, but I'm definitely not familiar with this angular calculation, so any input would be awesome.

Edit: Yeah... I don't think the angular distance thing I implemented works....

jeff-regier:
Those scores for overlapping astronomical objects look really good!

The X and Y centers don't seem to match the truth at all, for either single of joint, for any galsim tests except the overlapping ones, e.g.
```
│ 1   │ "simple_star"                 │ "X center (world coords)" │ 36.375   │ 0.00533401          │ 0.00235311         │
```
 A bug? When I run Steve's `run_galsim_benchmark.jl` script, I see that `expected` is `0.005335` rather than `36.375`. 

Interesting finding about the degrees vs radians....potentially a pretty serious bug. Within Celeste, I believe we're using radians everywhere. At least here we're converting the existing catalogs, in degrees, to radians:
https://github.com/jeff-regier/Celeste.jl/blob/master/src/SDSSIO.jl#L567

Is it different in the `galsim` code?

jeff-regier:
The distance formula looks OK. Maybe just encode the `CatalogEntry` instances in radians rather than degrees? Or we could change Celeste to use degrees everywhere, and use `cosd` and `sind` rather than `cos` and `sin`, since astronomers don't seem to use radians. That's more likely to break something though, and the angle code can be really tedious. Thanks much for working on it.

agnusmaximus:
Yeah, encoding the CatalogEntry in radians seems to be the best way. 

Also, how are you running run_galsim_benchmark.jl? When I run it (on the master branch), for the simple_star test case I keep getting an expected of 36.375.  And I have to add a 'include("GalsimBenchmark.jl")' at the top of run_galsim_benchmark for things to run

jeff-regier:
That's what I'm doing too.  Once @gostevehoward moves that module to the `src` directory for the unit tests you shouldn't need to include it any more.

Are you using the latest version of steve's galsim code? Steve made a PR today and I merged it. Not sure if that's the reason we're seeing different output.

agnusmaximus:
Hmm yeah I think I have the latest version of the galsim code, I'm doing a "git fetch && git reset --hard origin/master" before doing the runs.

gostevehoward:
You may need to rebuild or redownload the GalSim test image FITS file.

gostevehoward:
To elaborate (I should put this in a README later; you may already know
this, but it's good context anyhow):

The most natural coordinate system in an image is, obviously, pixel
coordinates. A FITS header includes a description of a world coordinate
system, which (in our case) is just a 2-D (invertible) affine
transformation to convert between pixel coordinates and "world
coordinates". Nothing in the FITS file specifies the units of these world
coordinates. It's just an affine transformation; the units of world
coordinates are simply a matter of interpretation. You could make a FITS
file with 0.396 world coordinates per pixel coordinate
<http://classic.sdss.org/dr7/>, and interpret world coordinates as Dec/RA
in arcseconds. Or, you could multiply that transformation by 1/3600 and
interpret the world coordinates for the resulting FITS file as Dec/RA in
degrees. Of course, you could just as well interpret the latter FITS file
as having world coordinates in arcseconds, but this would lead to
unrealistic interpretation---the image would be interpreted as having
extremely high resolution and imaging extremely tiny light sources.

Almost everything in Celeste either works either in pixel coordinates or
transforms through the WCS. Hence you should be able to pass an image with
whatever WCS and all the calculations adapt accordingly. However, there is
one hard-coded constant in world coordinate units (maybe others too?): the
default value of `loc_width`, the maximum radius from the initial position
of a light source within which Celeste will search for the true position
during inference. This value is 0.0015 world coordinate units
<https://github.com/jeff-regier/Celeste.jl/blob/master/src/deterministic_vi/maximize_elbo.jl#L112>.
That's a reasonable value if world coordinates are in degrees, and a
ridiculously small value if world coordinates are in pixels, for our images
at least.

So the change I sent yesterday changes the GalSim image generation script
to generate a WCS which is reasonably interpreted as putting world coords
in degrees rather than arcseconds. Then the default value of `loc_width` is
sensible. However, there is one other hard-coded constant in
world-coordinates at the moment: the default initial position for the
GalSim benchmark script
<https://github.com/jeff-regier/Celeste.jl/blob/master/benchmark/galsim/GalsimBenchmark.jl#L205>.
This is just a temporary lazy hack which (clearly) I should have done right
in the first place. It only uses this default initialization for
single-source images, in which the true light source position is always
close to the center, and we want to test that Celeste can find the correct
position when it's initialized slightly off. For multi-source images, it
may be that none of the sources are near the center, so in this case we
initialize Celeste with the true position, read out of the FITS header.

That's why when you run the current code on the old FITS file, it's
complete nonsense for single source images---the initial position (~5e-3)
is way, way off from the true position (~36), and `loc_width` prevents
Celeste from possibly correcting this. So inference is bogus. But for
multi-source images, Celeste starts with the true position.

Let me know if this is helpful/confusing and please correct me if you see
any mistakes or misunderstandings here!

agnusmaximus:
Awesome!! That worked @gostevehoward! Sorry about this confusion... I'll post the updated results in a second

agnusmaximus:
Pulling the FITS data file fixed everything, including the find_neighbors. 

Here are the results: http://pastebin.com/MSvrtrf4

jeff-regier:
These numbers look pretty good. Because our galaxy model needs work, it's not so easy to compare joint and single on synthetic data with overlapping galaxies. The overlapping stars test cases gives pretty good validation that joint inference is working:
```
│ 97  │ "overlapping_stars"           │ "Brightness (nMgy)"          │ 40         │ 36.9041             │ 39.9526            │
```
So nice work! Seems like this PR closes #405. It's also a good sign that joint inference matches the performance of single inference on all the test cases that have just one light source.

jeff-regier:
This line is really good too!
```
│ 102 │ "overlapping_galaxies"        │ "Angle (degrees)"            │ 135        │ 47.1564             │ 134.948            │
```
Joint inference gets the angle almost perfect. Single inference must be modeling parts of multiple galaxies with a single galaxy. That's great to see.

benchmark/galsim/GalsimBenchmark.jl
```diff
@@ -240,10 +265,13 @@ function main(; test_case_name=Nullable{String}())
         variational_parameters = DeterministicVI.infer_source(
```
jeff-regier: Rather than calling `infer_source`, would you call your `parallel_inference` routine here with `joint_infer = false`? That would be more symmetric.

benchmark/galsim/GalsimBenchmark.jl
```diff
@@ -164,6 +163,19 @@ function benchmark_comparison_data(params, header)
...
+            joint_infer_params[ids.u[2]],
+            joint_infer_params[ids.e_axis],
+            canonical_angle(joint_infer_params) * 180 / pi,
+            joint_infer_params[ids.e_scale],
+            exp(joint_infer_params[ids.r1[star_galaxy_index]]),
+            exp(joint_infer_params[ids.c1[1, star_galaxy_index]]),
+            exp(joint_infer_params[ids.c1[2, star_galaxy_index]]),
+            exp(joint_infer_params[ids.c1[3, star_galaxy_index]]),
+            exp(joint_infer_params[ids.c1[4, star_galaxy_index]]),
+            joint_infer_params[ids.a[2]],
```
jeff-regier: Looks like some duplicated code here. Would put the common code in a function called something like `vs_to_actual` and call it twice? Something like
```
single_infer_actual = vs_to_actual(params)
joint_infer_actual = vs_to_actual(joint_infer_params)
```
And maybe rename `params` to `single_ infer_params` for symmetry.

benchmark/galsim/GalsimBenchmark.jl
agnusmaximus: sounds good!

PR #436, opened 2016-11-27
Memory-efficient SensitiveFloats

jeff-regier:
`SensitiveFloat` instances always allocate memory for gradients and Hessians, regardless of whether gradients and Hessians are being computed. This PR makes those allocations optional, and adds two new booleans to SensitiveFloat: `has_gradient` and `has_hessian`. 

Now that each `SensitiveFloat` records whether it is being used in a gradient or a hessian calculation, it isn't as necessary to pass those `calculate_derivs` and `calculate_hessian` flags around all over the place. For example, to set a every of field of a `SensitiveFloat` to `0`, you'd call `clear!(sf)`, not `clear!(sf, calculate_hessian)`. The new `clear!` method always zeros out the SensitiveFloat's value, the gradient (if it has one), and the Hessian (if it has one). There's no way to mess up and 1) forget to clear a field that is in use, 2) spend cycles clearing a field that isn't used. The latter was happening a lot. 

As a sanity check, with the PR, `benchmark_elbo.jl` shows a 30% drop in memory use, but a 40% drop in runtime. On master:
```
  0.389335 seconds (414.96 k allocations: 17.611 MB, 1.59% gc time)
```
And on this branch:
```
  0.238657 seconds (360.80 k allocations: 11.466 MB)
```

For a version of `benchmark_infer.jl` that doesn't measure the data loading parts, I see a 10% drop in runtime and 20% drop in memory use and memory allocations.
```
  2.140580 seconds (1.52 M allocations: 86.873 MB, 0.59% gc time)
```
vs
```
  1.864795 seconds (1.26 M allocations: 72.500 MB, 0.66% gc time)
```

A few other changes

- I'm using the term `gradient` everywhere now rather than `derivs` because there were places in the code where the latter was being used to refer to both the gradient and the Hessian, and then some places where `derivs` was used to refer only the gradient. The new terminology avoids this ambiguity.
- Newly initialized SensitiveFloats have a value of 0.
- SensitiveFloats no longer are parameterized by a `ParamType`. Doing so causes a new version of every method to be compiles for each `ParamType` it is called with. All a `SensitiveFloat` needs to know if the length of the ParamType, e.g. 32 for `CanonicalParams`. This change also makes `SensitiveFloat` clearer conceptually, because it isn't specific to Celeste's types of parameter vectors.

### DIFF
--- a/src/SensitiveFloats.jl
+++ b/src/SensitiveFloats.jl
@@ -1,7 +1,6 @@
 module SensitiveFloats
 
 export SensitiveFloat,
-       zero_sensitive_float,
        clear!,
        multiply_sfs!,
        add_scaled_sfs!,
@@ -9,8 +8,6 @@ export SensitiveFloat,
        add_sources_sf!,
        set_hess!
 
-
-abstract ParamSet
 
 """
 A function value and its derivative with respect to its arguments.
@@ -24,7 +21,7 @@ Attributes:
       in the same format as d.  This is used for the full Hessian
       with respect to all the sources.
 """
-immutable SensitiveFloat{ParamType <: ParamSet, NumType <: Number}
+immutable SensitiveFloat{NumType}
     v::Base.RefValue{NumType}
 
     # local_P x local_S matrix of gradients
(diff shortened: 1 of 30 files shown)


src/SensitiveFloats.jl
```diff
@@ -154,76 +128,61 @@ each evaluated at (sf1, sf2).
...
+                        sf1::SensitiveFloat{T1},
+                        sf2::SensitiveFloat{T1},
+                        sf_result::SensitiveFloat{T1},
+                        v::T1,
+                        g_d::Vector{T2},
+                        g_h::Matrix{T3})
     # You have to do this in the right order to not overwrite needed terms.
-    if calculate_hessian
+    if sf_result.has_hessian
         combine_sfs_hessian!(sf1, sf2, sf_result, g_d, g_h)
```
rgiordan, line 140: The reason it was called ```calculate_derivs``` not ```calculate_gradient``` was because of the many places like this in the code where it is not possible to calculate a hessian without having gradients.

You might worry that if the user specifies ```calculate_hessian=true``` but ```calculate_gradient=false```, perhaps expecting to get Hessians but not gradients, strange things might happen.  I'm not sure I'll be able to catch all those spots.  I guess I'll leave it up to your judgement that the new nomenclature is actually clearer.  It wasn't trivial to make sure that Hessians were always zero when ```calculate_derivs``` was false, and I'll trust that you looked carefully to make sure that sensible behavior happens when the user specifies ```calculate_gradients=false``` but ```calculate_hessians=true```.  Better yet would be to just disallow that behavior when you create an SF.

src/SensitiveFloats.jl
```diff
@@ -271,8 +235,19 @@ function add_sources_sf!{ParamType <: ParamSet, NumType <: Number}(
...
 
-    true # Set definite return type
+
+function zero_sensitive_float_array(NumType::DataType,
+                                    local_P::Int,
+                                    local_S::Int,
+                                    d::Integer...)
+    sf_array = Array(SensitiveFloat{NumType}, d)
+    for ind in 1:length(sf_array)
+        # Do we always want these arrays to have gradients and hessians?
```
rgiordan, line 247: No; like the rest of the code, it will depend on the arguments passed in via ```ElboArgs```, though that may not be all implemented in the PSF code right now..

src/deterministic_vi/elbo_args.jl
```diff
@@ -50,78 +50,75 @@ type ElboIntermediateVariables{NumType <: Number}
...
     - calculate_hessian: If false, only calculate gradients. Note that if
-                calculate_derivs = false, then hessians will not be
+                calculate_gradient = false, then hessians will not be
                 calculated irrespective of the value of calculate_hessian.
 """
 function ElboIntermediateVariables(NumType::DataType,
                                    S::Int,
                                    num_active_sources::Int;
-                                   calculate_derivs::Bool=true,
+                                   calculate_gradient::Bool=true,
```
rgiordan, line 80: As long as you're making all these changes, I'd like to mention that I never quite understood why ```calculate_gradient``` and ```calculate_hessian``` aren't included in ```ElboArgs```.

src/deterministic_vi/elbo_kl.jl
```diff
@@ -40,23 +40,21 @@ function gen_beta_kl{NumType <: Number}(alpha2::NumType, beta2::NumType)
...
-            hess[1, 1] = alpha_diff * polygamma(2, alpha1) +
-                         both_inv_diff * quadgamma_both +
-                         trigamma_alpha1 - trigamma_both
-            hess[2, 2] = beta_diff * polygamma(2, beta1) +
-                         both_inv_diff * quadgamma_both +
-                         trigamma_beta1 - trigamma_both
-            hess[1, 2] = hess[2, 1] =
-                -trigamma_both + both_inv_diff * quadgamma_both
-        end
+        trigamma_alpha1 = trigamma(alpha1)
```
rgiordan, line 43: You are not worried about these wasted computations since they only happen once per ELBO value?

src/deterministic_vi/elbo_objective.jl
```diff
@@ -20,18 +20,17 @@ Returns:
 function calculate_source_pixel_brightness!{NumType <: Number}(
                     elbo_vars::ElboIntermediateVariables{NumType},
                     ea::ElboArgs{NumType},
-                    E_G_s::SensitiveFloat{CanonicalParams, NumType},
-                    var_G_s::SensitiveFloat{CanonicalParams, NumType},
-                    fs0m::SensitiveFloat{StarPosParams, NumType},
-                    fs1m::SensitiveFloat{GalaxyPosParams, NumType},
+                    E_G_s::SensitiveFloat{NumType},
+                    var_G_s::SensitiveFloat{NumType},
+                    fs0m::SensitiveFloat{NumType},
```
rgiordan, line 25: In my opinion, by removing the parameter types here, not documenting the inputs more carefully, and not asserting the expected lengths of the SF parameters, you are opening the door for some real difficult-to-detect bugs in the future if someone gets their param types and arguments mixed up.  These functions will silently do very wrong things if the user passes in param types that are not the ones it expects.  In my opinion, the possibility of this kind of error is not worth simplifying the type.  But it's your call, ultimately.  Some length asserts at least would be a good idea.

src/deterministic_vi/source_brightness.jl
```diff
@@ -74,135 +75,136 @@ function SourceBrightness{NumType <: Number}(
...
+        end
+
+        E_ll_a[3, i].v[] = exp(2 * r1[i] + 2 * r2[i])
+        E_ll_a[4, i].v[] = exp(2 * c1[3, i] + 2 * c2[3, i])
+        E_ll_a[5, i].v[] = exp(2 * c1[4, i] + 2 * c2[4, i])
+        E_ll_a[2, i].v[] = exp(-2 * c1[2, i] + 2 * c2[2, i])
+        E_ll_a[1, i].v[] = exp(-2 * c1[1, i] + 2 * c2[1, i])
+
+        if calculate_gradient
         # Band 3, the reference band.
```
rgiordan: Missing indent

src/deterministic_vi/source_brightness.jl
```diff
@@ -17,54 +17,55 @@ and all other rows are lognormal offsets.
...
+        end
+
+        E_l_a[3, i].v[] = exp(r1[i] + 0.5 * r2[i])
+        E_l_a[4, i].v[] = exp(c1[3, i] + .5 * c2[3, i])
+        E_l_a[5, i].v[] = exp(c1[4, i] + .5 * c2[4, i])
+        E_l_a[2, i].v[] = exp(-c1[2, i] + .5 * c2[2, i])
+        E_l_a[1, i].v[] = exp(-c1[1, i] + .5 * c2[1, i])
+
+        if calculate_gradient
         # band 3 is the reference band, relative to which the colors are
```
rgiordan: Missing indent

test/test_elbo.jl
```diff
@@ -36,23 +38,6 @@ function test_bvn_cov()
 end
 
 
-function test_derivative_flags()
```
rgiordan, line 39: Per my other comments about the changed semantics, I would really leave this test in, and add an additional test that if calculate_gradient is false but calculate_hessian is either the combination is disallowed or you get all zeros and not garbage.

rgiordan (approved):
This looks good to me modulo minor changes.  I think the test_derivative_flags should be restored and extended.  I like the has_gradient and has_hessian -- that tidies up a lot of functions and the speed gains are pretty impressive.  I don't think removing the parameter types in the SF specification is a good idea -- if I understood right, it does not give any performance gains and makes the code more error prone in what may someday prove very hard to debug.  But since you made the change, I don't expect to convince you, so go ahead and merge after the other minor fixes.

src/deterministic_vi/elbo_args.jl
jeff-regier: We can do that. Before I had in mind that `ElboArgs` would be just what was mathematically necessary to evaluate the objective function: the data, the model parameters, and the variational parameters. The calculate flags were kind of different. But since then `ElboArgs` has become kind of a catch all, which isn't necessarily a problem.

src/SensitiveFloats.jl
jeff-regier: I did find that happening in a couple of tests. I changed those, and adding an assertion to `SensitiveFloat` to ensure that if we're storing/computing a Hessian, we're also storing/computing a gradient. I think we're OK just not allowing users to compute Hessians wo gradients.

src/SensitiveFloats.jl
jeff-regier: Not yet, you may want to change this eventually. For the `SensitiveFloat` constructor, I prefer not having default values for `has_gradient` and `has_hessian` to make it explicit what memory is going to be allocated.

src/deterministic_vi/elbo_kl.jl
jeff-regier: That was my thinking. @jrevels's implementation with auto diff may be more memory/computation efficient.

src/deterministic_vi/source_brightness.jl
jeff-regier: fixed

src/deterministic_vi/source_brightness.jl
jeff-regier: fixed

src/deterministic_vi/elbo_objective.jl
jeff-regier: I see your point. At first I thought those kind of checks were a good use of the type system. But when I realized that that was causing Julia to compile a different version for each `ParamSet` it got called with, I wasn't so sure.  I do like now how `SensitiveFloat` isn't specific to any other Celeste data types. I'll just add some length asserts for the time being.

PR #455, opened 2016-12-13
More kernels

rgiordan:


rgiordan:
Looks like ```Optim.jl``` might not be passing on the julia nightly:
```
ERROR: LoadError: LoadError: syntax: field name "Optim.method_string" is not a symbol
```

jeff-regier:
It looks like the Optim.jl devs are working on it. Glad to see the cubic interpolation is working out.

andreasnoack:
> It looks like the Optim.jl devs are working on it

A new version of `Optim.jl` has been tagged so this should work again.

PR #465, opened 2016-12-16
callback functions

### LINKED ISSUE #456: Flexibly specify the ELBO function and initialization
I'd like to be able to flexibly specify alternative version of the ELBO function for stripe 82, the galsim benchmarks, and ordinary SDSS data.  Right now, the DeterministicVI ELBO is hard coded in a variety of places.

@jeff-regier , you have strong opinions about how to organize things, so I think it's best to let you handle this.  The ideal solution in my opinion would be a user-specifiable callback taking an ElboArgs and returning an optimized ELBO.

jeff-regier:
Closes #456 . @rgiordan -- The changes to `validate_on_stripe82.jl` show how to call your fft with the callbacks. If you specify `--fft` when calling that script, it runs your elbo code.


benchmark/speed/benchmark_rcf.jl
```diff
@@ -1,80 +0,0 @@
-#!/usr/bin/env julia
```
jeff-regier, line 1: `benchmark_quarter_degree.jl` is a better benchmark, so I'm getting rid of this one to avoid confusion. There aren't enough sources in this rcf for each thread to have enough work.

benchmark/stripe82/validate_on_stripe82.jl
```diff
@@ -65,25 +69,38 @@ where
...
     # the user gets the option to just run the scoring mode. Running a scoring
     # alone would primarily be useful for debugging.
-    if length(ARGS) == 0 || ARGS[1] != "--score-only"
-        # adding `; objid=...` to the call limits infer_field to that source.
-        # Potentially useful for debugging.
-        @time infer_field(rcf, datadir, outdir, joint_infer=true)
+    if !("--score-only" in ARGS)
+        wrap_joint(cnti...) = one_node_joint_infer(cnti...;
+                                                   termination_percent=0.9)
+        source_callback = "--fft" in ARGS ? infer_source_fft : infer_source
```
jeff-regier, line 85: @rgiordan  -- You might do the same kind of thing in the galsim benchmarks with a `--fft` flag.

src/ParallelRun.jl
```diff
@@ -12,7 +12,7 @@ import ..PSF
 
 import ..DeterministicVI: infer_source
 
-include("cyclades.jl")
+include("joint_infer.jl")
```
jeff-regier, line 15: @agnusmaximus -- I changed the name from `cyclades.jl` to `joint_infer.jl`...hope that's OK. I thought it was more clear for people not familiar with cyclades.

src/ParallelRun.jl
```diff
@@ -322,60 +322,24 @@ function infer_init(rcfs::Vector{RunCamcolField},
...
-A simplified helper method to choose between one_node_single_infer and one_node_joint_infer
-"""
-function parallel_infer(catalog, target_sources, neighbor_map, images;
-                        joint_infer=false,
-                        joint_infer_n_iters=50,
-                        joint_infer_batch_size=60,
-                        joint_infer_shuffle=true,
-                        reserve_thread=Ref(false),
-                        thread_fun=phalse,
-                        timing=InferTiming())
```
jeff-regier, line 335: No longer necessary because we just pass the particular inference method to `one_node_infer`, rather than passing it a flag. @agnusmaximus let me know if this doesn't make sense.

src/ParallelRun.jl
```diff
@@ -538,35 +499,6 @@ function get_overlapping_fields(query::BoundingBox, stagedir::String)
...
-"""
-called from main entry point for inference for one field
-(used for accuracy assessment, infer-box is the primary inference
-entry point)
-"""
-function infer_field(rcf::RunCamcolField,
-                     stagedir::String,
-                     outdir::String;
-                     joint_infer=false,
-                     objid="")
```
jeff-regier, line 550: This function does almost nothing, and it's only used by `validate_on_stripe82.jl` so I just did what this function used to do directly in `validate_on_stripe82.jl`.

test/test_kl.jl
```diff
@@ -71,7 +71,7 @@ function test_subtract_kl()
     Celeste.DeterministicVI.subtract_kl_source!(sf, kl_result, vs, kl_helper)
     @test sf.v[] == DiffBase.value(kl_result)
     @test sf.d === DiffBase.gradient(kl_result)
-    test_sf = JLD.load(joinpath(dirname(@__FILE__), "kl_values.jld"), "sf")
+    test_sf = JLD.load(joinpath(datadir, "kl_values.jld"), "sf")
```
jeff-regier, line 74: @jrevels -- I moved your jld file to the data directory with the other jld test files. Hope that's OK.

jrevels: Good call!

rgiordan:
Thanks for doing this, Jeff!  I ran it on stripe 82 and commented on
https://github.com/jeff-regier/Celeste.jl/pull/466.

Issue #457, opened 2016-12-13
Reduce the dimensions of FSM matrices


rgiordan:
Most of the 5x slowdown in the new rasterized PSF ELBO is in the FFT, and that can be made much faster by being a little bit smarter about the size of the matrices in ```FSMSensitiveFloatMatrices```. Jeff, you expressed interest in doing this.  Here are some ideas:

- The default size of the PSF matrix is probably bigger than it needs to be right now.  For most PSFs, most pixels are dark.
- The fs*m_image matrices are big enough to contain all pixels in the pixel masks.  They could instead be big enough to contain only the active pixels.
- As-is, a single image only has one set of fs*m_* matrices.  This means each must be large enough to contain all the sources.  If you arranged it so that each source had its own fs*m_* matrices, then they could be much smaller.
- Currently, active_pixels are set by the size of the image post convolution.  However, fs1m_image only needs to be as big as the image pre-convolution, which could be much smaller.  This would require estimating and keeping track of two sets of active pixels, though.

rgiordan:
Done in https://github.com/jeff-regier/Celeste.jl/pull/466

PR #466, opened 2016-12-16
Psf improvements

rgiordan:
I implement the changes I suggested for speeding up the FFT ELBO.  It was probably easier for me than it would be for someone else.  It's now only about 2x slower than the original ELBO.


jeff-regier (commented):
Great, thanks for figuring out #457 .

PR #474, opened 2016-12-18
Joint infer improve

agnusmaximus:
Fix joint infer in Galsim Benchmark to optimize all sources (not just the first one) since this is required for the joint infer to work.


jeff-regier:
Ah I see, I shouldn't have changed the target sources.

I wouldn't worry too much about the stripe 82 scores as long as they are close. Getting a higher value of the objective function and finding a stationary point is more important. The stripe 82 scores should improve as our model improves.

PR #476, opened 2016-12-19
Add optional FFT to Celeste

rgiordan:


gostevehoward (commented):
LGTM

Any objections if I make a PR changing `run_galsim_benchmark.jl` to use the `ArgParse` package?

jeff-regier:
It's kind of slow. Last I checked it adds about 2 seconds to start up time. But go ahead if you think it's worth it, or if it's gotten faster. (It's not `using ArgParse` that's slow, and it's not the parsing, it's compiling that function that does the parsing.)

gostevehoward:
For what it's worth, I timed this. Here are some wall clock times when
there's an invalid command-line arg (so execution is just initialization
and command line parsing):

without argparse: 6.78, 6.50, 6.29, 6.30, 6.45
with argparse: 9.34, 8.85, 8.67, 8.59, 8.55

It does add ~2s. Here are some times running the simple_star benchmark:

without argparse: 42.4, 44.8, 41.8, 43.6, 45.0
with argparse:  45.7, 45.4, 44.7, 44.9, 44.7

The added time is negligible when you're actually running a benchmark to
completion. However it could be annoying when you're debugging some new
code. So I'll leave it as is for now. It is really annoying that
ArgParse.jl is implemented that way but what can you do.

jeff-regier:
`DocOpt.jl`was even worse, adding like 4 seconds to start up. Even with pre-compilation, and with `SnoopCompile.jl`, it took that long. As you say, not a big deal for long runs but annoying for debugging.

PR #493, opened 2016-12-22
comment out a failing galsim test; restore CelesteEDA

jeff-regier:


jeff-regier:
@andreasnoack On Julia 0.6, on travis, we're getting a StackOverflowError from StaticArrays. Is this a known problem? Tests are all passing now on 0.5.

```
ERROR: LoadError: LoadError: StackOverflowError:
 in broadcast(::Base.#*, ::Float64, ::StaticArrays.SMatrix{2,2,Float64,4}) at /home/travis/.julia/v0.6/StaticArrays/src/arraymath.jl:43 (repeats 80000 times)
```

gostevehoward:
FYI it is possible the failing test was due to an out-of-date GalSim FITS
file, I had that problem yesterday. But the right solution is to version
the FITS file. I'll make an issue.

PR #495, opened 2016-12-23
Update to changes in Optim

andreasnoack:


andreasnoack:
The test errors on nightly will go away when https://github.com/JuliaArrays/StaticArrays.jl/pull/90 has been merged and a new version tagged.

jeff-regier:
Great, thanks for fixing static arrays, and for updating our calls to Optim.jl.

PR #499, opened 2016-12-28
CelesteEDA: iota after epsilon, not before

jeff-regier:
@rgiordan  I may have spotted a bug in `CelesteEDA`. Currently if both `include_iota` and `include_epsilon` are `true`, epsilon gets added to the synthetic image after the pixels are scaled by iota. But epsilon is in nanomaggies, not counts. So I think it epsilon should be added beforehand. Is that the case?

### DIFF
--- a/src/CelesteEDA.jl
+++ b/src/CelesteEDA.jl
@@ -93,11 +93,11 @@ function render_source(ea::ElboArgs, s::Int, n::Int;
         else
             error("Unknown field ", field)
         end
-        if include_iota
-            image[h2, w2] *= ea.images[n].iota_vec[h]
-        end
         if include_epsilon
             image[h2, w2] += ea.images[n].epsilon_mat[h, w]
+        end
+        if include_iota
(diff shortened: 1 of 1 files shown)


gostevehoward:
I made this change in d6956c
<https://github.com/jeff-regier/Celeste.jl/commit/d6956cb0de7d927fff1c90fe8037d8830f00de05#diff-af77e89e6d15d6681955569f5b5e87b6>
but
it looks like it got undone (maybe accidentally?) in 85870d
<https://github.com/jeff-regier/Celeste.jl/commit/85870d0563325a8c67865a2c243edd1b82e586ff#diff-af77e89e6d15d6681955569f5b5e87b6>
.

rgiordan:
Yeah, maybe it got accidentally undone in a merge.  Thank you for catching
this.

jeff-regier:
We may want to change the name `epsilon` to something like `sky_nmgy`, and `iota` to `electrons_per_nmgy` in the future. It's pretty easy to these backwards.

PR #500, opened 2016-12-29
Joint infer fft

agnusmaximus:
- Add use_fft flags to joint_infer.
- Reduce test cases for test_joint_infer from 104 sources -> 20 sources to speed up testing.
- Add code to test that unconstrained gradients are zero for 4 sources and stripe82. The zero gradient test for stripe82 is disabled since it fails on some sources where the gradient is not close enough to 0.
- Add code to test that joint infer on stripe82 achieves a better objective value than single infer.


jeff-regier:
Looks great! Thanks for making these additions.

Issue #471, opened 2016-12-17
Generate a field with ~500 sources from GalSim


gostevehoward:
Use the same dimensions as SDSS images, and report how often parameters are off by >1 SD, etc.

gostevehoward:
For reference, from #461 (my bad making the duplicate):

Generate a field-sized image (in all 5 bands) containing around 500 light sources. Get the parameters for these light sources by sampling from our prior distributions. Place the light sources uniformly at random in the image. Some should overlap. Modify the galsim benchmark so that it can run on just this image, maybe by specify the FITS file containing this data.

gostevehoward:
@jeff-regier: did you want to draw these from prior defined in `Model.load_prior()`? It has 99% prior probability of being a star, which it sounds like has been set artificially high to counteract the flexibility of the galaxy model? Do you want to bring that down for the purposes of generating this field? Or instead of drawing from the distribution did you want to just sample rows from a catalog such as [`coadd_for_4263_5_119.fit`](https://github.com/jeff-regier/Celeste.jl/blob/master/test/data/coadd_for_4263_5_119.fit) (and maybe randomize positions)?

jeff-regier:
Ah, no, I'll find a better way to encode a "penalty" for not having parameterization-invariant priors on the galaxies' parameters. For now, please just hard code in your script `a = [0.28, 0.72]`, i.e. 28% stars and 72% galaxies.

jeff-regier:
I haven't list priors yet for `e_fracdev`, `e_angle`, and `e_axis`. For `e_angle`, uniform on [0,180] is a reasonable prior. For `e_axis`, let's go with Beta(2,2) for now. For `e_fracdev`, maybe Beta(0.5, 0.5). I'll add more exact values to `Model.prior` in the future.

gostevehoward:
FYI the galsim image generation code currently doesn't have any support for `e_dev`, it generates all galaxies using a pure exponential profile. We can make an issue to add that support and add some appropriate test cases.

jeff-regier:
Is it galsim that doesn't support De Vaucouleurs galaxy profiles? Or you is it your Python code that only generates exponential galaxies? If the latter, what's involved in generating De Vaucouleurs galaxies too?

gostevehoward:
[...]

PR #501, opened 2016-12-30
Generate 500-source field using GalSim

gostevehoward:
Here's the code to generate a 500-source catalog from the Celeste prior and then generate a 1361 x 2048 pixel field image from that catalog. I refactored the GalSim generation code a fair bit. This mostly finishes off #471. The last piece of the puzzle is making a script that actually runs Celeste appropriately. We want to do joint inference over the whole field, with all sources as target sources? Do we want to initialize the Celeste catalog from the true values plus some noise in each field? Or did you have something else in mind?

FYI you can download the generated FITS file [here](http://portal.nersc.gov/project/dasrepo/celeste/galsim_fits/galsim_field_d485f30ae6.fits).


gostevehoward:
I will update the [wiki page](https://github.com/jeff-regier/Celeste.jl/wiki/GalSim-benchmarks) too...

jeff-regier:
Awesome! Yes, let's make all the sources target sources, and make their initial positions wrong by about 0.5 pixels on average, by drawing noise from a uniform distribution (not a gaussian). The initial position should never be off by more than 1 pixel. You could get all the other parameter's initial values, call `Model.init_source(init_pos::Vector{Float64})`. (Note that, unfortunately, there are two functions called `Model.init_source`. I'm suggesting using the one that takes a vector as an argument.) I think this is going to be really great test!

PR #519, opened 2017-01-11
[WIP] Refactor constraint transformations + ELBO maximization

### LINKED ISSUE #492: ReverseDiff for Transform.jl

jrevels:
There's still quite a bit of work to be done here, but I wanted to go ahead and open this PR so that it gets some exposure to the other maintainers. It might be quite a slog towards the finish line. I've added the new code to the source tree so that it can be reviewed, but haven't "hooked it up" to the other code yet (e.g. the old code is still being used on this branch). 

Following @jeff-regier's advice, this PR will ignore `PSF.jl` (and thus will keep the old `Transform.jl` around so `PSF.jl` can use it).

This refactors the constraint transformation code (previously in `Transform.jl`, now in `ConstraintTransforms.jl`) to reduce allocations, make things type-stable, and use AD for the derivatives. IMO, this new design should also be much easier to extend/modify in the future than the old design. This additionally refactors `maximize_elbo.jl` to use the new `ConstraintTransforms.jl` code.

I should write up a more detailed description of the changes at some point, but feel free to ask questions in the meantime.

There are a number of things left to do:

- [x] explore/replace `enforce_bounds`
- [x] allow callers to pass in their own constrain/unconstrain functions (and provide an equivalent of the default transforms given by `get_mp_transform` in the old code)
- [x] default constrain/unconstrain functions can optionally accept pre-allocated location constraints
- [x] add replacement for `omitted_ids`
- [x] write tests for new code and fix the inevitable bugs that will crop up
    - [x] test ConstraintTransforms.jl
    - [x] rewrite existing optimization tests to use NewtonMaximize
    - [x] remove/fix any other broken tests 
- [x] refactor callers of `maximize_f`:
    - [x] `infer_source` in `DeterministicVI.jl`
    - [x] `one_node_joint_infer` in `joint_infer.jl`
    - [x] `infer_source_fft` in `DeterministicVIImagePSF.jl`
- [x] run benchmarks in `benchmarks` folder once the new code is "hooked up" (I've done piecemeal benchmarks and the performance looks good so far)

Closes #492.

### DIFF
--- a/src/DeterministicVIImagePSF.jl
+++ b/src/DeterministicVIImagePSF.jl
@@ -13,7 +13,9 @@ import ..DeterministicVI:
     SourceBrightness, GalaxyComponent, SkyPatch,
     load_source_brightnesses, add_elbo_log_term!,
     calculate_source_pixel_brightness!,
-    KLDivergence, init_sources
+    KLDivergence, NewtonMaximize, init_sources
+
+using ..DeterministicVI.NewtonMaximize: maximize!, Config, custom_optim_options
 
 import ..Model:
     populate_gal_fsm!, getids, ParamSet, linear_world_to_pix, lidx,
@@ -31,7 +33,6 @@ import ..Infer:
 
 import ..PSF: get_psf_at_point, trim_psf
 
-import ..Transform: get_mp_transform
 
 import WCS
 
@@ -42,24 +43,24 @@ include("deterministic_vi_image_psf/elbo_image_psf.jl")
 
 export elbo_likelihood_with_fft!, FSMSensitiveFloatMatrices,
        initialize_fsm_sf_matrices!, initialize_fft_elbo_parameters,
-       get_fft_elbo_function, load_fsm_mat
+       FFTElboFunction, load_fsm_mat
 
 
 function infer_source_fft(images::Vector{Image},
                           neighbors::Vector{CatalogEntry},
                           entry::CatalogEntry;
                           min_radius_pix=Nullable{Float64}())
-   cat_local = vcat([entry], neighbors)
-   vp = init_sources([1], cat_local)
-   patches = get_sky_patches(images, cat_local)
-   load_active_pixels!(images, patches, min_radius_pix=min_radius_pix)
+    cat_local = vcat([entry], neighbors)
+    vp = init_sources([1], cat_local)
+    patches = get_sky_patches(images, cat_local)
+    load_active_pixels!(images, patches, min_radius_pix=min_radius_pix)
 
-   ea_fft, fsm_mat = initialize_fft_elbo_parameters(
-       images, vp, patches, [1], use_raw_psf=true)
-   elbo_fft_opt = get_fft_elbo_function(ea_fft, fsm_mat)
-   maximize_f(elbo_fft_opt, ea_fft, max_iters=150)
+    ea_fft, fsm_mat = initialize_fft_elbo_parameters(images, vp, patches, [1], use_raw_psf=true)
+    elbo_fft_opt = FFTElboFunction(fsm_mat)
+    cfg = Config(ea_fft; optim_options=custom_optim_options(max_iters=150))
+    maximize!(elbo_fft_opt, ea_fft, cfg)
 
-   vp[1]
+    return vp[1]
 end
 
 
@@ -73,7 +74,7 @@ function infer_source_fft_two_step(images::Vector{Image},
 
    ea_fft, fsm_mat = initialize_fft_elbo_parameters(
        images, vp, patches, [1], use_raw_psf=true)
-   elbo_fft_opt = get_fft_elbo_function(ea_fft, fsm_mat)
+   elbo_fft_opt = FFTElboFunction(fsm_mat)
 
    maximize_f_two_steps(elbo_fft_opt, ea_fft)
 
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,13 +1,13 @@
 julia 0.5
-Compat 0.13.0
+Compat 0.18.0
 DataFrames
 DataStructures 0.5.2
 Distributions
 FITSIO 0.8.4
-DiffBase 0.0.2
-ForwardDiff 0.3.3 0.4
-ReverseDiff 0.0.1 0.1
+DiffBase 0.0.3
+ForwardDiff 0.3.4 0.4
+ReverseDiff 0.0.3 0.1
 JLD 0.6.0
-Optim 0.7.3
+Optim 0.7.4
 StaticArrays 0.1.4
 WCS 0.1.3
--- a/bin/render_source.jl
+++ b/bin/render_source.jl
@@ -59,7 +59,7 @@ matshow(image_orig); colorbar(); title("original Celeste")
 ##############################################
 # Optimize the same source with both methods.
 
-elbo_fft_opt = DeterministicVIImagePSF.get_fft_elbo_function(ea_fft, fsm_mat);
+elbo_fft_opt = DeterministicVIImagePSF.FFTElboFunction(fsm_mat);
 f_evals_fft, max_f_fft, max_x_fft, nm_result_fft, transform_fft =
     DeterministicVI.maximize_f(elbo_fft_opt, ea_fft, verbose=true, max_iters=100);
 vp_opt_fft = deepcopy(ea_fft.vp[1]);
--- a/src/Celeste.jl
+++ b/src/Celeste.jl
@@ -9,6 +9,7 @@ include("SensitiveFloats.jl")
 
 include("Model.jl")
 include("Infer.jl")
+include("ConstraintTransforms.jl")
 include("Transform.jl")
 include("PSF.jl")
 include("SDSSIO.jl")
--- a/src/ConstraintTransforms.jl
+++ b/src/ConstraintTransforms.jl
@@ -1,0 +1,483 @@
+module ConstraintTransforms
+
+using ..Model
+using ..SensitiveFloats
+using Compat
+
+typealias ParameterBatch{T<:Number} Vector{Vector{T}}
+
+####################
+# Constraint Types #
+####################
+
+abstract Constraint
+
+immutable BoxConstraint <: Constraint
+    lower::Float64
+    upper::Float64
+    scale::Float64
+    function BoxConstraint(lower, upper, scale)
+        @assert lower > -Inf
+        @assert lower < upper
+        @assert upper < Inf
+        @assert 0.0 < scale < Inf
+        return new(lower, upper, scale)
+    end
+end
+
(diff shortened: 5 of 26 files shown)


jeff-regier:
It looks to me like `enforce_bounds!` is just used to initialize the variational parameters. But shouldn't they already be in bounds, if they're initialized with [`generic_init_source `](https://github.com/jeff-regier/Celeste.jl/blob/master/src/DeterministicVI.jl#L27) or [`catalog_init_source`](https://github.com/jeff-regier/Celeste.jl/blob/master/src/DeterministicVI.jl#L49)?

@rgiordan -- Would you explain the idea behind `enforce_bounds!`? If it's just for unit tests, I'd rather move it to the `tests/` directory, and explicitly call it in the tests before calling `maximize_f`.

rgiordan:
```enforce_bounds!``` ensures that the program will not terminate if you pass parameters to the optimization that cannot be expressed within the transform bounds.

Here's an example of why that's useful outside of tests.  Through ```simplex_min``` in ```Transform.jl```, we don't allow E[a] to be less than ```0.005```.  Suppose I want to start optimization at essentially "all star".  If I actually initialize to ```a[1] = 1```, Celeste will crash when it is unable to express this in constrained space.  To avoid having to look up the legal minimum, you can just set the initial parameters to what you want and guarantee that Celeste will not crash.

As another use case, I don't think we want Celeste to crash if there is an initial catalog value that cannot be expressed in the constrained space.  Rather, we'd like it to be brought as close as is allowed and proceed with optimization.

**Comment by jeff-regier** [comment]
I think the easiest way to get this PR merged is to keep `enforce_bounds!` around, effectively re-implementing it to do what the old version does. I don't entirely agree with what Ryan says about its usefulness outside of tests:  `catalog_init_source` already does thresholding, so it should give reasonable, in-bounds values for initialization, regardless of the current catalog. (In the future, we may want `catalog_init_source` to also return the constraints for a particular light source.) Second, nowhere outside of the tests are parameters set to values that could be out of bounds, e.g. `a[1] = 1` doesn't happen outside of tests. That said, it is easiest not to change too many aspects of the code at once, and `enforce_bounds!` doesn't seem like a problem now that I understand it, just a particular design decision.

**Comment by jrevels** [comment]
I'm fine with reimplementing `enforce_bounds!`, I'll play around with that tomorrow.

**Comment by rgiordan** [comment]
On Mon, Jan 16, 2017 at 1:46 PM, Jeffrey Regier <notifications@github.com>
wrote:

> e.g. a[1] = 1 doesn't happen outside of tests.
>

I do this all the time when debugging the model or using it interactively.

**Comment by jeff-regier** [comment]
@jrevels Would add an equality constraint type too? Something like

```
immutable EqualityConstraint <: Constraint
    c::Vector{Float64}
end
```

For a `ParameterConstraint{EqualityConstraint}` instance, I think `free_indices` would just be `Int[]`---the bound parameters are omitted from unconstrained parameterization. When mapping from the unconstrained parameters to the constrained parameters, the entries of `c` are inserted. So `length(c) == length(bound_indices)`, if I understand your approach correctly.

@rgiordan -- Will this functionality suffice for issue #480?

**Comment by jrevels** [comment]
Adding that should be pretty easy. Is there any other code I'd need to update to use that, or will it mainly be useful in the future/for debugging?

**Comment by jeff-regier** [comment]
Great, thanks Jarrett. It's just for debugging, at least for now. Ryan is having trouble getting to the bottom of some problems with the FFT optimization without it.

**Comment by rgiordan** [comment]
@jeff-regier , I'm not sure what you're proposing re: https://github.com/jeff-regier/Celeste.jl/issues/480.  Are you suggesting adding the capability to specify a custom ```constrain_model``` function when you call ```newton_maximize!```?  Wouldn't it be simpler to just fix ```omitted_ids```?

**Comment by jeff-regier** [comment]
Yes, that's the idea. It should be easier to add and maintain an `EqualityConstraint` type than to add `omitted_ids` logic back.

As a stopgap, could you set the bounds for the parameters you don't want to change to a very narrow range? Would the let you continue to make progress while Jarrett works on Transform?

**Comment by rgiordan** [comment]
Tight bounds will result in large gradients and Hessians for those parameters.

I don't quite understand your reluctance to fix ```omitted_ids```.  It looks like someone even tried to implement it (e.g., there is an ```all_kept_ids``` defined), but they didn't test it and it has a (probably minor) bug.  Please fix ```omitted_ids```.

**Comment by jrevels** [comment]
@jeff-regier I've updated this PR's TODO list, can you review what I've put there and let me know if I missed anything/should change anything? 

I finished a couple of the points we talked about in today's meeting. If I can keep going at this pace (i.e. the testing/bug-fixing phase doesn't reveal any monstrous problems), then hopefully this PR can make it in before the January hackathon.

**Comment by jeff-regier** [comment]
@jrevels -- That TODO list looks good to me. Please also include a unit test that demonstrates how to use `EqualityConstraint` to optimize just the star parameters, while leaving the other parameters fixed at their initial value. Thanks for working on this issue.

File: src/DeterministicVIImagePSF.jl [hunk 1]
```diff
@@ -49,17 +49,18 @@ function infer_source_fft(images::Vector{Image},
...
-
-   vp[1]
+    cat_local = vcat([entry], neighbors)
+    vp = init_sources([1], cat_local)
+    patches = get_sky_patches(images, cat_local)
+    load_active_pixels!(images, patches, min_radius_pix=min_radius_pix)
+
+    ea_fft, fsm_mat = initialize_fft_elbo_parameters(images, vp, patches, [1], use_raw_psf=true)
+    elbo_fft_opt = get_fft_elbo_function(ea_fft, fsm_mat)
+    bound_params = ea.vp[ea.active_sources]
```
jeff-regier: @jrevels This line should be ```bound_params = ea_fft.vp[ea_fft.active_sources]``` presumably. Not sure if that's the source of the test error you're seeing.

jrevels: Thanks. It's unfortunately not (UndefVarErrors/syntax errors are pretty easy to track down, thankfully). I fixed this and a couple of other things locally, I just haven't pushed yet.

jeff-regier:
All tests pass for me after rebasing on master.

Should this whole PR get squashed into one commit? Or do you want to group it into multiple commits?

jeff-regier:
Scores look good from `validate_on_stripe82.jl --joint`! I think we're in business.

jeff-regier:
The performance regression looks real, unfortunately, at least with Julia 0.5. From `benchmark_sixteenth_degree.jl`, with 3 threads, on a macintosh:

On master:
```
386.197440 seconds (1.04 G allocations: 63.712 GB, 4.50% gc time)
```

On "jr/constraints" (following rebase)
```
477.714822 seconds (1.05 G allocations: 52.766 GB, 4.10% gc time)
```

jrevels:
Hmm...benchmarking `NewtonMaximize.maximize!` vs. the old `maximize_f` with a single source and `DeterministicVI.elbo` as the objective indicates that `NewtonMaximize.maximize!` is slightly faster with a fairly substantial (~40%) memory improvement (on v0.6 anyway). I would benchmark with multiple simultaneous sources, but the old Transform code errors out in that case.

This PR shouldn't cause drastic performance changes in non-NewtonMaximize related code, so there must be a performance bug somewhere to get a regression that large. I'll do some benchmarking/regression hunting tomorrow to figure this out.

jeff-regier:
I ran `benchmark_seven_light_sources.jl --profile` on master and this branch. The lines below show up in the profiler output only on this branch. They seem to account most of the difference in runtime (total runtime is 747).

```
    79 .../derivatives/elementwise.jl   197 (::ReverseDiff.##207#208{ForwardDiff.Dual{... 
    80 ./broadcast.jl                   117 _broadcast!(::ReverseDiff.##207#208{Forwar... 
    80 ./broadcast.jl                   172 broadcast!                                    
    80 ./broadcast.jl                   123 macro expansion                               
    80 ./broadcast.jl                   129 macro expansion                               
    80 ./simdloop.jl                     73 macro expansion                               
    80 .../derivatives/elementwise.jl   259 special_forward_exec!(::ReverseDiff.Specia... 
```

jrevels:
Do you have non-truncated output (maybe post it in a gist)? This PR doesn't employ ReverseDiff in any new way at all vs. master - all the AD that is added here is ForwardDiff - so ReverseDiff shouldn't be involved in the regressions we're seeing.

jrevels:
@jeff-regier I've fixed the performance regression I was seeing locally, let me know if that fixes it on your end. It was from an unnecessary change I made that stack-allocated the `ids` indices instead of heap-allocating them (i.e. nothing to do with AD or the substantive part of this PR). I'm still not certain why that caused a regression, but reverting that change fixed it.

jeff-regier:
It's close now, about 5% difference. Looks like it could be due to GC time. Strange, since this branch has both fewer allocations and less memory allocated in total.

These runtimes are on a dedicated Cori Phase 1 node, with the version of Julia 0.6 that Andreas built, stored in his scratch directory. I'm running `benchmark_seven_light_sources.jl`.

master:
```
  5.207721 seconds (4.71 M allocations: 312.227 MiB, 1.20% gc time)
  5.175237 seconds (4.71 M allocations: 312.227 MiB, 1.19% gc time)
  5.203087 seconds (4.71 M allocations: 312.227 MiB, 1.21% gc time)
```

This branch:
```
  5.499711 seconds (4.26 M allocations: 183.664 MiB, 4.01% gc time)
  5.489853 seconds (4.26 M allocations: 183.664 MiB, 4.07% gc time)
  5.540679 seconds (4.26 M allocations: 183.664 MiB, 4.02% gc time)
```

jeff-regier:
Do you want to merge this though as it is? Please go ahead whenever you're ready.

jrevels:
I'm going to go ahead and merge (EDIT: well, in a few minutes, I'm cleaning up the commit history now), but I'll investigate the regression you're seeing.

Performance seems pretty variable on my machine, with this branch appearing to perform better in most cases:

```
jr/constraints: 
7.301494 seconds (4.26 M allocations: 183.985 MiB, 0.61% gc time)
7.459311 seconds (4.26 M allocations: 183.985 MiB, 2.39% gc time)
7.855283 seconds (4.26 M allocations: 183.985 MiB, 2.50% gc time)

master: 
8.326991 seconds (4.74 M allocations: 324.078 MiB, 0.98% gc time)
7.576315 seconds (4.74 M allocations: 323.980 MiB, 0.80% gc time)
7.630184 seconds (4.74 M allocations: 323.980 MiB, 2.69% gc time)
```

That's only with a single thread though - what was your `JULIA_NUM_THREADS`?

jrevels:
Alright, cleaned up the commit history, let's make sure this passes CI and then I'll merge.

PR #536, opened 2017-01-25
Update to Julia 0.6

andreasnoack:
This is mainly fixing syntax deprecations on master. It shouldn't change anything on 0.5. However, it revealed `NaN`s in the `fft` tests.

Some dependencies still need to be tagged in METADATA so this won't pass on master but should go through on 0.5.


jeff-regier:
Great, thanks a lot Andreas!

Issue #482, opened 2016-12-19
GalSim noisy galaxy cases highly sensitive to `noise_fraction` in `load_active_pixels!`


gostevehoward:
I noticed this past weekend that estimated brightness and HL radius are way off for noisy galaxy cases in the GalSim benchmarks:

```
│ Row │ label               │ field                        │ expected   │ single_infer_actual │ joint_infer_actual │
├─────┼─────────────────────┼──────────────────────────────┼────────────┼─────────────────────┼────────────────────┤
│ 1   │ "galaxy_with_noise" │ "X center (world coords)"    │ 0.00541833 │ 0.00542928          │ 0.00542929         │
│ 2   │ "galaxy_with_noise" │ "Y center (world coords)"    │ 0.00514056 │ 0.00513782          │ 0.00513782         │
│ 3   │ "galaxy_with_noise" │ "Minor/major axis ratio"     │ 0.4        │ 0.415016            │ 0.415019           │
│ 4   │ "galaxy_with_noise" │ "Angle (degrees)"            │ 15         │ 14.7824             │ 14.7822            │
│ 5   │ "galaxy_with_noise" │ "Half-light radius (pixels)" │ 6.31313    │ 25.6438             │ 25.6095            │
│ 6   │ "galaxy_with_noise" │ "Brightness (nMgy)"          │ 15         │ 37.1759             │ 37.158             │
│ 7   │ "galaxy_with_noise" │ "Probability of galaxy"      │ 1          │ 0.995               │ 0.995              │
```

I sadly hadn't noticed this before, but it's present in some [old benchmark results](https://github.com/jeff-regier/Celeste.jl/pull/435) (click the pastebin link). I spent a while manually tracking this down (git bisect wasn't possible because the benchmark harness was often broken) and I think it started in 5878520b41c00a9567fced3d29923f08eb3a8e97 (the change to eliminate tiles).

Here's the inferred brightness and radius for various values of `noise_fraction`, for the "galaxy_with_noise" case (true brightness 15, true radius 6.31):

noise_fraction|brightness|HL radius
------------|-----------|---------
-0.9|14.9161|5.87814
-0.5|15.2943|6.13193
-0.4|15.6574|6.37313
-0.3|16.3363|6.83299
-0.2|17.5608|7.77823
-0.1|24.4099|14.7574
0|26.7117|16.9204
0.1|29.3006|19.2417
0.2|31.951|21.5993
0.5|37.1759|25.6438

[...]

PR #538, opened 2017-01-25
Set `min_radius_pix` for GalSim unit tests

gostevehoward:
This gets the GalSim unit tests passing again, per #537.

This is the simplest implementation but I think adding all these keyword args is ugly and not sustainable. First, it's just ugly to have a web of kw args spreading through the code. Second, because you want to make it optional, but don't want to duplicate the default value so many times, you end up doing all this Nullable business, which isn't particularly pretty in Julia. (You could alternatively make a global const holding the default value but I find that ugly too for all the global references.)

I don't know of a good clean alternative (setting global flags is a bad idea in my opinion) other than trying to solve the problem more holistically. There's just no easy way to pass options deep down into the call tree from the top level, that I know of. One possibility would be to make a big Configuration type, which holds options like this for the entire Celeste engine. I think it'd be best structured as a nested set of types, with a structure roughly mirroring the high-level modules of the Celeste code. (I'm using "modules" informally, not implying it has to match the Julia module structure of Celeste exactly.)

For example, maybe three modules of the Celeste code are 1) image preparation (choosing patches, active pixels, etc), 2) multiple source handling (single/joint inference & threading), 3) the inference method (MOG, FFT). (This is probably not a good/accurate separation but bear with me for the sake of example.) You might have something like

```julia
type ImagePreparationConfig
  radius_override_px::Nullable{Float64}
  min_radius_px::Float64
  noise_fraction::Float64
end
ImagePreparationConfig() = ImagePreparationConfig(Nullable{Float64}(), 8.0, 0.5)

abstract InferenceConfig
type MixtureOfGaussiansConfig <: InferenceConfig ...(options)... end
type ImageFFTConfig <: InferenceConfig ...(options)... end

abstract MultipleSourceConfig
type SingleInferConfig <: MultipleSourceConfig
  inference_config::InferenceConfig
  ...
end
type JoinInferConfig <: MultipleSourceConfig
  inference_config::InferenceConfig
  ...
end

type CelesteConfig
  image_preparation_config::ImagePreparationConfig
  multiple_source_config::MultipleSourceConfig
end
CelesteConfig() = ...(defaults)...

function run_celeste(config::CelesteConfig, catalog, images, target_sources, ...etc...)
  patches = prepare_images(config.image_preparation_config, images, catalog)
  results = run_inference(config.multiple_source_config, patches, catalog, target_sources)
  ...etc...
end

function run_inference(config::SingleInferConfig, ...)
  ...
  results = infer_source(config.inference_config, ...)
  ...
end
function run_inference(config::JointInferConfig, ...)
  ...
  results = infer_source(config.inference_config, ...)
  ...
end

# in DeterministicVI.jl
function infer_source(config::MixtureOfGaussiansConfig, images, neighbors, entry) ...
# in DeterministicVIImagePSF
function infer_source(config::ImageFFTConfig, images, neighbors, entry) ...

# top-level caller
config = CelesteConfig()
config.image_preparation_config.min_radius_px = 40.0
config.inference_config = ImageFFTConfig()
run_celeste(config, catalog, ...)
```

Some notable things:
* Top-level caller code is clean, and only does as much work as it wants to to customize the defaults.
* Defaults are encoded in one place, the config constructor.
* Each submodule only receives the config relevant to its operation, which should make the code a bit easier to understand (than if a single giant config object were passed everywhere).
* Instead of having different entry points for different versions of a module (e.g., `infer_source()` and `infer_source_fft()`) and explicitly dispatching with conditionals or callback params (e.g., `infer_source_callback`), we dispatch a single entry point method call on the type of the config object, for each module.

This is benefits but it's not simple, which is why I coded the "simple" approach in this PR so you can see it. Of course we don't need a mass conversion, this could be implemented somewhat piecemeal.

I also admit this is inspired by how I'd structure such a thing in an OO paradigm, so I hope there's a simpler, equally beneficial and more idiomatic solution in the Julia paradigm which I'm not thinking of, if so let me know!

### DIFF
--- a/src/DeterministicVI.jl
+++ b/src/DeterministicVI.jl
@@ -111,7 +111,8 @@ Arguments:
 """
 function infer_source(images::Vector{Image},
                       neighbors::Vector{CatalogEntry},
-                      entry::CatalogEntry)
+                      entry::CatalogEntry;
+                      min_radius_pix=Nullable{Float64}())
     if length(neighbors) > 100
         Log.warn("Excessive number ($(length(neighbors))) of neighbors")
     end
@@ -122,7 +123,7 @@ function infer_source(images::Vector{Image},
     cat_local = vcat([entry], neighbors)
     vp = init_sources([1], cat_local)
     patches = Infer.get_sky_patches(images, cat_local)
-    Infer.load_active_pixels!(images, patches)
+    Infer.load_active_pixels!(images, patches, min_radius_pix=min_radius_pix)
 
     ea = ElboArgs(images, vp, patches, [1])
     f_evals, max_f, max_x, nm_result = maximize_f(elbo, ea)
--- a/src/DeterministicVIImagePSF.jl
+++ b/src/DeterministicVIImagePSF.jl
@@ -43,11 +43,12 @@ export elbo_likelihood_with_fft!, FSMSensitiveFloatMatrices,
 
 function infer_source_fft(images::Vector{Image},
                           neighbors::Vector{CatalogEntry},
-                          entry::CatalogEntry)
+                          entry::CatalogEntry;
+                          min_radius_pix=Nullable{Float64}())
    cat_local = vcat([entry], neighbors)
    vp = init_sources([1], cat_local)
    patches = get_sky_patches(images, cat_local)
-   load_active_pixels!(images, patches)
+   load_active_pixels!(images, patches, min_radius_pix=min_radius_pix)
 
    ea_fft, fsm_mat = initialize_fft_elbo_parameters(
        images, vp, patches, [1], use_raw_psf=false)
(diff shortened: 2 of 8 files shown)


gostevehoward:
I should mention for reference that the `min_radius_px` change is also fixing (or working around) #482.

jeff-regier:
Let's bring up some of these issues with Keno tomorrow. I bet he has away do it that won't leave you missing OOP.

gostevehoward:
OK. I added some abstract types in my sample code where they ought to be used. For what it's worth I don't think the above solution is necessarily bad, I can just see how some might find it overly complex, and I'm open to simpler alternatives. It's a matter of taste.

PR #539, opened 2017-01-25
Fully support `e_dev` in GalSim benchmarks

gostevehoward:
All fairly routine, probably the only thing of note is that I started getting errors like this from GalSim:

```
Traceback (most recent call last):
  File "galsim_field.py", line 53, in <module>
    main()
  File "galsim_field.py", line 50, in main
    generate_test_image.generate_fits_file('galsim_field')
  File "/vagrant/generate_test_image.py", line 357, in generate_fits_file
    image = test_case.construct_image(band_index, uniform_deviate)
  File "/vagrant/generate_test_image.py", line 279, in construct_image
    galsim_light_source.drawImage(image, add_to_image=True)
  File "/usr/lib/python2.7/dist-packages/galsim/base.py", line 1337, in drawImage
    image.added_flux = prof.SBProfile.draw(imview.image, gain, wmult)
RuntimeError: SB Error: fourierDraw() requires an FFT that is too large, 6144
If you can handle the large FFT, you may update gsparams.maximum_fft_size.
```

My hunch is that the de Vaucouleurs galaxies are effectively bigger and so the occasionally draw that gets both a large HL radius and a large dev weight will cause the above error (I saw some direct evidence for this but didn't prove it). I tried setting `maximum_fft_size` as suggested but then got a MemoryError, I think originating from the underlying GalSim C code so not easy to debug.

Anyhow, GalSim's supports another drawing method called photon shooting, an alternative to the default FFT method. I didn't read in detail but it isn't hard to imagine what the two are doing. So I switched to photon shooting with the requirement that the results are accurate to within 0.1% of sky noise. It's much, much faster to generate the field images now, and the unit tests still pass for the benchmarks, so I think this is a win.

### DIFF
--- a/benchmark/galsim/galsim_benchmarks.py
+++ b/benchmark/galsim/galsim_benchmarks.py
@@ -65,6 +65,14 @@ def bright_galaxy(test_case):
     test_case.add_galaxy().reference_band_flux_nmgy(20)
 
 @generate_test_image.galsim_test_case
+def de_vaucouleurs_galaxy(test_case):
+    test_case.add_galaxy().de_vaucouleurs_mixture_weight(1)
+
+@generate_test_image.galsim_test_case
+def exp_dev_mixture_galaxy(test_case):
+    test_case.add_galaxy().de_vaucouleurs_mixture_weight(0.4)
+
+@generate_test_image.galsim_test_case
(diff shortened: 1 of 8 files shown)


jeff-regier:
Sounds great!

PR #541, opened 2017-01-26
Multinode functionality

kpamnany:
Also changed parts of `ParallelRun.jl`; @jeff-regier: please review.


Keno:
Do we really want to comming cfg/sky_boxes?

kpamnany:
It's the current best-known sorting of boxes in decreasing order of difficulty. We need to use that for the hero run.

jeff-regier:
@kpamnany Looks fine except that you accidentally included `sky_boxes` in the commit!

kpamnany:
I did that intentionally @jeff-regier; I consider the sorting to be useful/important information about SDSS. I can remove it though, if you prefer?

jeff-regier:
Yeah, please exclude it. We've been keeping data files on remote servers, and having Celeste fetch them as needed, so that's one option.

Keno:
You'll need to squash this. Other it'll take up space in the repo

PR #551, opened 2017-01-31
Fft mag r

rgiordan:
This fixes a major bug in the FFT code in which I wasn't adding E_G before calculating the ELBO.  It seems to get rid of some egregious errors on stripe 82, though the aggregate scores are still worse for FFT than MOG.

I also add two-step optimization which seems to do slightly better on distinguishing stars from galaxies.  The first step only fits the star and skips parts of the code that do galaxy rendering.  Starting from the star optimum, the full model is then fit.

Finally, it expands the rendering functions to allow viewing multiple sources at once.


jeff-regier:
Looks like the travis tests are failing with the error
```
The job exceeded the maximum time limit for jobs, and has been terminated.
```

The benchmarks are taking twice as long too. Are you doing a lot of additional computation that we weren't doing before? Could this be a type instability thing?

rgiordan:
It appears that there's a lot more time spent evaluating the ELBO.  I took a brief look at a profiler though and nothing jumps out to me as an obvious error.  Maybe the pros want to take a look at my changes.  There are two more calls to ```add_sources_sf!``` per pixel in ```accumulate_image_pixel_brightness!``` than before.  I'm not sure how they can be avoided, though.

Tests are slower on my machine but do not time out.

jeff-regier:
With this PR the mog version of Celeste becomes 5x or 10x slower too. At
least that's what it seemed like when I ran `validate_on_stripe82.jl`
without any flags. Does this PR intentionally change code that affects mog
Celeste?

rgiordan:
I pushed a change that seems to help.

rgiordan:
Yes, it allows you to calculate the ELBO for the star only for active sources, which also affects the old code.

jeff-regier:
Many light sources' brightnesses are off by multiple orders of magnitude still, some by even more than before.
```
julia> tail(sort(rae["celeste_err"][:, [:mag_r, :objid]]))
6×2 DataFrames.DataFrame
│ Row │ mag_r   │ objid               │
├─────┼─────────┼─────────────────────┤
│ 1   │ 2.28066 │ 8647474692482203916 │
│ 2   │ 2.53062 │ 8647474692482203881 │
│ 3   │ 3.01757 │ 8647474692482203819 │
│ 4   │ 3.16659 │ 8647474692482203817 │
│ 5   │ 6.11071 │ 8647474692482138341 │
│ 6   │ 6.54213 │ 8647474692482203737 │
```

Am I running the wrong version or with the wrong flags? I did 
```
$ ./validate_on_stripe82.jl --fft_two_step
```

rgiordan:
Yes, that's the right command.  All of those are dim objects, except for the last one, which is a missed galaxy.  Maybe that's throwing the coadd estimate off for some reason.  In each case, if you look at what the FFT is doing, it looks sensible, and similar to the MOG.

Since this is a bug fix, let's merge it.  We can talk sometime about how to best make use of stripe 82 results offline.

PR #552, opened 2017-01-31
Avoid overflow in constrain_to_simplex and add missing assignment

### LINKED ISSUE #550: NaN in VP transform
This occurs in objid ```1237663784734490687``` and is an example of a source that causes the stripe 82 tests to fail.

The cause is in the transform of the k variable.  The optimization method puts a large weight on one component, e.g. something like

```
|  Row │ id        │ x         │
├─────┼───────────┼───────────┤
│ 28  │ "k_1_1"   │ 5.29276   │
│ 29  │ "k_2_1"   │ 327668.0  │
│ 30  │ "k_3_1"   │ -4.33043  │
│ 31  │ "k_4_1"   │ -2.01636  │
│ 32  │ "k_5_1"   │ -2.18093  │
│ 33  │ "k_6_1"   │ -2.661    │
│ 34  │ "k_7_1"   │ -2.18017  │
```
What we want to happen after the transform is 

```
|  Row │ id        │ x         │
├─────┼───────────┼───────────┤
│ 29  │ "k_1_1"   │ 0.00125     │
│ 30  │ "k_2_1"   │ 0.99125         │
│ 31  │ "k_3_1"   │ 0.00125     │
│ 32  │ "k_4_1"   │ 0.00125     │
│ 33  │ "k_5_1"   │ 0.00125     │
│ 34  │ "k_6_1"   │ 0.00125     │
│ 35  │ "k_7_1"   │ 0.00125     │
│ 36  │ "k_8_1"   │ 0.00125     
```

But instead we get ```k_2_1 = NaN```.  The reason is that ```constrain_to_simplex``` does
```
        z = exp.(x)
        z_sum = sum(z) + 1
        z ./= z_sum
```

Dividing ```Inf``` (in the second component of z) by ```Inf``` (in the sum) gives ```NaN```.

Since I'm stepping back from the project, I'll leave it to the current maintainer of the Transform module to fix and test.  @jeff-regier , I'm not sure who this is.

andreasnoack:
This uses the log-sum-exp trick to avoid that the simplex constraint gives `NaN`. There was also a missing assignment in the branch that handles cases with infinite values. Consequently, the sum of the simplex parameters wasn't one if two input values were `Inf`. This makes all the tests in `validate_on_stripe82.jl` pass on my machine.

There is a test failure in `test_Transforms.jl`, though. Basically, the test examines the case where all input values are `-Inf` (in this case a single value). Is that really a relevant test case? If so, what is a reasonable value of the simplex parameters if all inputs are `-Inf`?

Fixes #550

### DIFF
--- a/src/Transform.jl
+++ b/src/Transform.jl
@@ -46,19 +46,20 @@ Convert an (n - 1)-vector of real numbers to an n-vector on the simplex, where
 the last entry implicitly has the untransformed value 1.
 """
 function constrain_to_simplex{NumType <: Number}(x::Vector{NumType})
-    if any(x .== Inf)
+    m = maximum(x)
+    if m == Inf
         # If more than 1 entry in x is Inf, it may be because the
         # the last entry in z is 0. Here we set all those entries to the
         # same value, though that may not be strictly correct.
-        z = NumType[ x_entry .== Inf ? one(NumType) : zero(NumType) for x_entry in x]
-        z ./ sum(z)
+        z   = NumType[ x_entry .== Inf ? one(NumType) : zero(NumType) for x_entry in x]
+        z ./= sum(z)
         push!(z, 0)
         return z
     else
-        z = exp.(x)
-        z_sum = sum(z) + 1
+        z = exp.(x .- m)
+        z_sum = sum(z) + exp(-m)
         z ./= z_sum
-        push!(z, 1 / z_sum)
+        push!(z, inv(z_sum)*exp(-m))
         return z
     end
 end
--- a/test/test_transforms.jl
+++ b/test/test_transforms.jl
@@ -191,6 +191,13 @@ function test_basic_transforms()
     @test Transform.unconstrain_simplex([1.0, 0.0]) ≈ [Inf]
 
     @test Transform.constrain_to_simplex([Inf, 5])  ≈ [1.0, 0.0, 0.0]
+
+    @test sum(Transform.constrain_to_simplex([Inf, Inf])) ≈ 1.0 # Make sure that it's a simplex when there is more than one Inf
+
+    @test sum(Transform.constrain_to_simplex([709.0, 709.0, 709.0])) ≈ 1.0 # sum overflows
+    @test sum(Transform.constrain_to_simplex([710.0, 710.0, 710.0])) ≈ 1.0 # each element overflows
+    @test sum(Transform.constrain_to_simplex([88.0f0, 88.0f0, 88.0f0])) ≈ 1.0f0 # sum overflows
+    @test sum(Transform.constrain_to_simplex([89.0f0, 89.0f0, 89.0f0])) ≈ 1.0f0 # each element overflows
 end
 
 


jeff-regier:
Thanks for the fix Andreas.

I suspect supporting `Inf` values in Transform was a mistake, no doubt introduced to handle some situation, but probably not the best way to deal with it: numerical optimizers can't do anything with `Inf` values. Perhaps we'd want to have done something  differently earlier in the code, or throw an exception in `Transform`. @jrevels What are you doing about `Inf` values in your version of `Transform`? Can we just drop support for this case?

jrevels:
I had also noticed the missing assignment, so I just removed it from the `ContraintTransforms` version of that function since I figured the goal was to match the original `Transform`'s behavior. It seems that the only places `Inf`s should reasonably come into play are in the constraint bounds, but that shouldn't mean that `Inf`s every really get introduced in the constrained/unconstrained parameters. 

Not having to support `Inf` values in the free parameterization would clean up some code and make things more performant. It would also make testing simpler, since I wouldn't have to test the `Inf` branches that are currently there. Definitely let me know if we decide we don't need to support `Inf` parameters.

jeff-regier:
Seems like `prevfloat(Inf)` would work just was well as `Inf` as an upper bound to a box constraint.

jrevels:
Right, but regardless, just setting the constraint bounds to `Inf` shouldn't introduce any `Inf`s into the parameters (in other words, I'm agreeing with you that I don't see any reason to support `Inf`s in the parameters)

jeff-regier:
Maybe two different issues here: whether `Inf` is allowed as an upper bound (on a constrained parameter), and whether `Inf` is allow as an unconstrained parameter value. But come to think of it, maybe we shouldn't allow `Inf` in either case. It's certainly better to catch a problem earlier than to try to do something sensible with `Inf` values.

andreasnoack:
I just tried to run the stripe82 benchmarks with the branch that handles `Inf`s disabled and they give the same as before so I think that branch can probably just be substituted for an `isfinite` test. I think something like this will be included in @jrevels's PR.

jeff-regier:
I'm seeing julia 0.5 tests errors now on master, I think stemming from this PR. I thought the 0.5 tests were passing for this PR before I merged it but maybe I was mistaken? Now of master when I run `./runtests transforms` I see

```
Running test_transforms.jl
Test Failed
  Expression: param ≈ new_param
   Evaluated: [0.1,0.9] isapprox [NaN,NaN]
	FAILED: test_transforms.jl
ERROR: LoadError: LoadError: There was an error during testing
 in record(::Base.Test.FallbackTestSet, ::Base.Test.Fail) at ./test.jl:397
 in do_test(::Base.Test.Returned, ::Expr) at ./test.jl:281
 in (::#simplex_and_unsimplex#170)(::Array{Float64,1}, ::Celeste.Transform.SimplexBox) at /home/jeff/git/Celeste/test/test_transforms.jl:104
 in test_transform_simplex_functions() at /home/jeff/git/Celeste/test/test_transforms.jl:114
 in include_from_node1(::String) at ./loading.jl:488
 in macro expansion; at ./util.jl:184 [inlined]
 in macro expansion; at /home/jeff/git/Celeste/test/runtests.jl:74 [inlined]
 in anonymous at ./<missing>:?
 in include_from_node1(::String) at ./loading.jl:488
 in process_options(::Base.JLOptions) at ./client.jl:262
 in _start() at ./client.jl:318
while loading /home/jeff/git/Celeste/test/test_transforms.jl, in expression starting on line 267
while loading /home/jeff/git/Celeste/test/runtests.jl, in expression starting on line 71
```

andreasnoack:
I mentioned that in the top post, https://github.com/jeff-regier/Celeste.jl/pull/552#issue-204374695, and asked you what the right thing to do would be. It seems like a pathological case that we don't need to handle.

jeff-regier:
Ah, sorry, my mistake, I'll go ahead and remove the `Inf` checks.

PR #555, opened 2017-02-03
restore elbo support for dual numbers as argument

jeff-regier:
@jrevels I was trying to compare the runtimes for dual numbers vs Float64, to a get a sense of the overhead, but I'm getting an error from the KL divergence code when I try calling it with ForwardDiff's dual numbers. Any idea what's happening? It's at elbo_kl.jl:187. The test I wrote to recreate this error is part of this branch but not master.

Thanks!

```
jeff@dean ~/git/Celeste/test $ ./runtests.jl elbo
make: Nothing to be done for 'all'.
make: Nothing to be done for 'all'.
make: Nothing to be done for 'fetch'.
WARNING: Skipping long running tests.  To test everything, run tests with the flag --long-running
Running test_elbo.jl
  2.129635 seconds (1.97 M allocations: 76.162 MB, 0.78% gc time)
	FAILED: test_elbo.jl
ERROR: LoadError: LoadError: MethodError: Cannot `convert` an object of type ForwardDiff.Dual{88,Float64} to an object of type Float64
This may have arisen from a call to the constructor Float64(...),
since type constructors fall back to convert methods.
 in copy!(::Base.LinearFast, ::Array{Float64,1}, ::Base.LinearFast, ::Array{ForwardDiff.Dual{88,Float64},1}) at ./abstractarray.jl:559
 in value! at /home/jeff/.julia/v0.5/ReverseDiff/src/tracked.jl:146 [inlined]
 in seeded_forward_pass! at /home/jeff/.julia/v0.5/ReverseDiff/src/api/tape.jl:35 [inlined]
 in gradient!(::DiffBase.DiffResult{1,ForwardDiff.Dual{88,Float64},Tuple{Array{ForwardDiff.Dual{88,Float64},2}}}, ::ReverseDiff.Compiled{ReverseDiff.GradientTape{Celeste.DeterministicVI.KLDivergence.#subtract_kl,ReverseDiff.TrackedArray{Float64,Float64,1,Array{Float64,1},Array{Float64,1}},ReverseDiff.TrackedReal{Float64,Float64,Void}},Celeste.DeterministicVI.KLDivergence.#subtract_kl,ReverseDiff.TrackedArray{Float64,Float64,1,Array{Float64,1},Array{Float64,1}},ReverseDiff.TrackedReal{Float64,Float64,Void},##1#2,##3#4}, ::Array{ForwardDiff.Dual{88,Float64},1}) at /home/jeff/.julia/v0.5/ReverseDiff/src/api/gradients.jl:79
 in subtract_kl_source!(::Celeste.SensitiveFloats.SensitiveFloat{ForwardDiff.Dual{88,Float64}}, ::DiffBase.DiffResult{1,ForwardDiff.Dual{88,Float64},Tuple{Array{ForwardDiff.Dual{88,Float64},2}}}, ::Array{ForwardDiff.Dual{88,Float64},1}, ::Celeste.DeterministicVI.KLDivergence.KLHelper{ReverseDiff.##301#302{ReverseDiff.Compiled{ReverseDiff.GradientTape{Celeste.DeterministicVI.KLDivergence.#subtract_kl,ReverseDiff.TrackedArray{Float64,Float64,1,Array{Float64,1},Array{Float64,1}},ReverseDiff.TrackedReal{Float64,Float64,Void}},Celeste.DeterministicVI.KLDivergence.#subtract_kl,ReverseDiff.TrackedArray{Float64,Float64,1,Array{Float64,1},Array{Float64,1}},ReverseDiff.TrackedReal{Float64,Float64,Void},##1#2,##3#4}},Celeste.DeterministicVI.KLDivergence.##4#6{ForwardDiff.JacobianConfig{9,Float64,Array{ForwardDiff.Dual{9,Float64},1}},Celeste.DeterministicVI.KLDivergence.##3#5{Array{ForwardDiff.Dual{9,Float64},1},ReverseDiff.##301#302{ReverseDiff.Compiled{ReverseDiff.GradientTape{Celeste.DeterministicVI.KLDivergence.#subtract_kl,ReverseDiff.TrackedArray{ForwardDiff.Dual{9,Float64},ForwardDiff.Dual{9,Float64},1,Array{ForwardDiff.Dual{9,Float64},1},Array{ForwardDiff.Dual{9,Float64},1}},ReverseDiff.TrackedReal{ForwardDiff.Dual{9,Float64},ForwardDiff.Dual{9,Float64},Void}},Celeste.DeterministicVI.KLDivergence.#subtract_kl,ReverseDiff.TrackedArray{ForwardDiff.Dual{9,Float64},ForwardDiff.Dual{9,Float64},1,Array{ForwardDiff.Dual{9,Float64},1},Array{ForwardDiff.Dual{9,Float64},1}},ReverseDiff.TrackedReal{ForwardDiff.Dual{9,Float64},ForwardDiff.Dual{9,Float64},Void},##5#6,##7#8}}}}}) at /home/jeff/.julia/v0.5/Celeste/src/deterministic_vi/elbo_kl.jl:187
```

### DIFF
--- a/test/test_elbo.jl
+++ b/test/test_elbo.jl
@@ -4,8 +4,10 @@ using Base.Test
 using Distributions
 using DerivativeTestUtils
 using StaticArrays
-using ForwardDiff
 
+using ForwardDiff
+import ForwardDiff.Dual
+import DeterministicVI.KLDivergence.KLHelper
 
 import SampleData: gen_two_body_dataset, true_star_init
 
@@ -326,8 +328,8 @@ function test_elbo_supports_dual_numbers()
     # the hessian (it doesn't matter that the "perturbation" are all zero, it's just
     # for testing the speed and verifying that it works)
     P = length(ea0.vp[1])
+
     # `1` "perterbation" per dual number is enough for a hessian-vector mulitiply,
-    # I think
     T = ForwardDiff.Dual{1, Float64}
     vp = Vector{T}[zeros(T, P) for s=1:ea0.S]
     for s=1:ea0.S
@@ -336,8 +338,11 @@ function test_elbo_supports_dual_numbers()
     ea1 = ElboArgs(ea0.images, vp, ea0.patches, ea0.active_sources, calculate_hessian=false)
 
     # evaluate the elbo for both argument types
-    @time DeterministicVI.elbo(ea0)
-    @time DeterministicVI.elbo(ea1)
+    DeterministicVI.elbo(ea0)
+
+    chunksize = ForwardDiff.pickchunksize(length(Model.ids))
+    kl_helper = KLHelper(Dual{chunksize, Dual{1, Float64}})
+    DeterministicVI.elbo(ea1; kl_helper=kl_helper)
 end
 
 
--- a/test/test_elbo.jl
+++ b/test/test_elbo.jl
@@ -326,7 +326,9 @@ function test_elbo_supports_dual_numbers()
     # the hessian (it doesn't matter that the "perturbation" are all zero, it's just
     # for testing the speed and verifying that it works)
     P = length(ea0.vp[1])
-    T = ForwardDiff.Dual{P * ea0.S, Float64}
+    # `1` "perterbation" per dual number is enough for a hessian-vector mulitiply,
+    # I think
+    T = ForwardDiff.Dual{1, Float64}
     vp = Vector{T}[zeros(T, P) for s=1:ea0.S]
     for s=1:ea0.S
         vp[s][:] = ea0.vp[s][:]
--- a/test/test_elbo.jl
+++ b/test/test_elbo.jl
@@ -12,6 +12,10 @@ import DeterministicVI.KLDivergence.KLHelper
 import SampleData: gen_two_body_dataset, true_star_init
 
 
+chunksize = ForwardDiff.pickchunksize(length(Model.ids))
+kl_helper = KLHelper(Dual{chunksize, Dual{1, Float64}})
+
+
 function test_set_hess()
     sf = SensitiveFloat{Float64}(length(ids), 1, true, true)
     set_hess!(sf, 2, 3, 5.0)
@@ -339,9 +343,6 @@ function test_elbo_supports_dual_numbers()
 
     # evaluate the elbo for both argument types
     DeterministicVI.elbo(ea0)
-
-    chunksize = ForwardDiff.pickchunksize(length(Model.ids))
-    kl_helper = KLHelper(Dual{chunksize, Dual{1, Float64}})
     DeterministicVI.elbo(ea1; kl_helper=kl_helper)
 end
 
--- a/test/test_elbo.jl
+++ b/test/test_elbo.jl
@@ -4,6 +4,8 @@ using Base.Test
 using Distributions
 using DerivativeTestUtils
 using StaticArrays
+using ForwardDiff
+
 
 import SampleData: gen_two_body_dataset, true_star_init
 
@@ -314,6 +316,30 @@ function test_num_allowed_sd()
 end
 
 
+function test_elbo_supports_dual_numbers()
+    _, ea, _ = gen_two_body_dataset()
+
+    # create elbo arguments of type Float64 that compute the gradient but not the hessian
+    ea0 = ElboArgs(ea.images, ea.vp, ea.patches, ea.active_sources, calculate_hessian=false)
+
+    # create elbo arguments of the Dual number type, that compute the gradient but not
+    # the hessian (it doesn't matter that the "perturbation" are all zero, it's just
+    # for testing the speed and verifying that it works)
+    P = length(ea0.vp[1])
+    T = ForwardDiff.Dual{P * ea0.S, Float64}
+    vp = Vector{T}[zeros(T, P) for s=1:ea0.S]
+    for s=1:ea0.S
+        vp[s][:] = ea0.vp[s][:]
+    end
+    ea1 = ElboArgs(ea0.images, vp, ea0.patches, ea0.active_sources, calculate_hessian=false)
+
+    # evaluate the elbo for both argument types
+    @time DeterministicVI.elbo(ea0)
+    @time DeterministicVI.elbo(ea1)
+end
+
+
+test_elbo_supports_dual_numbers()
 test_active_sources()
 test_set_hess()
 test_bvn_cov()
--- a/test/test_elbo.jl
+++ b/test/test_elbo.jl
@@ -7,13 +7,8 @@ using StaticArrays
 
 using ForwardDiff
 import ForwardDiff.Dual
-import DeterministicVI.KLDivergence.KLHelper
 
 import SampleData: gen_two_body_dataset, true_star_init
-
-
-chunksize = ForwardDiff.pickchunksize(length(Model.ids))
-kl_helper = KLHelper(Dual{chunksize, Dual{1, Float64}})
 
 
 function test_set_hess()
@@ -343,7 +338,7 @@ function test_elbo_supports_dual_numbers()
 
     # evaluate the elbo for both argument types
     DeterministicVI.elbo(ea0)
-    DeterministicVI.elbo(ea1; kl_helper=kl_helper)
+    DeterministicVI.elbo(ea1)
 end
 
 
--- a/src/SensitiveFloats.jl
+++ b/src/SensitiveFloats.jl
@@ -141,10 +141,16 @@ function combine_sfs!{T1 <: Number, T2 <: Number, T3 <: Number}(
     end
 
     if sf_result.has_gradient
+        for ind in eachindex(sf_result.d)
+            sf_result.d[ind] = g_d[1] * sf1.d[ind] + g_d[2] * sf2.d[ind]
+        end
+#=
         sf_result.d[:] = sf1.d
         n = length(sf_result.d)
+        @show (n, g_d[1], sf_result.d, 1)
         LinAlg.BLAS.scal!(n, g_d[1], sf_result.d, 1)
         LinAlg.BLAS.axpy!(g_d[2], sf2.d, sf_result.d)
+=#
     end
 
     sf_result.v[] = v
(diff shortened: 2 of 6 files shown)


jrevels:
The KL divergence's derivative code only accepts `Float64` numbers (or numbers that can be converted to `Float64`). This is because I only precompiled the ReverseDiff tape there for `Float64` inputs in mind ([here's the ReverseDiff tape documentation](http://www.juliadiff.org/ReverseDiff.jl/api/#the-abstracttape-api)).

test/test_elbo.jl
```diff
@@ -314,6 +316,30 @@ function test_num_allowed_sd()
...
+    _, ea, _ = gen_two_body_dataset()
+
+    # create elbo arguments of type Float64 that compute the gradient but not the hessian
+    ea0 = ElboArgs(ea.images, ea.vp, ea.patches, ea.active_sources, calculate_hessian=false)
+
+    # create elbo arguments of the Dual number type, that compute the gradient but not
+    # the hessian (it doesn't matter that the "perturbation" are all zero, it's just
+    # for testing the speed and verifying that it works)
+    P = length(ea0.vp[1])
+    T = ForwardDiff.Dual{P * ea0.S, Float64}
```
jrevels: This is a very bad idea. It's type unstable, but more importantly, it could lead to very high chunk sizes, which could kill performance. For example, in the stacktrace in the OP, it seems you are constructing `Dual{88,Float64}`. That's going to absolutely thrash the stack. If you need to take that many simultaneous partial derivatives, you need to use chunking (which ForwardDiff does automatically) or use reverse-mode AD.

jrevels:
To expand on the taping thing: if you want this to play around with this for benchmarking purposes, it should be enough just choose the appropriate element type when allocating the [`KLHelper`](https://github.com/jeff-regier/Celeste.jl/blob/master/src/deterministic_vi/elbo_kl.jl#L170). As you can see, the current element type is `Dual{N,Float64}` (where `N` is determined by `ForwardDiff.pickchunksize( length(Celeste.Model.ids))`), so you'd probably want it to be `Dual{N,Dual{1,Float64}}`.

jeff-regier:
Much better with `88` changed to `1`. We've still got values like 88 elsewhere in the unit tests--that explains why our derivative unit tests take so long!

I see very little overhead now from calling our objective function (excluding the kl code) with dual numbers rather than floats. Amazing! That's SIMD I take it? It's like getting our first hessian-vector product for free.

```
julia> @time DeterministicVI.elbo_likelihood(ea0);  # Float64 numbers
  0.022719 seconds (51.60 k allocations: 1.636 MB)

julia> @time DeterministicVI.elbo_likelihood(ea1);  # dual numbers
  0.027346 seconds (193.08 k allocations: 6.366 MB)
```

jrevels:
Woohoo! 

I do wonder where the extra allocations/memory usage is coming from, I wouldn't have expected it to increase by that much (if at all). Looks like a symptom of a typing problem somewhere.

> That's SIMD I take it?

I usually confirm the presence of SIMD instructions in intermediate functions by looking at the output of `@code_native` in top-level scope but IIUC, these kinds of top-level checks aren't necessarily indicative of what's happening in the actual calling context where these function calls can get inlined.

Maybe @Keno or @kpamnany knows the right steps to confirm SIMD usage in real-world contexts?

jeff-regier:
Is there a way to change the default `kl_helper` [here](https://github.com/jeff-regier/Celeste.jl/blob/master/src/deterministic_vi/elbo_objective.jl#L540) to work with parameters of type `T`? Usually `T` is `Float64`, but in this case, `T` is `ForwardDiff.Dual`.

andreasnoack:
We can see in the SDE output if the wide instructions are used. Does any of this change `benchmark_one_light_source`? Should I rerun the flame graph and FLOP count?

jrevels:
> Is there a way to change the default kl_helper here to work with parameters of type T? Usually T is Float64, but in this case, T is ForwardDiff.Dual.

You can just preallocate whatever `KLHelper`s you want using the constructor I linked earlier, and use dispatch to switch between them.

jeff-regier:
I'm making my own `kl_helper` now---it's chocked-full of dual variables. I passing it to `elbo` as a keyword argument, but I get roughly the same error. Any advice?

```
	FAILED: test_elbo.jl
ERROR: LoadError: LoadError: MethodError: Cannot `convert` an object of type ForwardDiff.Dual{88,Float64} to an object of type Float64
This may have arisen from a call to the constructor Float64(...),
since type constructors fall back to convert methods.
 in copy!(::Base.LinearFast, ::Array{Float64,1}, ::Base.LinearFast, ::Array{ForwardDiff.Dual{88,Float64},1}) at ./abstractarray.jl:559
 in value! at /home/jeff/.julia/v0.5/ReverseDiff/src/tracked.jl:146 [inlined]
 in seeded_forward_pass! at /home/jeff/.julia/v0.5/ReverseDiff/src/api/tape.jl:35 [inlined]
 in gradient!(::DiffBase.DiffResult{1,ForwardDiff.Dual{88,Float64},Tuple{Array{ForwardDiff.Dual{88,Float64},2}}}, ::ReverseDiff.Compiled{ReverseDiff.GradientTape{Celeste.DeterministicVI.KLDivergence.#subtract_kl,ReverseDiff.TrackedArray{Float64,Float64,1,Array{Float64,1},Array{Float64,1}},ReverseDiff.TrackedReal{Float64,Float64,Void}},Celeste.DeterministicVI.KLDivergence.#subtract_kl,ReverseDiff.TrackedArray{Float64,Float64,1,Array{Float64,1},Array{Float64,1}},ReverseDiff.TrackedReal{Float64,Float64,Void},##1#2,##3#4}, ::Array{ForwardDiff.Dual{88,Float64},1}) at /home/jeff/.julia/v0.5/ReverseDiff/src/api/gradients.jl:79
 in subtract_kl_source!(::Celeste.SensitiveFloats.SensitiveFloat{ForwardDiff.Dual{88,Float64}}, ::DiffBase.DiffResult{1,ForwardDiff.Dual{88,Float64},Tuple{Array{ForwardDiff.Dual{88,Float64},2}}}, ::Array{ForwardDiff.Dual{88,Float64},1}, ::Celeste.DeterministicVI.KLDivergence.KLHelper{ReverseDiff.##301#302{ReverseDiff.Compiled{ReverseDiff.GradientTape{Celeste.DeterministicVI.KLDivergence.#subtract_kl,ReverseDiff.TrackedArray{Float64,Float64,1,Array{Float64,1},Array{Float64,1}},ReverseDiff.TrackedReal{Float64,Float64,Void}},Celeste.DeterministicVI.KLDivergence.#subtract_kl,ReverseDiff.TrackedArray{Float64,Float64,1,Array{Float64,1},Array{Float64,1}},ReverseDiff.TrackedReal{Float64,Float64,Void},##1#2,##3#4}},Celeste.DeterministicVI.KLDivergence.##4#6{ForwardDiff.JacobianConfig{9,Float64,Array{ForwardDiff.Dual{9,Float64},1}},Celeste.DeterministicVI.KLDivergence.##3#5{Array{ForwardDiff.Dual{9,Float64},1},ReverseDiff.##301#302{ReverseDiff.Compiled{ReverseDiff.GradientTape{Celeste.DeterministicVI.KLDivergence.#subtract_kl,ReverseDiff.TrackedArray{ForwardDiff.Dual{9,Float64},ForwardDiff.Dual{9,Float64},1,Array{ForwardDiff.Dual{9,Float64},1},Array{ForwardDiff.Dual{9,Float64},1}},ReverseDiff.TrackedReal{ForwardDiff.Dual{9,Float64},ForwardDiff.Dual{9,Float64},Void}},Celeste.DeterministicVI.KLDivergence.#subtract_kl,ReverseDiff.TrackedArray{ForwardDiff.Dual{9,Float64},ForwardDiff.Dual{9,Float64},1,Array{ForwardDiff.Dual{9,Float64},1},Array{ForwardDiff.Dual{9,Float64},1}},ReverseDiff.TrackedReal{ForwardDiff.Dual{9,Float64},ForwardDiff.Dual{9,Float64},Void},##5#6,##7#8}}}}}) at /home/jeff/.julia/v0.5/Celeste/src/deterministic_vi/elbo_kl.jl:187
```

test/test_elbo.jl
```diff
@@ -314,6 +318,35 @@ function test_num_allowed_sd()
...
+    for s=1:ea0.S
+        vp[s][:] = ea0.vp[s][:]
+    end
+    ea1 = ElboArgs(ea0.images, vp, ea0.patches, ea0.active_sources, calculate_hessian=false)
+
+    # evaluate the elbo for both argument types
+    DeterministicVI.elbo(ea0)
+
+    chunksize = ForwardDiff.pickchunksize(length(Model.ids))
+    kl_helper = KLHelper(Dual{chunksize, Dual{1, Float64}})
```
jrevels: You shouldn't allocate `KLHelper`s inside of other functions, only in top-level scope. For example, I'm pretty sure this can never pass on v0.6 since the world-ages of the functions created by the `KLHelper` are too young compared to the world-age of the calling scope.

jeff-regier:
Moving the `kl_helper` creation to global scope doesn't seem to have changed the error. I don't really get why kl_helper can't be a local variable, unfortunately.

jrevels:
Yeah, that recommendation didn't have anything to do with the error you're seeing, it was just something I noticed. I actually don't see why the error you posted would happen...are you sure you're not calling old code? I'm trying to look into it now, but I'm getting a different error.

> I don't really get why kl_helper can't be a local variable, unfortunately.

The `KLHelper` constructor uses ReverseDiff's tape compilation API to generate gradient functions. Compiling gradients from instruction tapes pretty fundamentally requires `eval`, as it generates compiled code from runtime values. Thus, one should only ever apply this feature at top-level (like any other use of `eval`). This is enforced by the compiler on v0.6 through the world-age mechanism, and in turn, enforces good style for the ReverseDiff compilation functions. Eventually I'm going to write a centralized section in the ReverseDiff docs about this (the info is actually there currently, but is scattered about and thus probably harder to grok than it should be).

Note that you don't need to compile the instruction tapes to take derivatives with ReverseDiff, it's just an extra feature we can take advantage of Celeste's case, since the functions have the right properties to benefit from it.

PR #556, opened 2017-02-03
pick proper KLHelper automatically via dispatch

jrevels:
@jeff-regier This is what I meant about picking the correct KLHelper automatically via dispatch. I've opened this PR against your branch so you can include it if you like (I didn't address the error you're seeing though, haven't been able to reproduce it yet)


jrevels:
@jeff-regier I pushed a fix for a type genericity bug in the `KLHelper` constructor here, this fixes the error I was getting and now `DeterministicVI.elbo(ea1)` works.

jeff-regier:
Thanks a lot Jarrett! This fixes the problem I was seeing. I like how `elbo` can handle either type automatically now.

PR #563, opened 2017-02-16
excessive neighbors

### LINKED ISSUE #562: excessive neighbors: 309.0 309.5 45.5 46.0

jeff-regier:
We were finding an excessive number of neighbors because our distance calculation made no sense. First, it was computing taking the `cos` of degrees---but `cos` in Julia an argument in radians. Second, there's no reason to take that cosine anyway, as as I can tell. Pixel scale is 0.396 arc seconds everywhere in SDSS, regardless of the declination. (@rcthomas Is that correct? Is there any reason to scale the difference in RA by the cosine of the DEC?)

@gostevehoward  Heads up! I think this "distance" formula shows up in `Stripe82Score.jl` too--you were wondering about it, I seem to recall. It probably doesn't throw things off as much there because the stripe 82 declinations are all around 0, and `cos(0)` is about 1, but still, wow, that's a serious bug.

Fixes #562 .

### DIFF
--- a/src/DeterministicVI.jl
+++ b/src/DeterministicVI.jl
@@ -114,7 +114,9 @@ function infer_source(images::Vector{Image},
                       entry::CatalogEntry;
                       min_radius_pix=Nullable{Float64}())
     if length(neighbors) > 100
-        Log.warn("Excessive number ($(length(neighbors))) of neighbors")
+        msg = string("objid $(entry.objid) [ra: $(entry.pos)] has an excessive",
+                     "number ($(length(neighbors))) of neighbors")
+        Log.warn(msg)
     end
 
     # It's a bit inefficient to call the next 5 lines every time we optimize_f.
--- a/src/Infer.jl
+++ b/src/Infer.jl
@@ -50,9 +50,7 @@ function find_neighbors(target_sources::Vector{Int64},
         @assert radii_map[s] <= 25
     end
 
-    # compute distance in pixels using small-distance approximation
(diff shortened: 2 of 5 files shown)


gostevehoward:
This is the distance calculation I should use instead?

dist(ra1, dec1, ra2, dec2) = (3600 / 0.396) * max(abs(dec2 - dec1), abs(ra2
- ra1))

jeff-regier:
Euclidian distance (just dropping the `cos(dec2-dec1)`) might be even better than that for matching across catalogs, but either should be fine.

rcthomas:
@jeff-regier @gostevehoward Be very careful about dropping the cos(dec).  A trick to avoid cos altogether might be to compare squared 3D Euclidean distances.

jeff-regier:
Antarctica gets elongated on a 2D map, and the cos of the latitude/declination adjusts for that. I'm unclear about whether that kind of adjustment is relevant here. 1) We're concerned with how far apart lights sources appear in images, not how far they are if projected onto a hollow sphere. 2) Each image is only a fraction of a square degree.

If cosine is appropriate, I'm pretty sure it should be `cosd` (degrees) though rather than `cos` (radians).

rcthomas:
... distance in RA depends on cos(Dec); right?

jeff-regier:
Are pixels everywhere in SDSS nearly 0.396 arc second x 0.396 arc seconds ? Or does that depend on the declination?

PR #570, opened 2017-02-17
`cos` -> `cosd` + unit test

jeff-regier:
@gostevehoward @rcthomas 

So now I think that the distance formula with `cos` changed to `cosd` is correct. Pixel-scale must only refer to the scale of the pixels at declination 0. It isn't constant across declinations. I'd appreciate you two look over my unit test to confirm my latest thinking.

### DIFF
--- a/test/test_score.jl
+++ b/test/test_score.jl
@@ -1,7 +1,34 @@
 import Celeste.Stripe82Score
+import FITSIO
+import WCS
 
 
-function test_score_field()
+@testset "test the distance calculation" begin
+    wd = pwd()
+    cd(datadir)
+    # this field is pretty near a pole--high declination means fewer pixels
+    # per arc second of right ascention
+    run(`make RUN=6075 CAMCOL=2 FIELD=29`)
+    high_fits = FITSIO.FITS("6075/2/29/frame-r-006075-2-0029.fits")
+    cd(wd)
+
+    # we're just loaded this image to use its wcs transform
+    high_wcs = WCS.from_header(FITSIO.read_header(high_fits[1], String))[1]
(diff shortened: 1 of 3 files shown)


test/test_score.jl
```diff
@@ -1,7 +1,34 @@
...
+    pt0_pix = [10; 10.]
+    pt0 = WCS.pix_to_world(high_wcs, pt0_pix)
+    # pt1 is 1 arc second to the right of pt0
+    pt1 = pt0 + [1. / 3600, 0]
+
+    exact_dist = norm(WCS.world_to_pix(high_wcs, pt1) - WCS.world_to_pix(high_wcs, pt0))
+    our_dist = Stripe82Score.dist(pt0..., pt1...)
+
+    # both distances are about 0.315 pixels -- an arc second shift to the right
+    # is less, in pixels, at higher elevations than at the equator, where it's
```
jeff-regier, line 26: @rcthomas Is declination 0 the equator (or, more precisely, the plane that contains the equator)? I assumed it was in this comment.

rcthomas: Dec = 0 at equator, +90 at north pole, -90 at south yes.

gostevehoward (approved):
Seems good to me. Does the test fail with `cos` in place of `cosd`?

jeff-regier:
Good point, I reordered the commits--first unit test, then bug fix--to show it fails without the bug fix.

Btw, there may be some other places where we're converting from arc seconds to pixel scale without accounting for the declination. For computing `gal_scale`, for example, in `Stripe82Score.jl`. Because Stripe 82 is near the equator (where `cos(dec)` is about 1) we've probably been able to get away with it so far.

PR #572, opened 2017-02-20
unit tests and a bug fix

jeff-regier:
This PR

- restores a unit test (`test_e_g_s_functions`)
- improves a test set (`Tests elbo gradient, and that elbo supports dual numbers`)
- fixes a bug that was keeping the derivative one place from getting computed if the hessian wasn't also getting computed

These commits are cherry picked from my `jcr/cg` branch. I figured we should make sure these tests pass on the `jr/constraints` branch too before merge, though I don't believe they'll present any trouble.

### DIFF
--- a/src/DeterministicVI.jl
+++ b/src/DeterministicVI.jl
@@ -17,6 +17,8 @@ using ..Transform
 import DataFrames
 import Optim
 using StaticArrays
+import ForwardDiff.Dual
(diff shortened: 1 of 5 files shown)


jeff-regier:
Nevermind, this creates a small merge conflict with the `jr/constraints` branch. Let's just get that one committed first.

PR #573, opened 2017-02-23
remove old maximize_elbo.jl code (follow-up to #519)

jrevels:


jeff-regier:
Looks good! We're stuck with the PSF.jl code now for awhile since the FFT
approach hasn't panned out, I didn't plan on that originally. Do you think
it's worth revising PSF.jl to use `ConstraintTransforms` rather
`Transforms`, so we can delete `Transforms`?

jrevels:
> Do you think it's worth revising PSF.jl to use `ConstraintTransforms` rather `Transforms`, so we can delete `Transforms`?

If PSF isn't going away, than yeah, it probably should be. I'm not going to have time do so before the hackathon, though...maybe that can be my hackathon activity?

On the subject of this PR:

The `test_galaxy_optimization_fft` test is failing because the evaluation of the objective (`elbo_fft_opt`) on the given value (`ea_fft`) is returning a SensitiveFloat containing `NaN`s. This in turn causes an indexing error in Optim when it tries to access the result of an `eigfact` on the Hessian ([relevant issue can be found here](https://github.com/JuliaLang/LinearAlgebra.jl/issues/406)).

I'll keep looking into it myself, but in case anybody knows what might be going on, I wrote up [a gist](https://gist.github.com/jrevels/e6a639024d3abd27660cd4441722acda) which demonstrates the values I'm seeing. It shows that there aren't any `NaN`s in either the bound/free parameterizations of the inputs, and that the input vector hardly differs from the initial values, yet the objective function is producing a SensitiveFloat containing `NaN`s. 

The weird part is that if I manually recreate the steps taken in `evaluate!` with the values I printed out, there aren't any `NaN`s in the output. That might be an artifact of the printed values not being exactly the same, but that would imply pretty drastic sensitivity on the part of objective function, wouldn't it?

jeff-regier:
It might make sense to move the FFT code to a branch rather than keeping it
in master. That needs a lot of work to run efficiently, and we're not using
it for GB now. In the meantime, if you're pretty sure the problem isn't
introduced by your changes and the issue only involves the FFT code, you
could just use the `@test_skip` macro. And maybe file an issue with the
problematic input so we can get to it in the future? Don't want you to get
hung up on some nan issue from code we aren't using for GB. Thanks for
finding the problem.

src/deterministic_vi/NewtonMaximize.jl
```diff
@@ -169,4 +176,42 @@ function from_vector!(sources, x)
...
+            ParameterConstraint(SimplexConstraint(0.01/D, 1.0, D), ids.k[:, 1])
+        ]
+    end
+    return Config(ea; constraints = ConstraintBatch(boxes, simplexes), kwargs...)
+end
+
+function maximize_two_steps!{F,T}(f::F, ea::ElboArgs{T},
+                                  cfg_star::Config = star_only_config(ea),
+                                  cfg_both::Config = Config(ea))
+    ea.vp[1][ids.a] = [1, 0]
```
jrevels, line 209: @jeff-regier Sounds good, though I think I may have figured out the problem.

Initializing `ea.vp[1][ids.a] = [1, 0]` here, combined with not placing a `SimplexConstraint` on `a` (in `star_only_config` above), is the cause of the `NaN`s. This problem occurs for both of our objective function implementations (`DeterministicVI.elbo` and the FFT version).

The reason I excluded the `a` constraint from `star_only_config` is because the original code did so (i.e., the `a` indices were in the `star_omitted_ids`). Adding  the `a` constraint back (or not initializing it to `[1, 0]` in the first place) allows the optimization to run without encountering `NaN`s and allows the optimization tests to pass. However, this means that the new code doesn't exactly match the behavior of the old code. I'm not sure if that's significant or not.

I've just pushed a commit which adds the `a` constraint; hopefully tests should pass now. If the behavioral change caused by adding the new constraint is significant, let me know and I'll revert that change and just disable the test.

Issue #510, opened 2017-01-04
unify `GalsimBenchmark` and `Stripe82Score`


jeff-regier:
Maybe call the unified module something like `AccuracyBenchmark`.

We'd like to be able to pick arbitrary combinations of 1) a ground truth catalog (either `coadd`, or a synthetic catalog drawn from the Celeste model), 2) an image generation routine (either galsim, `test/Synthetic.jl` or SDSS), and 3) prediction catalog (`primary`, Celeste MOG, or Celeste FFT).

It's tempting to make the synthetic catalog have the same format as `coadd`, so the script doesn't have to "know" the source of the ground truth, and thus can't possibly treat `coadd` and the synthetic catalog differently. But I don't think that's the way to go, because the `coadd` catalog has a lot of quirks.

jeff-regier:
@gostevehoward  We could go over the `Stripe82Score.jl` code tomorrow before or after the DESI meeting if you like. I'm sure you could figure it out too. The basic idea is that there's a function to load each type of catalog (i.e., primary, coadd, celeste) into a data frame having a common format. These functions are called `load_primary`, `load_coadd`, and `celeste_to_df`. (The first two read from disk, the third takes a vector of results read from disk---for better or worse, that's why the naming convention is slightly different for `celeste_to_df`.) Because the loaded data frames have a common format, any pair can be compared by calling `get_err_df`. You'd presumably create a new one of these functions for the galsim catalog, something like `gimsim_to_df`, that calls `load_ce!` (`ce` stands for `CatalogEntry`) if your galsim catalog is stored as a vector of `CatalogEntry` instances.

You'll notice `load_primary` and `load_coadd` return data frames with a lot of `NA` values. That's because there's no obvious mapping from those catalogs to the common data frame format for some fields, for some types of light sources. You wouldn't need to worry about `NA` values through for the galsim catalog, I think.

gostevehoward:
Here are the pieces that need to be put together, I think:

- [x] `load_coadd_catalog(filename) -> DF` (already done in [load_s82](https://github.com/jeff-regier/Celeste.jl/blob/master/src/Stripe82Score.jl#L93), I've made some refactorings). The DF this already returns will become the common catalog DF format throughout this benchmark module.
[...]

PR #574, opened 2017-02-23
A generalized framework for accuracy benchmarks

gostevehoward:
This is kind of an unsatisfying PR, because

* `Stripe82Score.jl` and `validate_on_stripe82.jl` can't be deleted yet, because the new versions aren't quite working the same yet, and we can't abandon the old files until the new implementation is satisfactory
* The output of `GalsimBenchmark.jl` is a bit less useful than it was before (but that will change), and I had to comment out some test cases (temporarily), something I swore I wouldn't do anymore, and
* We don't have support yet for synthetic imagery from the Celeste likelihood, one of the main goals.

But I wanted to get a PR out there because it's already been way too long and there are many lines of diff. And the new implementation *is* a lot more flexible than the old one, and hopefully cleaner.

The [wiki page](https://github.com/jeff-regier/Celeste.jl/wiki/Accuracy-benchmarks) should give some context, but feel free to ask questions or request more context.

This implements most of #510.


jeff-regier:
LGTM. Nice to see it coming together. I appreciate the wiki documentation. Please go ahead and merge this PR whenever the Travis tests are passing.

gostevehoward:
For what it's worth, I think things are OK now but data.sdss3.org seems to be down so the travis tests can't run :-/. I'll wait for it to come back up.

PR #579, opened 2017-03-02
WIP: enable "Hessian-free" optimization

jeff-regier:


src/deterministic_vi/NewtonMaximize.jl
```diff
@@ -28,79 +32,101 @@ immutable Config{N,T}
...
+    elbo_vars = ElboIntermediateVariables(Dual{1, Float64}, ea.S, length(ea.active_sources); 
+                                          calculate_gradient=true, 
+                                          calculate_hessian=false)
+    sf_free = SensitiveFloat{Dual{1, Float64}}(length(cfg.free_params[1]), 
+                                    length(cfg.bound_params), true, false)
+    return Hessian(ea, elbo_vars, sf_free, cfg)
 end
 
 function evaluate!(f::Objective, x::Vector)
-    if x != f.previous_x
```
jrevels: Is the derivative caching here being handled somewhere else (e.g. maybe Optim is doing it for us now), or perhaps the new method here doesn't benefit from the caching? Otherwise, removing this could cause a lot of redundant calculation across repeated calls of the other oracles with the same input.

jeff-regier: Yes, that's it, the CG trust region code (not yet added to Optim) pretty much avoids calling the objective function twice with the same argument.

Keno:
Not to be be discouraging here in any way, but before merging this, we should do some careful performance measurements to make sure we don't lose any of the careful optimizations done on the existing code.

PR #586, opened 2017-03-07
Cut down on some memory allocations

Keno:
I'm not saying this is the best way to do it, but it's one way to do it. Feel free to do it another way.


andreasnoack:
This is part of #590

PR #593, opened 2017-03-09
fix redundant ReverseDiff.compile calls from #587

jrevels:


jrevels:
Actually, this will be fixed by https://github.com/JuliaDiff/ReverseDiff.jl/pull/58 (which I'll tag ASAP once tests pass).

PR #594, opened 2017-03-09
bump ReverseDiff version

jrevels:
This supercedes https://github.com/jeff-regier/Celeste.jl/pull/593 (which is already closed).

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -6,7 +6,7 @@ Distributions
 FITSIO 0.8.4
 DiffBase 0.0.3
 ForwardDiff 0.3.4 0.5
-ReverseDiff 0.1.0 0.2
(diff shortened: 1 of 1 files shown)


jrevels:
Ah Optim upper bounds ForwardDiff, causing a version conflict here. I'll just bump ReverseDiff here, then, which should still fix the problem from #593, and open a PR in Optim to raise their ForwardDiff upper bound.

jrevels:
This is failing due to unrelated Rmath build errors. Looks like curl is getting back a server error (error code 500) when trying to download https://github.com/JuliaLang/Rmath-julia/archive/v0.1.tar.gz (if I go to that link in the browser, I also get a server error).

jrevels:
This passes locally, so I'm going to merge.

PR #596, opened 2017-03-10
Two fixes

Keno:
- Don't try to download image files when using the one source benchmark
  for precompilation.
- Fix an accidentally introduced memory allocation

### DIFF
--- a/src/bivariate_normals.jl
+++ b/src/bivariate_normals.jl
@@ -326,7 +326,7 @@ Note that nubar is not included.
 """
 function GalaxySigmaDerivs{NumType <: Number}(
     e_angle::NumType, e_axis::NumType, e_scale::NumType,
-    XiXi::SMatrix{2,2,NumType,4}, nuBar::NumType = one(NumType), calculate_tensor::Bool=true)
+    XiXi::SMatrix{2,2,NumType,4}, nuBar::Float64=1.0, calculate_tensor::Bool=true)
 
   cos_sin = cos(e_angle)sin(e_angle)
   sin_sq  = sin(e_angle)^2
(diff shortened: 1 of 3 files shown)


src/bivariate_normals.jl
```diff
@@ -326,7 +326,7 @@ Note that nubar is not included.
 """
 function GalaxySigmaDerivs{NumType <: Number}(
     e_angle::NumType, e_axis::NumType, e_scale::NumType,
-    XiXi::SMatrix{2,2,NumType,4}, nuBar::NumType = one(NumType), calculate_tensor::Bool=true)
+    XiXi::SMatrix{2,2,NumType,4}, nuBar::Float64=1.0, calculate_tensor::Bool=true)
```
andreasnoack, line 329: Are you sure this isn't too restrictive? Couldn't a `Dual` stop by here? Which definition caused the allocation?

Keno: As far as I can tell nuBar is always a Float64 (I had to fix the case you added, but in terms of the other use). Did you have a reason to think otherwise?

jeff-regier: `nuBar` can always be a float, we don't optimize that parameter.

andreasnoack: > Did you have a reason to think otherwise?

No. I was just curious what was causing the allocation and it wasn't obvious to me. I didn't think that it should matter if you used `one` or `1.0` here.

Keno: Ah that's not the issue. The compiler decided to insert an optimization break. That gets fixed by making the other signature more aggressive and preventing and error when NumType is dual (because nuBar was still Float64).

PR #600, opened 2017-03-10
split ea into ea, vp, and elbo_vars

jeff-regier:
This PR breaks out `vp` and `elbo_vars` from `ElboArgs`. That should let use reused `vp` and `elbo_vars` for all the light sources a single thread processes. That should let each thread fetch less memory, perhaps avoiding the memory bandwidth problems @kpamnany reports on knl with many processes running concurrently.


src/deterministic_vi/elbo_objective.jl
```diff
@@ -393,12 +385,13 @@ end
...
 function add_pixel_term!{NumType <: Number}(
-                    ea::ElboArgs{NumType},
+                    ea::ElboArgs,
+                    vp::VariationalParams{NumType},
                     n::Int, h::Int, w::Int,
                     star_mcs::Array{BvnComponent{NumType}, 2},
                     gal_mcs::Array{GalaxyCacheComponent{NumType}, 4},
-                    sbs::Vector{SourceBrightness{NumType}})
-    elbo_vars = ea.elbo_vars
+                    sbs::Vector{SourceBrightness{NumType}};
```
Keno: better to avoid keyword arguments (can still be an optional argument).

Keno (commented):
This is great, thanks!

PR #602, opened 2017-03-11
Remove old Stripe82 benchmark code

gostevehoward:
Now superseded by [accuracy benchmarks](https://github.com/jeff-regier/Celeste.jl/wiki/Accuracy-benchmarks)

We could keep the exist "test scoring a whole field" test by uploading a bunch of catalog CSVs to
portal.nersc, but I think it's probably more desirable to write a small suite of self-contained unit
tests for the AccuracyBenchmark module, which ought to be done anyway. (#601)


jeff-regier:
Are the tests passing locally Steve? It looks like sdss.org is down yet again. I wonder how they found such unreliable web hosting....

gostevehoward:
Ah bummer...yeah I got a pass locally: 
[tmux-main-0-0-20170311T105158.log.zip](https://github.com/jeff-regier/Celeste.jl/files/836048/tmux-main-0-0-20170311T105158.log.zip)

PR #608, opened 2017-03-12
fix Jacobian bounds at compile-time for ConstraintTransform derivatives

jrevels:
@keno let me know if this helps, if so we can merge it.

This PR hardcodes bounds for Jacobians propagated within ConstraintTransforms. This basically means ConstraintTransforms is now only usable for... 

- ...use with ElboMaximize where...
- ...the default `ConstraintBatch` is used, and...
- ...the number of bound parameters equals `length(CanonicalParams)`.

This is an okay cost to pay for now if it improves performance; the only other place we'll use ConstraintTransforms in the future is PSF, probably, at which point it will be much easier to re-generalize the code here if need be.

### DIFF
--- a/src/deterministic_vi/ConstraintTransforms.jl
+++ b/src/deterministic_vi/ConstraintTransforms.jl
@@ -3,6 +3,7 @@ module ConstraintTransforms
 using ..Model
 using ..SensitiveFloats
 using ..DeterministicVI: VariationalParams
+using StaticArrays
 using Compat
 
 ####################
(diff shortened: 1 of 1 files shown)


jrevels:
Looks like tests are failing because we no longer need @andreasnoack's DataFrames branch? Does that mean we can switch over to master/a new DataFrames tag?

src/deterministic_vi/ConstraintTransforms.jl
```diff
@@ -359,22 +360,26 @@ end
...
 
+const NUM_ELBO_BOUND_PARAMS = length(CanonicalParams)
+const NUM_ELBO_FREE_PARAMS = (c = ConstraintBatch(); free_length(c.boxes[1]) + free_length(c.simplexes[1]))
+const NUM_ELBO_JACOBIAN_PARAMS = NUM_ELBO_BOUND_PARAMS * NUM_ELBO_FREE_PARAMS
+
 @compat const TransformJacobianConfig{M,N,T} = JacobianConfig{N,T,Tuple{Array{Dual{N,T},M},Vector{Dual{N,T}}}}
 
 immutable TransformDerivatives{N,T}
-    jacobian::Matrix{T}
+    jacobian::MMatrix{NUM_ELBO_BOUND_PARAMS,NUM_ELBO_FREE_PARAMS,Float64,NUM_ELBO_JACOBIAN_PARAMS}
```
Keno: SizedMatrix may have better optimization properties, because the compiler knows about arrays.

src/deterministic_vi/ConstraintTransforms.jl
```diff
@@ -359,22 +360,26 @@ end
 using ForwardDiff
 using ForwardDiff: Dual, jacobian!, JacobianConfig
 
+const NUM_ELBO_BOUND_PARAMS = length(CanonicalParams)
+const NUM_ELBO_FREE_PARAMS = (c = ConstraintBatch(); free_length(c.boxes[1]) + free_length(c.simplexes[1]))
```
Keno: ERROR: LoadError: LoadError: LoadError: MethodError: no method matching Celeste.DeterministicVI.ConstraintTransforms.ConstraintBatch()
Closest candidates are:
  Celeste.DeterministicVI.ConstraintTransforms.ConstraintBatch(!Matched::Array{Array{T,1},1}) where T at /home/ubuntu/.julia/v0.6/Celeste/src/deterministic_vi/ConstraintTransforms.jl:61
  Celeste.DeterministicVI.ConstraintTransforms.ConstraintBatch(!Matched::Array{Array{T,1},1}, !Matched::Real) where T at /home/ubuntu/.julia/v0.6/Celeste/src/deterministic_vi/ConstraintTransforms.jl:61
  Celeste.DeterministicVI.ConstraintTransforms.ConstraintBatch(!Matched::Array{Array{T,1},1}, !Matched::Real, !Matched::Real) where T at /home/ubuntu/.julia/v0.6/Celeste/src/deterministic_vi/ConstraintTransforms.jl:61

jrevels:
Fixed the MethodError and switched to using `SizedArray`. I wonder if my method here to extract the bounds from the tuple-valued parameter actually works to propagate the bounds as compile-time info.

Keno:
Let's hold off on merging this until I get a chance to validate the performance.

jrevels:
It seems like this is now stale/has already been achieved in a different manner on master.

PR #620, opened 2017-03-15
Multinode functionality from the `multijoint` branch

kpamnany:
Recreated due to strange git problems with `multijoint`. Current state: single infer runs working. Also, fix for static compilation (can't pipe commands), and some affinity tweaking/testing.

### DIFF
--- a/bin/infer-boxes.jl
+++ b/bin/infer-boxes.jl
@@ -3,29 +3,42 @@
 import Celeste.ParallelRun: BoundingBox, infer_boxes
 import Celeste.Log
 
-
-const usage_info =
-"""
-Usage:
-  infer-boxes.jl <boxes_file> <out_dir>
-"""
-
-const stagedir = ENV["CELESTE_STAGE_DIR"]
-
 if length(ARGS) != 2
-    println(usage_info)
+    println("""
+        Usage:
+          infer-boxes.jl <boxes_file> <out_dir>
+
+        <boxes_file> format, one line per box:
+        <difficulty>	<#RCFs>	<#sources>	<ramin> <ramax> <decmin> <decmax>
+        """)
 else
     all_boxes = BoundingBox[]
-    sky_boxes = ARGS[1]
-    f = open(sky_boxes)
+    box_source_counts = Int64[]
+    boxes_file = ARGS[1]
+    f = open(boxes_file)
     for ln in eachline(f)
-        ss = split(ln, ' ')
-        ns = [parse(Float64, x) for x in ss]
-        bb = BoundingBox(ns...)
+        lp = split(ln, '\t')
+        if length(lp) != 4
+            println("malformed line in box file, skipping remainder")
+            println("> $ln")
+            break
+        end
+        sc = parse(Int64, lp[3])
+        push!(box_source_counts, sc)
+        ss = split(lp[4], ' ')
+        ramin = parse(Float64, ss[1])
+        ramax = parse(Float64, ss[2])
+        decmin = parse(Float64, ss[3])
+        decmax = parse(Float64, ss[4])
+        bb = BoundingBox(ramin, ramax, decmin, decmax)
         push!(all_boxes, bb)
     end
     close(f)
     outdir = ARGS[2]
-    infer_boxes(all_boxes, stagedir, outdir)
+    if length(all_boxes) < 1
+        println("box file is empty?")
+        exit(-1)
+    end
+    infer_boxes(all_boxes, box_source_counts, ENV["CELESTE_STAGE_DIR"], outdir)
 end
 
--- a/src/ParallelRun.jl
+++ b/src/ParallelRun.jl
@@ -15,7 +15,6 @@ import ..PSF
 
 import ..DeterministicVI: infer_source
 
-
 include("joint_infer.jl")
 
 # In production mode, rather the development mode, always catch exceptions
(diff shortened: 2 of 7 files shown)


kpamnany:
Was the node/rank and thread ID display on the Log functions useful @jeff-regier? Should I revert the changes to that file?

jeff-regier:
Yes, it's somewhat useful.

bin/infer-boxes.jl
```diff
@@ -3,29 +3,40 @@
...
-    f = open(sky_boxes)
+    box_source_counts = Int64[]
+    boxes_file = ARGS[1]
+    f = open(boxes_file)
     for ln in eachline(f)
-        ss = split(ln, ' ')
-        ns = [parse(Float64, x) for x in ss]
-        bb = BoundingBox(ns...)
+        lp = split(ln, '\t')
+        if length(lp) != 4
```
jeff-regier, line 21: maybe print an error message too...failing silently could be a problem

src/ParallelRun.jl
```diff
@@ -136,21 +141,21 @@ function infer_init(rcfs::Vector{RunCamcolField},
     tic()
     catalog = SDSSIO.read_photoobj_files(rcfs, stagedir,
                         duplicate_policy=duplicate_policy)
-    timing.read_photoobj = toq()
-    Log.info("$(length(catalog)) primary sources")
+    timing.read_photoobj = timing.read_photoobj + toq()
+    #ntputs(grank(), threadid(), "$(length(catalog)) primary sources")
```
jeff-regier: Are you worried the logging statements will print so much that they slow down the job? These print just once per box--not very often.

src/ParallelRun.jl
```diff
@@ -161,14 +166,14 @@ function infer_init(rcfs::Vector{RunCamcolField},
...
-            timing.read_img = toq()
+            timing.read_img = timing.read_img + toq()
         catch ex
             Log.exception(ex)
         end
 
         tic()
         neighbor_map = Infer.find_neighbors(target_sources, catalog, images)
-        Log.info("neighbors found in $(toq()) seconds")
+        timing.find_neigh = timing.find_neigh + toq()
```
jeff-regier: `+=`

src/ParallelRun.jl
```diff
@@ -250,7 +285,7 @@ function one_node_single_infer(config::Configs.Config,
                 unlock(results_lock)
             catch ex
                 if is_production_run || nthreads() > 1
-                    Log.exception(ex)
+                    Log.error(string(ex))
```
jeff-regier: `Log.exception` gives quit a bit more information. Are you sure you want to change it?

src/ParallelRun.jl
```diff
@@ -377,33 +412,56 @@ save_results(outdir, box, results) =
...
+    #@time results = one_node_single_infer(catalog, target_sources, neighbor_map, images; timing=times)
+    @time results = one_node_joint_infer(catalog, target_sources, neighbor_map, images; timing=times)
 
     tic()
     save_results(outdir, box, results)
-    timing.write_results = toq()
+    times.write_results = toq()
 
-    puts_timing(timing)
+    puts_timing(times)
```
jeff-regier: These instance of `InferTiming` are called `times` some places and `timing` elsewhere. Could we stick to just one of them?

src/multinode_run.jl
```diff
@@ -1,193 +1,700 @@
...
+Initialize elbo args for the specified target source.
+"""
+function init_elboargs(ts::Int, catalog::Vector{CatalogEntry},
+                       target_sources::Vector{Int},
+                       neighbor_map::Vector{Vector{Int}}, images::Vector{Image},
+                       ts_vp::Dict{Int64,Array{Float64}},
+                       ea_vec::Vector{ElboArgs},
+                       cfg_vec::Vector{Config{DEFAULT_CHUNK,Float64}};
+                       use_fft=false,
+                       trust_region=NewtonTrustRegion())
```
jeff-regier, line 390: These `use_fft` and `trust_region` flags aren't used any more in `joint_infer.jl`. Can you call the code in `joint_infer` to initialize elbo args, either `initialize_elboargs_sources!` or `initialize_elboargs_sources_kernel!` rather than duplicating the code here?

src/multinode_run.jl
```diff
@@ -1,193 +1,700 @@
...
+        #ntputs(grank(), threadid(), "trying source $s, ea $(cbox.ea_vec[s])")
+        try
+            ea, cfg = cbox.ea_vec[s], cbox.cfg_vec[s]
+            if use_fft
+                f = FFTElboFunction(load_fsm_mat(ea, cbox.images; use_raw_psf=true))
+            else
+                f = DeterministicVI.elbo
+            end
+            ntputs(grank(), threadid(), "maximizing source $s")
+            NewtonMaximize.maximize!(f, ea, cfg)
```
jeff-regier: This call won't work because `maximize!` also needs `vp` based to it. Can you reuse the function in `joint_infer.jl` that calls NewtonMaximize, modifying that function as necessary? Those functions all are exercised by unit tests.

src/ParallelRun.jl
jeff-regier: If you don't want the info commands printed, rather than comment these out, would you change Log.info to only print when the CELESTE_PROD environment variable is not set? I did that for Log.debug already.

jeff-regier: Or if they aren't necessary any more because of your timing code, let's just delete them rather than commenting them out.

jeff-regier (changes requested):

kpamnany:
Addressed all comments except the last two -- I'm still working on getting multi-node joint infer updated based on what's changed in `joint_infer.jl`. That'll be another PR.

PR #622, opened 2017-03-17
Multinode joint inference

kpamnany:
Is working. Also, thread affinity issues sorted and a couple of minor fixes.

### DIFF
--- a/src/multinode_run.jl
+++ b/src/multinode_run.jl
@@ -1,4 +1,39 @@
+# ----------------
+# A simple centralized sense reversing thread barrier. Needed to allow the
+# ordering constraints that Cyclades partitioning imposes on the processing
+# of sources.
+# ----------------
+
+type CentSRBarrier
+    thread_counter::Atomic{Int}
+    num_threads::Int
+    sense::Int
+    thread_senses::Vector{Int}
+end
+
+function setup_barrier(num_threads::Int)
+    CentSRBarrier(Atomic{Int}(num_threads), num_threads, 1, ones(Int, nthreads()))
+end
+
+function thread_barrier(bar::CentSRBarrier)
+    tid = threadid()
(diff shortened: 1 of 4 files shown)


src/multinode_run.jl
```diff
@@ -362,14 +356,15 @@ Return a Cyclades partitioning or an equal partitioning of the target sources.
 """
 function partition_box(npartitions::Int, target_sources::Vector{Int},
                        neighbor_map::Vector{Vector{Int}};
-                       cyclades=true,
+                       cyclades_partition=true,
                        batch_size=400)
-    if cyclades
+    if cyclades_partition
```
jeff-regier: This code is duplicated in `one_node_joint_infer`. Would you move this function to `joint_infer.jl` and call it from `one_node_joint_infer`, and remove the redundant code from `one_node_joint_infer`?

src/multinode_run.jl
```diff
@@ -380,47 +375,43 @@ end
...
+        entry_id = target_sources[ts]
+        entry = catalog[entry_id]
+        neighbor_ids = neighbor_map[ts]
+        neighbors = catalog[neighbor_ids]
+        cat_local = vcat([entry], neighbors)
+        ids_local = vcat([entry_id], neighbor_ids)
+
         patches = Infer.get_sky_patches(images, cat_local)
-        Infer.load_active_pixels!(images, patches)
+        Infer.load_active_pixels!(config, images, patches)
```
jeff-regier: This code is duplicated in `initialize_elboargs_sources_kernel!`. Would you make the shared parts a new function in `joint_infer.jl`? That way it'll be exercised by the unit tests and there won't be divergence between the two versions.

jeff-regier (changes requested):
Looks good Kiran, just 2 requested changes.

jeff-regier (approved):

PR #626, opened 2017-03-23
Joint infer dynamic

agnusmaximus:
Use dynamic load balancing with locks (I think locks are fine since the updates are so slow anyways)

EDIT: @kpamnany this change will mess up your multinode_run.jl since init_box calls partition_box which is changed in this PR. Let's not merge this change yet...


agnusmaximus:
Also, long running tests for joint infer are broken :(

It seems like they were broken amidst the previous changes...

PR #627, opened 2017-03-24
Remove unused variables

andreasnoack:
Ran the code through the linter while waiting for a run on Cori and deleted unused variables. I guess most of these should go but some of them might be placeholders for future use. Please let me know.


jeff-regier:
Great, thanks, these changes all seem like improvements.

PR #628, opened 2017-03-24
Fix bug + support plots of posterior z-scores

gostevehoward:
As described in commit messages.

`get_uncertainty_df` and `get_error_df` could probably be merged but it's nontrivial so i'm putting it off for now.


jeff-regier:
These Q-Q plots are a pretty nice to assess the UQ quality.

PR #630, opened 2017-03-27
Synthetic.jl image generation, asinh magnitudes, and various fixes

### LINKED ISSUE #618: Implement asinh magnitudes in `AccuracyBenchmark`
In scoring accuracy of reference band brightness, we look at error in units of both flux (nMgy) and mags (Pogson magnitudes). There seems to be some consensus that comparing on asinh magnitudes would be better. See #510 for some comments and [this page](http://www.sdss.org/dr12/algorithms/magnitudes/) for definitions and discussion.

### LINKED ISSUE #619: Script to generate FITS images from a catalog using `Synthetic.jl`
That is, drawing images from the Celeste likelihood model. An alternative to `benchmarks/galsim/galsim_field.py`. Broken out from #510.

gostevehoward:
Sorry this PR got so large, but here are the highlights:

* Implements generation of FITS imagery from a catalog CSV using Synthetic.jl (fixes #619)
* Replaces Pogson magnitudes with asinh magnitudes (fixes #618)
* In both GalSim and Synthetic.jl FITS generation, use a WCS which matches that used in SDSS imagery as closely as possible (see https://github.com/jeff-regier/Celeste.jl/wiki/Coordinate-systems for the gory details)
* Modify "epsilon" and "iota" in both GalSim and Synthetic.jl image generation to more closely match SDSS imagery (but not in GalSim *benchmark* image generation)
* When matching SDSS primary predictions to SDSS coadd predictions, limit position matching radius to 1 px (rather than 1 / .396 ~ 2.5 px). This was the original intention, it doesn't cut down on number of matched sources much, and it will hopefully avoid some bad matches which throw misleading outliers into the error analysis.
* Add various features to the benchmark/accuracy suite to support running on other SDSS RCFs and to aid debugging (e.g., writing raw matched predictions and raw normalized errors to CSVs)


gostevehoward:
Found and fixed the bug with angles in Synthetic.jl imagery (just a radians/degrees bug).

I found it by making [this catalog](https://github.com/jeff-regier/Celeste.jl/files/873593/angle_test.csv.zip) (by hand, in Google Sheets) and generating FITS imagery from it, which looks like this (after the fix):

image

The `benchmark/accuracy` suite makes it easy to set up and check stuff like this.

benchmark/accuracy/sdss_rcf_to_csv.jl
```diff
@@ -51,7 +51,7 @@ function match_and_load_objids(catalog_df::DataFrame, objid_csv::String)
             truth_df[:right_ascension_deg], truth_df[:declination_deg],
             catalog_df[source_index, :right_ascension_deg],
             catalog_df[source_index, :declination_deg],
-            1 / 0.396, # to match up with Stripe82Score
+            1.0,
```
jeff-regier, line 54: Great. Much less confusing.

benchmark/accuracy/write_ground_truth_catalog_csv.jl
```diff
@@ -15,62 +15,29 @@ GALAXY_ONLY_COLUMNS = [
...
-into coadd_field_catalog
-from stripe82.photoobj
-where
-  run in (106, 206) and
-  ra between 0.449 and 0.599 and
-  dec between 0.417 and 0.629) as tmp
-where
-  ((psfmag_i < 22 and probpsf = 1) or (probpsf = 0 and (expmag_i < 22 or devmag_i < 22))) and
-  (flags & @bad_flags) = 0
-"""
```
jeff-regier, line 52: Ah, I suspected you'd moved this to the wiki, and then I found it there.

benchmark/galsim/Sandbox.jl
```diff
@@ -0,0 +1,24 @@
+module Sandbox
```
jeff-regier, line 1: There's a known issue where ForwardDiff has to be loaded before StaticArrays. I wonder if some that's what's happening when you use Sandbox to workaround the bug. Hopefully the bug will just go away as Julia 0.6 matures.

benchmark/galsim/galsim_field.py
```diff
@@ -11,20 +11,22 @@
...
 def set_image_dimensions(test_case, catalog_rows):
     min_ra_deg = min(float(row['right_ascension_deg']) for row in catalog_rows)
     max_ra_deg = max(float(row['right_ascension_deg']) for row in catalog_rows)
     min_dec_deg = min(float(row['declination_deg']) for row in catalog_rows)
     max_dec_deg = max(float(row['declination_deg']) for row in catalog_rows)
 
-    width_arcsec = (max_ra_deg - min_ra_deg) * ARCSEC_PER_DEGREE + 2 * FIELD_EXPAND_ARCSEC
-    height_arcsec = (max_dec_deg - min_dec_deg) * ARCSEC_PER_DEGREE + 2 * FIELD_EXPAND_ARCSEC
+    height_arcsec = (max_ra_deg - min_ra_deg) * ARCSEC_PER_DEGREE + 2 * FIELD_EXPAND_ARCSEC
+    width_arcsec = (max_dec_deg - min_dec_deg) * ARCSEC_PER_DEGREE + 2 * FIELD_EXPAND_ARCSEC
```
jeff-regier, line 24: In hindsight, maybe "column" and "row" would have been better than "height" and "width". Unless the columns of the CCD become the rows in Julia image data structure. If that's the case, then any terms would be better than "column" and "row"...

benchmark/speed/benchmark_elbo_likelihood.jl
```diff
@@ -3,10 +3,8 @@
 import Celeste.DeterministicVI
 import Celeste.DeterministicVI: ElboIntermediateVariables, elbo_likelihood
 
-include(joinpath(Pkg.dir("Celeste"), "test", "Synthetic.jl"))
```
jeff-regier, line 6: Perfect, glad to have `Synthetic.jl` back in the `src` directory.

src/AccuracyBenchmark.jl
```diff
@@ -347,7 +364,7 @@ function celeste_to_df(results::Vector{ParallelRun.OptimizedSource})
 end
 
 ################################################################################
-# Generate a catalog from the Celeste prior
+# Generate a random catalog from the Celeste prior
```
jeff-regier, line 367: Though the catalog isn't random once it's been generated

jeff-regier:
Looks great Steve. Thanks for your work here.

benchmark/accuracy/write_ground_truth_catalog_csv.jl
gostevehoward: Oh yeah, I need to remove the 2-3 other places where this thing appears in the code. Will do on next PR.

benchmark/galsim/Sandbox.jl
gostevehoward: Ah, good to know. Also I didn't mean to check in this file and don't really think it should be checked in. Will remove.

benchmark/galsim/galsim_field.py
gostevehoward: Yeah I'm not sure what the best way is...but I did write a painstaking description of the situation, in case you didn't see it: https://github.com/jeff-regier/Celeste.jl/wiki/Coordinate-systems

jeff-regier: I think pixel coordinates actually begin at (0.5, 0.5). That way the center of the first pixel is (1.0,1.0). Since Julia is 1-based, `[1,1]` also happens to be the index of the first pixel. In C the first pixel's center still has coordinates (1.0,1.0), because pixel coordinates are independent of language, but that pixel's index is `[0,0]`. Tricky stuff.

gostevehoward: Ah, you're right, the first pixel is pixel coordinate (1, 1). Thanks! I'll have to update some things...

PR #633, opened 2017-03-28
Make Celeste work with latest version of StaticArrays

andreasnoack:
The promotion in `StaticArrays` has changes such that `SVector+Vector` now produces an `SVector`.


andreasnoack:
The change in `StaticArrays` might create some `Vector`s instead of `SVector`s in the code that don't cause test errors but will cause extra allocations. I'm investigating that now.

UPDATE: For `benchmark_one_lightsource`, the new version allocates less.

jeff-regier:
Great!

PR #634, opened 2017-03-30
removed checkout of dataframes and dataarray in runtests.jl

jeff-regier:
@andreasnoack says it's no longer necessary

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,4 @@
-julia 0.6-
+julia 0.6
 Compat 0.18.0
 DataFrames
 DataStructures 0.5.2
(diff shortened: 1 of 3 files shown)


REQUIRE
```diff
@@ -1,4 +1,4 @@
-julia 0.6-
+julia 0.6
```
andreasnoack, line 1: I believe this is too early. I'm pretty sure Julia master won't satisfy this requirement.

jeff-regier: It seems to. I didn't get a warning.

jeff-regier: I'm using `Version 0.6.0-pre.alpha.313 `

jeff-regier: Oh, you're right. I changed in back.

PR #635, opened 2017-03-31
automatic batchsize selection joint infer

agnusmaximus:


jeff-regier:
Test error... Should I commit this? I'd be good to have the thread load imbalance stats.

agnusmaximus:
Hmm, I don't think this helps much... Is the plan to remove the light sources that are causing the load imbalance? If so, then this probably won't be of much help.

PR #636, opened 2017-04-02
Use seeding constructor of MersenneTwister instead of srand

rfourquet:
This comes from https://github.com/JuliaLang/julia/pull/16984#issuecomment-290914407 (the zero-arg `MersenneTwister()` will be deprecated). Here the change is not necessary, but thought I could update the code as well.


jeff-regier:
Thanks @rfourquet !

PR #640, opened 2017-04-06
fix Optim.Options type over-restriction and add termination_callback test

jrevels:
The single type parameter of `Optim.Options` is actually the type of the callback, and we were always restricting it to `Void` before (since `callback = nothing` by default). I removed that restriction in `Config` to allow for actual callbacks to be passed.

This PR also adds a test to ensure that Optim is actually using/respecting the callback (which it seems to do appropriately).


jeff-regier:
@kpamnany

PR #643, opened 2017-04-07
Change to how `SArray`s are defined

kpamnany:


andreasnoack:
Requires `StaticArrays` master

PR #650, opened 2017-04-10
20 minute timeout for box processing

kpamnany:
Single node and multi node. Included a thread imbalance measurement heuristic for the kill switch (which doesn't kill the box, but just reports): at least half the threads have finished AND the last thread to finish has been waiting at least 25% of the time since optimization started.


jeff-regier:
@agnusmaximus Would you look over this and merge it if it's fine?

agnusmaximus:
Ok, I'm checking it out...

agnusmaximus:
Looks good!!

PR #656, opened 2017-04-12
Filter saturated light sources in scoring, not in catalog generation

### LINKED ISSUE #654: Keep saturated sources in coadd catalog, filter out in scoring only
Right now we filter out some oversaturated sources in the [CasJobs query](https://github.com/jeff-regier/Celeste.jl/wiki/About-SDSS-and-Stripe-82) we use to get coadd data, and some in the code to load the "primary" catalog (`AccuracyBenchmark.load_primary`). Per Jeff's comment in #625, it probably makes sense to keep all sources in catalogs and filter them out just when we're doing scoring. For example, this will ensure bright sources in the main image are also present when we generated synthetic imagery. It would also affect joint inference, I think (i..e, if an oversaturated source is near a dim source, we need the primary catalog to contain the oversaturated source so it will be initialized in Celeste, even though its inference results won't be scored, so it can be used in joint inference to make more accurate estimates for the nearly dim source).

gostevehoward:
Fixes #654


gostevehoward:
I updated the SQL query on https://github.com/jeff-regier/Celeste.jl/wiki/About-SDSS-and-Stripe-82

jeff-regier:
Looks good. Let's wait until after the GB runs to merge this PR---new light
sources may throw off the load balance.

gostevehoward:
Absolutely, I wasn't sure if you guys are running off master, but
definitely don't merge in any of my stuff if you are!

PR #657, opened 2017-04-13
New format bigfiles

kpamnany:
Bigfiles are restaged on SCRATCH at this moment, so need this to use 'em. BB restaging is in the queue.


kpamnany:
BB re-staging is done, so need this. Merging.

PR #670, opened 2017-06-06
Refactor IO strategies

Keno:
This adds some more flexibity to I/O, for non-NERSC environemnts.
In particular it allows:
- Data in the original SDSS directory layout, rather than the one that Celeste puts them in
- Decompresing data on the fly rather than requiring it to be decompressed on disk
- Clusters without shared filesystems by doing RPC to a master process

Since the options were getting a bit out of hand with these additions,
I also added a YAML based configuration format to specify these options.
I plan to extend that settings file to cover other aspects of the system
as well (e.g. the distributed options).

Finally, I removed the big file option. That was always a bit of a stopgap.
We didn't document the format. The scripts to create them are not in tree.
With more time, we can probably come up with a better way to do it. Also,
it's untested. It's still in the history of course, so we can always bring
it back.

Lastly, I added tests for all the new I/O modes.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -11,3 +11,7 @@ JLD 0.6.0
 Optim 0.7.4
 StaticArrays 0.5.1
 WCS 0.1.3
+YAML
+TranscodingStreams
+CodecBzip2
+CodecZlib
--- a/bin/infer-boxes.jl
+++ b/bin/infer-boxes.jl
@@ -10,12 +10,7 @@ function run_infer_boxes(args::Vector{String})
     if length(args) < 3
         println("""
 Usage:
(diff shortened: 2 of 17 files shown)


Keno:
Oh yeah, I should point out the the decompression packages that this depends on are in the process of being registered, so hold on until that's done, I suppose.

Keno:
Rebased. I have another batch of changes on top of this, but let's take this as a natural midpoint and get these changes up to par first.

jeff-regier:
Looks good, but one error, perhaps from the rebase:
```
ERROR: LoadError: LoadError: MethodError: no method matching load_raw_images(::Celeste.SDSSIO.RunCamcolField, ::String, ::Bool)
```

jeff-regier:
Great, thanks @Keno

PR #679, opened 2017-08-06
Detect objects on images

kbarbary:
This PR does the following in pursuit of initializing sources directly from images rather than using the SDSS catalog: 

- Add a self-contained submodule `SEP` that wraps the C library from http://github.com/kbarbary/sep . This library does heuristic spatially-variable background / noise estimation, image segmentation (source detection), source deblending, and calculates some basic source statistics. This is a wrapper that could eventually be moved to a package, but for now it is easier to only wrap what we need and be able to change the API rapidly.
- Add a `BinDeps`-based build script for the above.
- Add a `detect_sources()` function (in ParallelRun.jl) that takes a bunch of `RawImage`s and returns a `Vector{CatalogEntry}`.
- Changes `infer_init()` to use the above function.
- New neighbor map implementation in `infer_init` based on extent of a source's pixels that are "above background noise" (since we now have that information in `detect_sources()`.
- Removes `thingid` from `compare_vp_params` in tests, instead simply requiring that both vectors of parameters be ordered the same.

Still to do: Update accuracy benchmarks to be independent of `obj_id` / `thingid`.

### DIFF
--- a/src/ParallelRun.jl
+++ b/src/ParallelRun.jl
@@ -317,15 +317,15 @@ end
 
 
 """
-Given a list of RCFs, load the catalogs, determine the target sources,
-load the images, and build the neighbor map.
+New version of infer_init() that uses detect_sources() instead of SDSS
+photoObj catalog for initialization.
 """
-function infer_init(rcfs::Vector{RunCamcolField},
-                    strategy::SDSSIO.IOStrategy;
-                    objid="",
-                    box=BoundingBox(-1000., 1000., -1000., 1000.),
-                    primary_initialization=true,
-                    timing=InferTiming())
+function infer_init_new(rcfs::Vector{RunCamcolField},
+                        strategy::SDSSIO.IOStrategy;
+                        objid="",
+                        box=BoundingBox(-1000., 1000., -1000., 1000.),
+                        primary_initialization=true,
+                        timing=InferTiming())
 
     # check if any non-default values are passed for functionality that has
     # been removed.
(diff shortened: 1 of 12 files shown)


kbarbary:
Actually now that I'm running the benchmarks and looking at the code, it seems like `infer_init()` is not actually called in `benchmarks/accuracy/run_celeste_on_field.jl`, so I think this can be merged without affecting the benchmarks.

jeff-regier:
The SEP-related additions all look good. Can you tell whether SEP (as called in this PR) performs as well as primary? Would it be appropriate to add a unit test that verifies that SEP finds, say, at least 90% of the light sources that primary does?

jeff-regier:
I think we'll want the the accuracy benchmarks to have the option of initializing Celeste with SEP too. Currently there's a script `sdss_rcf_to_csv.jl` that takes a Run-Camcol-Field triple and writes as csv catalog based on Primary. Maybe it could take a flag `--sep` that causes it to write a csv catalog based on SEP rather than Primary. Does that make sense?  We could still call `run_celeste_on_field.jl` with the ` --use-full-initialization` as before, just now pas it a csv created by SEP.

src/ParallelRun.jl
```diff
@@ -153,61 +304,79 @@ function infer_init(rcfs::Vector{RunCamcolField},
...
-    entry_in_range = entry->((box.ramin < entry.pos[1] < box.ramax) &&
-                             (box.decmin < entry.pos[2] < box.decmax))
-    target_sources = find(entry_in_range, catalog)
+    # Initialize variables to empty vectors in case try block fails
+    catalog = CatalogEntry[]
+    source_rcfs = RunCamcolField[]
+    source_cat_idxs = Int16[]
+    source_radii = Float64[]
+    target_sources = Int[]
+    images = Image[]
```
jeff-regier, line 345: Nice, that is better.

src/ParallelRun.jl
```diff
@@ -153,61 +304,79 @@ function infer_init(rcfs::Vector{RunCamcolField},
...
-            Log.exception(ex)
-            empty!(target_sources)
+            images = [convert(Image, raw_image) for raw_image in raw_images]
+        catch exc
+            Log.exception(exc)
+            rethrow()
         end
 
-        tic()
-        neighbor_map = Infer.find_neighbors(target_sources, catalog, images)
```
jeff-regier: Can you delete the old `find_neighbors` too then?

src/ParallelRun.jl
```diff
@@ -153,61 +304,79 @@ function infer_init(rcfs::Vector{RunCamcolField},
...
+                 "$(box.ramax), $(box.decmin), $(box.decmax)")
 
-    # Load images and neighbor map for target sources
-    if length(target_sources) > 0
-        # Read in images for all (run, camcol, field).
+        # convert raw images to images
         try
-            tic()
-            images = SDSSIO.load_field_images(strategy, rcfs)
-            timing.read_img += toq()
```
jeff-regier: Yeah, all this timing code gets in the way. Glad you're deleting it.

kbarbary: Not deleted, just moved (this diff is hard to read). I can really delete the timing code in a separate PR if we really want to do that though.
More importantly, it would be good to know if these `try ... catch`s are still relevant or needed. I get the feeling that they were put in as temporary kludges and should be removed.

src/ParallelRun.jl
kbarbary: Still used in the accuracy benchmarks at the moment, but can be deleted once that is updated.

kbarbary:
Good idea to test SEP against primary. I can do that before this is merged.

I agree that accuracy benchmarks should be updated to actually use the SEP initialization. However, it would be nice to update accuracy benchmarks as a separate PR, as it will be non-trivial. But I understand that you might want to see the results before merging. What do you think about a separate PR for that?

jeff-regier:
I have mixed feelings: without the accuracy scores it's hard to be sure we're doing something reasonable. On the other hand, it's good to merge frequently.  So either way is OK I guess.

src/ParallelRun.jl
jeff-regier: They're still relevant if we process all of SDSS. Some weird things happen on like 1 in 1000 fields, and these keep the whole job from crashing. It would be better to get to the bottom of it but until then these are better than nothing.

kbarbary:
I haven't yet addressed the benchmark scores, but I've updated the initialization so that it runs detection on every image and merges the resulting catalogs. The results are pretty good:

red = SDSS catalog
green = new SEP catalog
(click image)
screenshot from 2017-08-11 14-44-49
(image is r band, but catalog built from detections in all images)

The major difference is that SDSS masks regions around bright stars and we don't do that yet. Therefore we find many extra objects around bright stars. (Deblending parameters set for faint galaxies don't work well on bright stars.)

Besides that there are some differences in deblending, and there are some differences in detection of very faint objects. Sometimes SDSS does better deblending and sometimes SEP does. Same with detection. There are also some moderately bright sources that are not in the SDSS catalog.  I imagine they were culled by some cut, but can't imagine what.

kbarbary:
@jeff-regier For benchmarking, I hear you. However, I *really* want to do some refactoring (such as removing run-camcol-field and objid, thing_id throughout the code) before I do any more additions. The refactoring will make the additions easier, and will allow simplifications in the code in this PR. Yes, I could do them as part of this PR, but that would make the refactored parts really hard to review and this PR is already big as it is.

As an intermediate solution, I've restored the existing `infer_init()` function in this branch, so nothing is lost and the new code here is not actually yet used anywhere.

So, I'd recommend that this be merged as is, now that there is a sanity-check test for `detect_sources()`.

jeff-regier:
Looks like SEP is working very well--great news. Sure, let's merge if you recommend that.

PR #681, opened 2017-08-14
Mcmc inference

andymiller:
Created a `mcmc` directory with three files

- `slicesample.jl` : julia implementation of a generic slice sampler
- `mcmc_functions.jl` : likelihood maker function wrappers
- `mcmc_misc.jl` : miscellaneous functions for handling samples 

`mcmc` files are included in the `MCMC` module at the `src` level.  The likelihood functions in `mcmc_functions.jl` rely on functions very similar to `Synthetic` --- they do not rely on the `logprob` in the `Model` module.  

Also included `test/test_mcmc.jl` that tests likelihood functions as well as the slice sample implementation (to make sure it runs, not for correctness yet).


src/mcmc/mcmc_functions.jl
```diff
@@ -0,0 +1,333 @@
...
+           for k in 1:nk]
+    lnpik = log.(pp.k[:, type_i])
+    llc   = Model.logsumexp(llk .+ lnpik)
+
+    # add and return
+    return llr + llc
+end
+
+"""
+More numerically stable converter from log fluxes to colors
```
jeff-regier, line 267: Could we replace the unstable one with this one, to avoid redundant code? (Where is the unstable one?)

test/test_mcmc.jl
```diff
@@ -0,0 +1,111 @@
+using Base.Test
+
+using Celeste: Model, Transform, SensitiveFloats, MCMC, Synthetic
```
jeff-regier, line 3: Are you using Transform, or did it get important accidentally?

src/MCMC.jl
```diff
@@ -1,3 +1,19 @@
 module MCMC
 
+using Celeste: Model, Transform, SensitiveFloats, MCMC
```
jeff-regier, line 3: Are you using Transform, or did it get imported accidentally?

src/mcmc/mcmc_misc.jl
```diff
@@ -0,0 +1,296 @@
...
+      psrfs = MCMC.potential_scale_reduction_factor(chains)
+      println(" potential scale red factor", psrfs)
+    end
+  end
+end
+
+
+######################################################################
+# unit flux image utils --- writes unit flux images to pixel array   #
+# Very similar to Synthetic.jl functions                             #
```
jeff-regier, line 176: Is there duplicate code here? Could we instead modify Synthetic.jl so that those funds are general enough to work for both your use case and whatever they're doing now?

jeff-regier (approved):
Looks good Andy, thanks for getting it in. I have a few comments about the duplicated code. Those can be addressed in a later PR.

src/mcmc/mcmc_functions.jl
andymiller: Sounds good --- I had been using the `fluxes_to_colors` method in [`log_prob.jl`](https://github.com/jeff-regier/Celeste.jl/blob/master/src/model/log_prob.jl#L317-L325), which required that I go from log-flux => flux space.  Some (ultimately rejected) big proposals would create `-Inf` and `Inf`s which would kill the sampler.

test/test_mcmc.jl
andymiller: Accidental --- I'll remove it in the next PR.

src/MCMC.jl
andymiller: accidental.

src/mcmc/mcmc_misc.jl
andymiller: I think it's possible to merge them and put overlapping code in a general module (maybe `Infer`?).   I intend to make these a bit more general (e.g. to incorporate patch offsets, and not necessarily take in an `Image`).

PR #684, opened 2017-08-16
move Config to top level

kbarbary:
It seemed like the `Config` type doesn't need to live in it's own submodule `Configs`, of which it was the only occupant.

I also changed the name of `DeterministicVI.ElboMaximize.Config` to `ElboConfig` to avoid name conflicts or confusion.

### DIFF
--- a/benchmark/accuracy/run_celeste_on_field.jl
+++ b/benchmark/accuracy/run_celeste_on_field.jl
@@ -4,7 +4,7 @@ using DataFrames
 
 import Celeste.AccuracyBenchmark
 import Celeste.ArgumentParse
-import Celeste.Configs
+import Celeste: Config
 import Celeste.Infer
 import Celeste.ParallelRun
 import Celeste.SDSSIO
@@ -101,7 +101,7 @@ else
 end
 neighbor_map = Infer.find_neighbors(target_sources, catalog_entries, images)
 
-config = Configs.Config()
+config = Config()
 if haskey(parsed_args, "min-radius-px")
     config.min_radius_pix = parsed_args["min-radius-px"]
 end
--- a/src/AccuracyBenchmark.jl
+++ b/src/AccuracyBenchmark.jl
@@ -6,12 +6,12 @@ import FITSIO
 import StaticArrays
 import WCS
 
-import Celeste.Configs
-import Celeste.DeterministicVI
-import Celeste.Infer
-import Celeste.Model
-import Celeste.ParallelRun
-import Celeste.SDSSIO
+import ..Config
+import ..DeterministicVI
+import ..Infer
+import ..Model
+import ..ParallelRun
+import ..SDSSIO
 
 const ARCSEC_PER_DEGREE = 3600
 const SDSS_ARCSEC_PER_PIXEL = 0.396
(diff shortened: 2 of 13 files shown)


jeff-regier:
`DeterministicVI.ElboMaximize.Config` -> `ElboConfig` should help avoid a lot of confusion.

@gostevehoward Do you recall why `Configs.Config` has it's own submodule, rather than existing the top-level scope? Was there going to be a whole network of Config objects at some point?

gostevehoward:
Certainly there was a thought that there might be a hierarchy of objects in there, as [discussed here](https://github.com/jeff-regier/Celeste.jl/wiki/The-%60Configs.Config%60-object). I'm not sure if that's the reason I made a module for it, that's pretty speculative. The other reason might have been that this is an object that everything else depends on, so it should be at the bottom of the dependency hierarchy. The Celeste module contains all the other ones, so it's at the top of the dependency hierarchy. Putting the Config object in the top-level module seems like it creates a circular dependency. In many languages that will break things, and in any case that can be a source of confusion. But in Julia, maybe that works, and maybe that's idiomatic. Am I just thinking like a dinosaur? Maybe the `import ..Foo` syntax helps with this? (I avoided that syntax because I like to find all users of the module `Celeste.Foo` by grepping for 'import Celeste.Foo'.) I never fully groked the Julia module system; I suspect you guys know best practices better than I did.

jeff-regier:
Thanks for the input, Steve!

Kyle, let me know what you think is best. There's also `src/settings.jl` now, which reads a yaml file. Maybe we could move the one setting in `Config` (`min_pixel_radius`) to the yaml file that `settings.jl` reads?

kbarbary:
I'll admit this PR is a little "code-churny" and not very important. But I think Celeste is a little "over-modularlized" and this was one really obvious place where the hierarchy could be flattened a bit. I think julia `Base` is a good pattern to follow: a lot of stuff lives in top-level `Base`, but there are also a few submodules. Plenty of those submodules import functions or types from top-level `Base`. Generally, modules are really useful for avoiding namespace collisions, but get confusing when used to organize concepts in a deep hierarchy.

This is no better or worse in terms of circular dependencies. The `Config` type still has to be defined before other modules are defined that try to import it, just like the `Configs` module had to be defined before other modules tried to import it. The only way you'd have a circular dependency is if `Config` (or `Configs`) had a dependency on another submodule that imported `Config`.

Regarding `import ..Foo` vs `import Celeste: Foo` or `import ..Celeste: Foo`, I *think* they're equivalent, but the latter seems more idiomatic now that I look in Base. I can change that back. It just strikes me as strange that `import ..Celeste: Foo` works at all from within a submodule `Celeste.Bar`: I take it to mean "look in the containing module of `Bar` (in this case that would be `Celeste`) for the submodule `Celeste`. But `Celeste` doesn't contain itself, does it? In Python you always use relative imports within a package, so the containing package name `Celeste` would never be written explicitly like this.

@jeff-regier Sounds like it makes sense to make `Config` and `read_settings_file()` play together in the future. That will probably have to wait until after I/O refactoring because of the dependence on SDSSIO though.

kbarbary:
Anyway, I think we should probably just merge this as is. It would be nice to settle on some import pattern and use it consistently everywhere. But it's already not consistent throughout Celeste and I'm honestly not sure what is considered best practices for this.

jeff-regier:
Sounds good

gostevehoward:
Thanks for the comments, that is interesting. If that's how Julia's base package works, then it must be a good pattern to follow. I was probably thinking of the situation in Python where module `foo` imports module `bar` and vice versa, and it causes problems. Sounds like this isn't exactly analogous.

PR #689, opened 2017-08-29
Remove multiple includes of bivariate_normals code

kbarbary:
This is simply code organization, removing the multiple `include`s of `bivariate_normals.jl`. Instead, the code now lives solely in the `Model` submodule and is imported into other submodules as needed (`PSF` and `Synthetic`.

~I also split the `bivariate_normals.jl` file into two files:~
- ~`model/bivariate_normals.jl` model independent stuff about bivariate normals. Could live outside `Model` if desired.~
- ~`model/derivatives.jl`: stuff specific to the Galaxy or PSF model, using bivariate normals. Mostly about derivatives.~

### DIFF
--- a/src/Model.jl
+++ b/src/Model.jl
@@ -32,6 +32,7 @@ import FITSIO, WCS
 import WCS.WCSTransform
 import ..Log
 import ..Celeste: Const, @aliasscope, @unroll_loop
+import ..SensitiveFloats: SensitiveFloat, clear!
 
 import Base.length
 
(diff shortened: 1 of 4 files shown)


jeff-regier:
I'm pretty sure we'll want to keep the bivariate normal derivative code together with the code that's being differentiated: the two are very tightly coupled. Even the slightest change to the code means the derivatives need to change. Is there some other way to achieve what you're going for in this PR?

kbarbary:
Assuming you're referring to splitting `bivariate_normals.jl` and `derivatives.jl`, absolutely I can put them back in the same file. (They're currently `include`d one after the other, so this is only a matter of organization.)

It wasn't apparent from reading the code that the stuff in the first file depended on anything in the second, and I just thought that a separation between "generic math stuff" and "model-specific stuff that uses the generic math stuff" might be useful for humans.

kbarbary:
Removed split.

jeff-regier:
Some of the variable names in that file unfortunately suggest that code is specific to the model (e.g. galaxies). In fact it's all better thought of as "generic math stuff", though you wouldn't know it from the variable names.

PR #691, opened 2017-09-01
Organization: moduleize bivariate normals

kbarbary:
This moves the part of `model/bivariate_normals.jl` that is model-independent (just about 2-d gaussians, not galaxies) to its own submodule `BivariateNormals`. This is the entire file, with the exception of `GalaxyCacheComponent` and `BvnBundle`. This will hopefully make things a little easier to understand, or make it easier to do so in the future.

I ran into a bit of a problem with the `clear!` function and I ended up renaming `clear!` to `zero!` as the former could be confused with the `clear!` in Base that has a completely different meaning. Several submodules now have their own `zero!` function that acts on their own types: `SensitiveFloats.zero!`, `Model.zero!`, `BivariateNormals.zero!` and `DeterministicVI.zero!`. We could make these methods of a single function `Celeste.zero!`, but that isn't done in this PR.

Finally, I updated a bit of code to the new `where` syntax while I was poking around.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,10 @@ notifications:
   email: false
 git:
   depth: 99999999
-
+addons:
+  apt:
+    packages:
+    - hdf5-tools
 after_success:
   - julia -e 'cd(Pkg.dir("Celeste")); Pkg.add("Coverage"); using Coverage; Codecov.submit(Codecov.process_folder())'
 
--- a/REQUIRE
+++ b/REQUIRE
@@ -2,7 +2,7 @@ julia 0.6
 BinDeps
 Compat 0.18.0
 DataFrames
-DataStructures 0.5.2
+DataStructures 0.7.0
 Distributions
 FITSIO 0.8.4
 DiffBase 0.0.3
--- a/src/BivariateNormals.jl
+++ b/src/BivariateNormals.jl
@@ -1,10 +1,19 @@
-# Bivarate normals -- 2-d gaussians and their derivatives, as well as
-# derivatives of model components.
+"""BivarateNormals
+
+2-d gaussians and their derivatives.
+"""
+module BivariateNormals
+
+using StaticArrays
+using ..Celeste: Const, @aliasscope, @unroll_loop
+
+# TODO: Many parameter and type names sound specific to "galaxies". However,
+# they're actually general to bivariate normals or mixtures thereof.
+# Parameters should be renamed to reflect this.
 
 # Hardcoding this might not be needed but it is critical that this is
 # compile time constant.
 const GAL_SHAPE_IDS_LENGTH = 3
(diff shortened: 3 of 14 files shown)


jeff-regier:
All 3 things seem like nice improvements. Will you merge this once the tests pass?

kbarbary:
There's a travis build error on HDF5. Already tried restarting the job to no avail:
```
LoadError: failed process: Process(`sudo apt-get install hdf5-tools`, ProcessExited(1)) [1]
while loading /home/travis/.julia/v0.6/HDF5/deps/build.jl, in expression starting on line 39
```

jeff-regier:
I tried restarting the build too, without luck. The weird thing is that master branch doesn't encounter build error. (I tried restarting it too.)

kbarbary:
Looks like this branch ran on Travis's trusty (Ubuntu 14.04) infrastructure, whereas master ran on precise (12.04). `hdf5-tools` is on the [apt whitelist](https://github.com/travis-ci/apt-package-whitelist) for both precise and trusty. I think maybe the difference is that trusty defaults to `sudo: false`, so we might have to add a section to `.travis.yml` like:
```
addons:
  apt:
    packages:
    - hdf5-tools
```
I'll try that.

jeff-regier:
We might want to turn on code inlining for the unit tests. The unit tests are twice as fast if inlining is allowed, but the code coverage statistics aren't as accurate. Maybe worth it for now.

kbarbary:
I thought inlining was already on? The test command run on Travis,
```
$ julia --check-bounds=yes --color=yes -e "Pkg.test(\"${JL_PKG}\", coverage=true)"
```
doesn't have `--inline=no`. Is there some other place that inlining is turned off?

jeff-regier:
Oh I didn't realize that inlining was enabled. So much for that idea. I'm not sure why these tests are taking so long.

kbarbary:
Could be all the deprecation warnings on Travis. Seems like some updated dependencies have deprecations. I'll try to fix those.

jeff-regier:
Still seeing depwarns in travis:
```
WARNING: Compat.ASCIIString is deprecated, use String instead.
  likely near /home/travis/.julia/v0.6/Celeste/test/test_galsim_benchmarks.jl:41
```

jeff-regier:
For most of the unit tests, travis runs 3x--5x slower than my machine (when it's not computing coverage statistics). But for the galsim_benchmarks, it's more like 12x. Kind of suspicious.

giordano:
Travis builds are being quite slow lately, most builds of Julia fail because of timeout. Julia developers had to ask for 30 minutes more of timeout time. This may be at least part of the reason

kbarbary:
The `Compat.ASCIIString` depwarns are most likely from FITSIO. Need to release a new version (but first fix JuliaAstro/FITSIO.jl#75).

PR #696, opened 2017-09-16
Remove objid and thing_id / thingid

kbarbary:
This removes the SDSS-specific `objid` and `thing_id` fields from the core of Celeste (primarily `CatalogEntry`). Mostly these were not used for anything, so this doesn't have much of an effect.

In the benchmarks, `objid` was sometimes used as an object index. It was necessary to have an object index when reshaping (`melt`ing or `stack`ing) DataFrames. However, it wasn't necessary for the index to be globally unique or meaningful. So here, I've replaced it with a simple running index (`1:length(catalog)`) within the catalog.

References to `objid`/`thing_id` still exist in the `SDSSIO` module which is totally fine. The code there has the ability to read tables with those fields, they're just not used by any of the rest of the code.


jeff-regier:
Great! Bonus: code base is now 56 lines shorter.

PR #705, opened 2017-09-19
increase test coverage for ParallelRun.jl

### LINKED ISSUE #700: unit tests for ParallelRun
We're only at 40% coverage currently. Getting rid of all the timing code will boost coverage somewhat.

jeff-regier:
ParallelRun.jl test coverage up from 39% to 78%. Would be near 100%, but `infer_init_new` hasn't yet replaced `infer_init`.

Closes #700


kbarbary (approved):
Looks good to me.

PR #706, opened 2017-09-19
fix SEP build on osx

### LINKED ISSUE #704: restore Mac OS X support
SEP doesn't build on OS X because `make` doesn't accept the `-D` flag on OS X.

kbarbary:
Fixes #704.

@jeff-regier This is now tested in the main sep repository, but you might want to try this fix locally since we don't have an OSX travis job on Celeste.

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -2,7 +2,7 @@ using BinDeps
 
 @BinDeps.setup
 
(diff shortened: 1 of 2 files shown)


jeff-regier:
Works on my os x laptop. Thanks for the fix.

PR #715, opened 2017-09-29
RawPSF -> AbstractPSFMap in Image

kbarbary:
This is a step in making the core of Celeste independent of the SDSS data model.

It Makes the `Image` type parametric in the last attribute (formerly called a `RawPSF`) which represents the rasterized PSF as a function of position in the image.  `Image` can now contain any type of `AbstractPSFMap`, which has one job: to return a rasterized PSF model, given a pixel position. This allows for different implementations, such as `SDSSPSFMap`, or `ConstantPSFMap` (a non-varying PSF), or any other survey's PSF representation, without any of the rest of Celeste knowing the type or the implementation.

The alternative to this would be fitting a spatially variable version of Celeste's PSF model (`Vector{PsfComponent}`) and storing that in `Image`. (That would be cleaner in some ways and can of course still be done in the future.)

Note: most of the big blocks of changes here are simply moving code around: the SDSS specific PSF has moved to SDSSIO.jl for example.

### DIFF
--- a/src/AccuracyBenchmark.jl
+++ b/src/AccuracyBenchmark.jl
@@ -541,22 +541,8 @@ function make_image(
         ones(height_px),
     )
 
-    # The next code block generates an SDSS-style raw_psf with just one
-    # eigenimage. That eigenimage is obtained by rendering the Celeste PSF
-    # on a grid.
-    psf_dims = (51, 51)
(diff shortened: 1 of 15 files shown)


jeff-regier:
Awesome!

PR #716, opened 2017-09-29
ais inference

### LINKED ISSUE #324: mcmc inference

### LINKED ISSUE #365: Score MCMC point estimates on galsim dataset

### LINKED ISSUE #387: Score MCMC point estimates with `benchmark/stripe82/validate_on_stripe82.jl`

andymiller:
added support for MCMC within AIS inference

- implemented annealed importance sampling to approximate the marginal likelihood of a star vs galaxy
- code for running slice-sampling within AIS on both star and galaxy conditional distributions
- code for consolidating samples and assembling output dataframes (`mcmc/mcmc_infer.jl`)
- added a `process_source_mcmc` function in `ParallelRun.jl`; added a boolean `do_vi` flag to `one_node_single_infer` (defaults to VI)

The following call should return a list of `res` objects (dictionaries)
```
 res_list = ParallelRun.one_node_single_infer(
     catalog_entries, target_sources, neighbor_map, images;
     config=Config(25.0), do_vi=false)
```
each object is a dictionary, with the following entries
- `star_samples` and `gal_samples` : dataframe of parameter samples
- `star_lls` and `gal_lls` : log joint probability for the star or galaxy samples
- `ave_pstar` : average log probability that the source is a star
- `star_bootstrap`, `gal_bootstrap`: bootstrap samples of the marginal likelihood of the data under the star model and the gal model, respectively (these are used to compute the posterior prob(star | data)). 

The list of dictionaries above can be consolidated into a one-source-per-row dataframe with approximate posterior means and stderr values with the following call
```
star_summary, gal_summary, joint_summary = MCMC.consolidate_samples(res_list)
```
Here, `star_summary` assumes all sources are stars, `gal_summary` assumes all sources are galaxies, and `joint_summary` presents star posterior inferences if `exp(ave_pstar) > .5`, and galaxy posterior inferences otherwise. 

Closes #387 . Closes #324 .

### DIFF
--- a/src/mcmc/mcmc_misc.jl
+++ b/src/mcmc/mcmc_misc.jl
@@ -103,7 +103,7 @@ function catalog_to_data_frame_row(catalog_entry; objid="truth")
 end
 
 function samples_to_dataframe(chain; is_star=true)
-    """ Turns MCMC samples (star or galaxy chain) into a dataframe of samples 
+    """ Turns MCMC samples (star or galaxy chain) into a dataframe of samples
     with parameters we can compare to Photo and VB inferences
 
     Conversions performed:
@@ -249,7 +249,7 @@ function write_galaxy_unit_flux(pos::Array{Float64, 1},
                                 pixels::Matrix{Float64};
                                 offset::Array{Float64, 1}=[0., 0.],
                                 flux::Float64=1.)
-    # write the unit-flux scaled, correctly offset galaxy shape + 
+    # write the unit-flux scaled, correctly offset galaxy shape +
     # psf convolution
     e_devs = [gal_frac_dev, 1 - gal_frac_dev]
     XiXi = Model.get_bvn_cov(gal_axis_ratio, gal_angle, gal_scale)
@@ -275,7 +275,7 @@ function write_galaxy_unit_flux_pixel(px_pos::Array{Float64, 1},
                                       gal_scale::Float64,
                                       pixels::Matrix{Float64};
                                       flux::Float64=1.)
-    # write the unit-flux scaled, correctly offset galaxy shape + 
+    # write the unit-flux scaled, correctly offset galaxy shape +
     # psf convolution
     e_devs = [gal_frac_dev, 1 - gal_frac_dev]
     XiXi = Model.get_bvn_cov(gal_axis_ratio, gal_angle, gal_scale)
@@ -356,7 +356,7 @@ end
 
 
 #########################################
-# older interface 
+# older interface
 #########################################
 function write_star_unit_flux(img0::Image,
                               pos::Array{Float64, 1},
@@ -451,7 +451,7 @@ function patch_to_image(patch::SkyPatch, img::Image; round_pixels_to_int=true)
     calibration = img.sky.calibration[Hr]
     sky_x       = img.sky.sky_x[Hr]
     sky_y       = img.sky.sky_y[Wr]
-    sky         = SkyIntensity(sky_small, sky_x, sky_y, calibration)
+    sky         = SDSSBackground(sky_small, sky_x, sky_y, calibration)
     nelec_per_nmgy = img.nelec_per_nmgy[Hr]
 
     # TODO create an adjusted WCS object
@@ -465,5 +465,3 @@ function patch_to_image(patch::SkyPatch, img::Image; round_pixels_to_int=true)
                         sky, nelec_per_nmgy, img.raw_psf_comp)
     return patch_image
 end
-
-
--- a/src/mcmc/mcmc_misc.jl
+++ b/src/mcmc/mcmc_misc.jl
@@ -15,9 +15,17 @@ end
 
 # output a single poisson log likelihood
 function poisson_lnpdf(data, lam)
-    return float(logpdf(Distributions.Poisson(lam), data))
+    #lldist = float(logpdf(Distributions.Poisson(lam), data))
+    return data*log(lam) - lam - lgamma(data+1.)
 end
 
+# constrain checking
+function inrange(val, a, b)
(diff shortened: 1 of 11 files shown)


jeff-regier:
Thanks for fixing the merge conflicts.

Would you add a unit test that calls `ParallelRun.one_node_single_infer` with `do_vi = false`? Maybe  use `config=Config(2.0)` to keep the test fast.

That test should exercise basically all your MCMC code, right? If there's code not exercised by that, can we omit it from the PR?

jeff-regier:
Stuff like `write_star_unit_flux_raw` and `write_galaxy_unit_flux` and `render_patch` duplicates code in `src/Synthetic.jl`. We want to avoid duplicating that code especially, since it could lead to us doing slightly different tests.

andymiller:
I agree there is some overlap in the `write_star_unit_flux_raw` and `render_patch` code.  I reimplemented the old version from `Synthetic`  to  control the `offset` so I could use a `WCS` object on a small patch of image.  I realize that is now supported in `write_star_nmgy!` method --- but one thing I'm hesitant of is re-instantiating a `SkyPatch` (and then loading the active pixels) every time the likelihood function is called.

jeff-regier:
Would it work to make `p` (the `SkyPatch`) an argument to `write_star_nmgy!`?

andymiller:
oh also the pixel map in `write_star_unit_flux_raw` and others is just an array.  What if I wrote a general method

```
function write_star_nmgy!(world_pos::Array{Float64,1}, patch::SkyPatch,  flux::float64, pixels::Matrix{Float64} )
    fs0m = SensitiveFloat{Float64}(length(StarPosParams), 1, false, false)

    H2, W2 = size(p.active_pixel_bitmap)
    for w2 in 1:W2, h2 in 1:H2
        h = p.bitmap_offset[1] + h2
        w = p.bitmap_offset[2] + w2
        Model.star_light_density!(fs0m, p, h, w, pos, false)
        pixels[h, w] += fs0m.v[] * flux
    end
end
```

and reimplemented what is in Synthetic with 

```
function write_star_nmgy!(img::Image, ce::CatalogEntry)
    p = Model.SkyPatch(img, ce, radius_override_pix=25)
    write_star_nmgy!(ce.pos, p, ce.star_fluxes[img.b], img.pixels) 
end
```

and did the same for `write_galaxy_nmgy!`.  Then I can call all of this code from the `MCMC` module.

src/mcmc/mcmc_misc.jl
```diff
@@ -174,44 +199,135 @@ end
...
     end
 end
 
+
+"""
+Generate a model image on a patch, according to that image/patch psf
+"""
+function render_patch(img::Image, patch::SkyPatch, n_bodies::Vector{CatalogEntry})
+    # sky noise an gain
+    epsilon = img.sky[1, 1]
```
jeff-regier, line 300: This could be a problem because the synthetic imagery uses a variable sky.

src/mcmc/mcmc_misc.jl
```diff
@@ -174,44 +199,135 @@ end
...
+    iota    = Float64(median(img.nelec_per_nmgy))
```
jeff-regier, line 301: Also potentially a problem because `nelec_per_nmgy` varies per row, both in the real images and the synthetic images (from Synthetic.jl).

src/mcmc/mcmc_misc.jl
```diff
@@ -174,44 +199,135 @@ end
...
-                the_mean = SVector{2}(WCS.world_to_pix(img0.wcs, pos)) +
-                           img0.psf[k].xiBar
-                the_cov = img0.psf[k].tauBar + gproto.nuBar * XiXi
-                intensity = iota * img0.psf[k].alphaBar * gal_frac_devs[i] *
-                    gproto.etaBar
+            for k in 1:length(psf)
+                the_mean = px_pos + psf[k].xiBar
+                the_cov = psf[k].tauBar + gproto.nuBar * XiXi
+                intensity = flux * iota * psf[k].alphaBar * e_devs[i] * gproto.etaBar
                 write_gaussian(the_mean, the_cov, intensity, pixels)
```
jeff-regier, line 288: This implementation of `write_gaussian`, back when it was used in `Synthetic.jl`, wasn't particularly efficient (compared to what's in `fsm_utils.jl`). That's because we were just using it to render Synthetic data, not to evaluate the objective function. But it sounds like you're using it at every iteration? You should get a good speed up from calling `write_galaxy_nmgy!`, which now uses `fsm_utils.jl`.

andymiller: Yes, I'm hoping this will result in a speedup. 

I'll incorporate `write_star_nmgy!` and `write_galaxy_nmgy!` into the the branch, and remove the redundant code.  I don't rely on `render_patch` for the likelihood, so I can remove it.

jeff-regier:
I just fixed a minor merge conflict with master: please pull before editing.

Looks like `make_position_transformations` and `sample_colors` and `sample_logr` and `parameters_from_catalog` aren't covered by tests. Are they necessary? Could we safely omit them from the master branch?

No need to write thorough tests, but it's nice if every function in the src directory of master gets called once, just to catch syntax errors if anyone is making changes without totally understanding what's going on.

jeff-regier:
Ah, this latest merge had more conflicts that I expected. Let's get this PR committed as is, to avoid more merge conflicts.

Andy, please do address the stuff in this thread in another PR. Please start a new branch off of master rather than continuing an old branch---otherwise all the old commits get mixed in.

andymiller:
ok, i moved all my recent changes to a new branch `ais-benchmark` and squashed the changes.  Will open a new PR shortly.

PR #718, opened 2017-10-04
remove RawImage type

kbarbary:
Removes `RawImage`, the SDSS specific version of `Image`, was very similar to `Image`, and only used in one place outside of SDSSIO. After this, the only image-reading interface is `SDSSIO.load_field_images(strategy, rcf)`, which returns a `Vector{Image}`; no need for `RawImage`.

A lot of the changeset here is just moving `IOStrategy` stuff earlier in the SDSSIO module (so that we can dispatch on it in functions below).

### DIFF
--- a/src/ParallelRun.jl
+++ b/src/ParallelRun.jl
@@ -200,7 +200,7 @@ end
 """
     detect_sources(images)
 
-Detect sources in a set of (possibly overlapping) `RawImage`s and combine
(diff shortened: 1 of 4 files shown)


jeff-regier (approved):
I think the idea behind RawImage was to separate the data loading (when all threads but 1 are blocked) from the expensive call to the function that fits the Celeste psf. In hindsight, it would be better just to just leave the celeste_psf as an empty vector at load time, and populate it later.

PR #725, opened 2017-12-13
Grab bag of organization and test cleanups

kbarbary:
This PR looks big but it makes no algorithmic changes. It is only code organization and test clean ups. (The files with big diffs are either moving code around or reindentation). The most notable feature is that tests now use `@testset` uniformly and all extraneous print output has been removed or made optional via logging levels. This means that the test output is now actually readable. Example:

```
$ runtests.jl sdssio sep wcs
Test Summary: | Pass  Total
sdssio        |   10     10
Test Summary: | Pass  Total
sep           |    5      5
Test Summary: | Pass  Total
wcs           |    5      5

Timing info:
                test_sdssio.jl:    3.02 seconds
                   test_sep.jl:    1.09 seconds
                   test_wcs.jl:    8.16 seconds
                    Total time:   12.27 seconds
```

Being able to read the test output made it possible to notice several deprecation warnings which I've fixed.

### Detailed summary of changes:

- Rename `SkyPatch` to `ImagePatch`, because it represents a pixel box on a given image, not a region on the sky.
- Code movement: move `detect_sources()` to top level and own file. Move contents of `joint_infer.jl` into `ParallelRun.jl`, except partitioning related code which gets its own file. This organization makes sense for future changes I plan to make to organize the main entry points for joint and single inference: having them in the same file is handy. The partitioning stuff is independent of everything else so it can live by itself.
- Make `neighbor_map` a dictionary mapping (number in catalog) -> (vector of numbers in catalog). I find this easier to think about and it allows some minor code simplifications.
- Log module: allow multiple logging levels: error, warn, info, debug. Allows one to show detail during debugging or running, but to hide the details when running tests.
- All tests now get their images and catalogs from disk via `SampleData.get_sdss_[images,catalog](run, camcol, field)` functions. This removes some code duplication and will make future refactoring easier.
- Extraneous printouts in tests and library code removed, moved to scripts or made optional.
- Remove some unused functions from SDSSIO.


jeff-regier:
Excellent. These all seem like nice improvements.

PR #726, opened 2017-12-21
fix test comparing joint and single infer

kbarbary:
Currently there isn't actually a test of joint inference. It had been tested in the long-running tests, but these have become quite out-of-date and thoroughly broken. This PR rescues one of those tests and leaves the rest of the long-running tests in a file `test/outofdate.jl`. (It is not clear how to fix these or whether they are really necessary and it's not currently a priority.)

The rescued test has been moved to `test_infer.jl` and is now always run. It runs joint VI and single VI on a few overlapping sources and checks that the joint result is better.

There are also a few whitespace changes and minor cleanups in `ParallelRun.jl`.

### DIFF
--- a/src/ParallelRun.jl
+++ b/src/ParallelRun.jl
@@ -193,7 +193,8 @@ function one_node_joint_infer(catalog, patches, target_sources, neighbor_map,
     for i = 1:n_sources
         entry = catalog[target_sources[i]]
         is_sky_bad = bad_sky(entry, images)
-        result = OptimizedSource(entry.pos[1], entry.pos[2], vp_vec[i][1], is_sky_bad)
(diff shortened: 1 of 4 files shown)


jeff-regier:
When possible, better to treat the joint inference code and constraint transforms code as black boxes. But perhaps some changes to them can't be avoided to run on DECaLS.

kbarbary:
> When possible, better to treat the joint inference code and constraint transforms code as black boxes.

Yes, it would be great to be able to do this. But the tests which need to be updated use many internals, so there is no avoiding looking at the internals in order to get the tests running.

Beyond the scope of my current project, I can imagine having a better "internal API" for ELBO. One that allows it to be treated more as a black box that can provide the objective function and gradient in single function calls.

PR #727, opened 2018-01-03
updates syntax to work with latest DataFrames package

jeff-regier:
They're using the keyword `missing` now instead of `NA`.


kbarbary:
:+1: I guess this stuff wasn't tested in the test suite, otherwise I would have got it in #724.

PR #730, opened 2018-01-08
update to new `where` syntax

kbarbary:
This updates all parameteric method definitions to use the new `where` syntax in Julia 0.6 (I believe required in 0.7).

Along the way, I made the style more uniform and conventional by always using `T` for a numeric type rather than `NumType` and always using 4-space indents. (Reindenting is the reason for the large diff.)

This also has two bugfixes:

- In `src/mcmc/mcmc_infer.jl`: Sometimes the `@sprintf` calls were resulting in a string with an embedded NULL, causing an error when passed to `Log.info()`. I believe (but am not certain) that this happens when the format code is not long enough for the value. I widened the format codes.

- In `src/model/light_source_model.jl`: We were getting errors from JLD/HDF5 that *might* have been due to from calling `load_prior()` from threads (`load_prior()` does I/O with JLD). I replaced the `load_prior()` function. It now returns a copy of the prior loaded on module import. `load_prior_init()` does the I/O and is now called only on module import.

### DIFF
--- a/src/BivariateNormals.jl
+++ b/src/BivariateNormals.jl
@@ -26,10 +26,9 @@ Args:
 Returns:
   The 2x2 covariance matrix parameterized by the inputs.
 """
-function get_bvn_cov{NumType <: Number}(
-    ab::NumType, angle::NumType, scale::NumType)
+function get_bvn_cov(ab::T, angle::T, scale::T) where {T<:Number}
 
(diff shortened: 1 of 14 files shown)


jeff-regier:
Great, glad we'll be ready for julia 1.0.

PR #738, opened 2018-09-14
Use column names with proper capitalization

### LINKED ISSUE #737: Celeste relies on case-insensitive names of columns
In `FITSIO.jl` we changed the default behavior of `fits_get_colnum` to be case-sensitive (see https://github.com/JuliaAstro/FITSIO.jl/issues/98 and https://github.com/JuliaAstro/FITSIO.jl/issues/99), but this can be controlled with the new `case_sensitive` keyword to `fits_get_colnum` and `read`.

`Celeste.jl` relies on the old case-insensitive behavior and in `src/SDSSIO.jl` it asks for the "rnrow" column while its actual name is "RNROW".  Similarly for other columns.  See https://github.com/JuliaLang/METADATA.jl/pull/18001

I see the following options:
1. Use the column names with the proper capitalization.  We should hope that all files are consistent
2. Require the new version of `FITSIO.jl` and use `case_sensitive=false`
3. Set an upper bound to the required version of `FITSIO.jl` and do nothing else.

giordano:
This is my take to fix #737 following the first solution I outlined there.  This doesn't need changes in the required version of FITSIO.jl.  If this solution turns out to not be enough because the columns are named with different capitalization, I'd suggest to move forward to a new version of FITSIO.jl (not yet tagged, though) and use the `case_sensitive=false` keyword.

### DIFF
--- a/src/SDSSIO.jl
+++ b/src/SDSSIO.jl
@@ -458,10 +458,10 @@ function read_psfmap(fitsfile::FITSIO.FITS, band::Char)
     nrows = FITSIO.read_key(hdu, "NAXIS2")[1]::Int
     nrow_b = Int((read(hdu, "nrow_b")::Vector{Int32})[1])
     ncol_b = Int((read(hdu, "ncol_b")::Vector{Int32})[1])
-    rnrow = (read(hdu, "rnrow")::Vector{Int32})[1]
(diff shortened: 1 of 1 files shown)


giordano:
Done in https://github.com/jeff-regier/Celeste.jl/commit/41c4897dde496ce5a8593593670d2e798a2a6076

